PR #2, opened 2015-04-08
Fix gh-page index page

merih:
Readme points to the old gh-pages index page. Since it is changed from `pointerEvents.html` to `index.html`, the page is 404.


merih:
@itsananderson was faster, removing

PR #13, opened 2015-04-21
Add direction-specific touch-action pan values

RByers:
Creating a pull request for my original changes.  Will apply feedback already received on top of this (for tracking purposes).


RByers:
Looks like squashing all the commits into one had the effect of loosing the comment history on the intermediate state.  Sorry.  I guess in the future I shouldn't squash pull requests (just thought it was a little messy having each independent intermediate commit in the history).

PR #18, opened 2015-07-12
Add missing pointerleave

patrickhlauke:
Somehow the `pointerleave` fell by the wayside in the non-normative note about expected events for non-hover capable devices.

This fixup would also apply to the current level 1 spec. Possible to add that to the actual REC as a quick fix?

### DIFF
--- a/index.html
+++ b/index.html
@@ -864,6 +864,7 @@ these pointers will all produce <a href="#compatibility-mapping-with-mouse-event
 				<li><code>click</code></li>
 				<li><code>pointerout</code></li>
 				<li><code>mouseout</code></li>
+				<li><code>pointerleave</code></li>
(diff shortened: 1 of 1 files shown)


RByers:
Thanks.
We haven't made any errata to the REC yet, right?  IMHO this isn't a big enough deal to open that can of worms, but maybe it's easy to do - I don't know.

patrickhlauke:
Ok to merge into Level 2 for now though?

RByers:
Yep, was just doing that (flaky WiFi, GitHub was timing out on me)

AFBarstow:
@patrickhlauke, no it isn't possible to directly edited a Recommendation (although in IMHO, it should be possible for tweaks like this).

PR #20, opened 2015-09-11
Add a note about the 300ms click delay

RByers:
Talking about specific gestures is out of scope for the pointer events charter / specification.  But hopefully we can at least have a note that describes the important impact touch-action has on the click delay?  Here I attempt to do that without even using the word "zoom", WDYT?

See related twitter discussion: https://twitter.com/awfulben/status/642163198511026176


RByers:
@jacobrossi WDYT?

jacobrossi:
Good idea. Can you s/tap gesture/click event?

RByers:
Thanks, how's this?  (I'll squash all commits together before merging - don't need the intermediate history here).

patrickhlauke:
From a reader's point of view, this version is a bit unsatisfactory as it doesn't make it clear _why_ user agents must add the 300ms delay. perhaps after `<code>click</code> event` we could have `to account for the possibility of a subsequent tap which may trigger further default behavior`? still a very roundabout way of saying what could really be said in four simple words, though...

RByers:
Seems like something like that could work.  It could even be "subsequent touch action" - eg. Chrome has double-tap-drag (tap and then drag to zoom) which can have the same issue as double tap (though strangely [doesn't](https://code.google.com/p/chromium/issues/detail?id=530717) today).

patrickhlauke:
perhaps also change `often must` to `typically`

jacobrossi:
Let me run this by some folks on my side to see how flexible we can be. :-)

RByers:
> perhaps also change often must to typically

done

I also added a bit more that should be uncontroversial: "(to resolve ambiguity around possible subsequent touches)".

> Let me run this by some folks on my side to see how flexible we can be. :-)

Ok.  Let me know if that's likely to take more than a couple days and we can land the uncontroversial wording for now, then hope to elaborate later :-)

jacobrossi:
We can use Patrick's proposed suggestions (including calling double-tap / tap out specifically for clarity) since that would make this clearer.  Perhaps we can add something like this just to make it clear what's out of scope?

```
Methods for determining a tap or double-tap gesture are out of scope for this specification.
```

patrickhlauke:
:shipit:

RByers:
Great, so like this then?  (I squashed all the commits)

RByers:
@patrickhlauke @jacobrossi let me know if either of you want to review the final text here before I merge it.  Otherwise it sounds like it's safe to merge and tweak later as desired.

patrickhlauke:
If I can be a right wordy pain in the neck... `other behavior` > `other behaviors`. Also, after mentioning the 300ms delay and `auto`, the note seems to just hang there and not follow through with the next part...that you can disable that delay. And although we're already in a note, that last sentence may need a bit of a lead-in. How's this:

```
<section class="note">Disabling some default touch behaviors may allow user agents to respond to other behaviors more quickly.  For example, with <code>auto</code> user agents typically add 300ms of delay before <code>click</code> to allow for double-tap gestures to be handled.  In these cases, explicitly setting <code>touch-action: none</code> or <code>touch-action: manipulation</code> will remove this delay. Note that the methods for determining a tap or double-tap gesture are out of scope for this specification.</section>
```

RByers:
Thanks, I like that better (I was definitely getting sloppy) - using it verbatim.  I think there's been enough discussion now that this can land, but happy to take more feedback in followup commits.

PR #21, opened 2015-09-17
Fix respec warnings about use of deprecated title attribute

RByers:
Also removed some unnecessary class=internalDFN attributes, and
adds links to a couple more uses of defined terms.


RByers:
With this CL, the spec now renders without any respec warnings

RByers:
Hearing no concerns I'm just going to merge this tweak.  Happy to address issues via follow-up CLs.

Issue #19, opened 2015-09-08
Touch action processing model doesn't match IE/Edge behavior around zooming


RByers:
Hey,
The [touch-action processing model](https://w3c.github.io/pointerevents/#the-touch-action-css-property) says:

> To determine the effect of a touch, find the nearest ancestor (starting from the element itself) that has a default touch behavior. Then examine the touch-action property of each element between the hit tested element and the element with the default touch behavior (including both the hit tested element and the element with the default touch behavior). If the touch-action property of any of those elements disallows the default touch behavior, do nothing. 

This makes sense for scrolling, but what about zooming?

In particular, we've been saying "use `touch-action: manipulation` to disable the click delay" but what if your page has scrollable divs?  Do you need to re-apply `touch-action manipulation` to each scroller to ensure the click delay is disabled everywhere?  The above spec text implies that you do. But this sucks for this click-delay use case (means touch-action isn't a complete substitute for fastclick.js afterall).

Happily, this isn't [actually what IE/Edge implement](http://jsbin.com/tipure/).  So I think we should correct the spec text and [blink implementation](https://code.google.com/p/chromium/issues/detail?id=529295) to match the IE / Edge behavior.  I seem to recall some earlier iteration of the spec that referred to the ancestor that implements the specific default touch behavior being considered (as opposed to any "default touch behavior").  I know we agonized over trying to keep this definition accurate yet simple.  Maybe we need to go back through the history here?

Anyway, if there's agreement that the spec should be corrected to match the Edge behavior, then I can take a crack at coming up with some text for this.  Perhaps we should just be more verbose - describing an algorithm in stages, etc.

@jacobrossi can you describe precisely what the Edge behavior here is to make sure I'm not missing something subtle?

RByers:
@teddink said (on the list):

> I spoke with Jacob (and others) here is what I was able to glean from those conversations:
> 
> This is covered by this section:
> 
> "When a user touches an element, the effect of that touch is determined by the value of the touch-action property and the default touch behaviors on the element and its ancestors. To determine the effect of a touch, find the nearest ancestor (starting from the element itself) that has a default touch behavior. Then examine the touch-action property of each element between the hit tested element and the element with the default touch behavior (including both the hit tested element and the element with the default touch behavior). If the touch-action property of any of those elements disallows the default touch behavior, do nothing. Otherwise allow the element to start considering the touch for the purposes of executing a default touch behavior."
> 
[...]

PR #24, opened 2015-10-22
Fix the touch-action processing model for zoom scenarios

RByers:
It's incorrect to consider the touch-action values only up to the nearest
ancestor that supports ANY default touch behavior.  Eg. in Edge you can
disable double-tap to zoom across an entire page (including any iframes
or other scrolling elements) by applying 'touch-action: manipulation' to
the html element.  See http://jsbin.com/tipure/ and http://jsbin.com/jucuna.
This is issue #19.

This change also moves the core processing model details up to
above the specific value definitions to make it easier to follow
without getting lost in all the value-specific notes.


RByers:
@jacobrossi I agonized over this text to try to be precise while also avoiding specific gesture details.  After an hour, this was the best I could come up with.  Do you think it's an improvement?

I'd like to add a web-platform-test case for this, but since which elements support "non-continuous zooming" is out of scope of the spec, I guess I can't - right?  Hopefully the jsbin demos are good enough.  

Now to fix blink....

RByers:
In particular, here's the relevant text (since I also foolishly shifted some things around in the same CL):

> A touch behavior is supported if allowed by the <code>touch-action</code> properties of all elements between the hit tested element and it's nearest ancestor with the default touch behavior (including both the hit tested element and the element with the default touch behavior).

RByers:
I talked with @jacobrossi offline and he agreed that (while not perfect) this was definitely a step in the right direction and we should just land it and continue to iterate.

Issue #33, opened 2016-01-31
Value of button attribute for pointermove event


bethge:
We have noticed an inconsistency when it comes to the `button` attribute of a `pointermove` event (https://github.com/jquery/PEP/pull/267).

In IE 11 and Edge, any `pointermove`, that does not express a button state transition, has `button: -1`.

In OSX and Win - FF 44 (with flag) a `pointermove`'s `button` attribute mirrors the `mousemove`'s `button` attribute, with the extension of `button: -1`, if no button is pressed.

According to the specification: https://www.w3.org/TR/pointerevents/#chorded-button-interactions `button` and `buttons` are inherited from: https://www.w3.org/TR/DOM-Level-3-Events/#attributes-2, which states:

> The value of button is not updated for events not caused by the depression/release of a mouse button. In these scenarios, take care not to interpret the value 0 as the left button, but rather as the un-initialized value.

Is there an expected value of `button` for a `pointermove` with a button pressed and no button state transition? Is `button: -1` the new _un-initialized value_?

The current situation for `mousemove` is quite diverse and varies across OS and browsers, with `button: 0` for Win IE and Win FF and `button: 0/1/2` (matching the currently pressed button) for OSX Safari, OSX FF, OSX CH and Win CH. Might PE be an opportunity to form a consistent behaviour?

P.S.: The same holds for `pointerenter`, `pointerover`, `pointerleave` and `pointerout`.

RByers:
I agree the spec doesn't seem entirely clear here.  In my reading, `pointermove`'s `button` should be -1 when no buttons are pressed, and the previous value otherwise (as for `mousemove`).  So it seems like an Edge bug.  @teddink WDYT?

But yes, the -1 on `pointermove` vs. 0 on `mousemove` difference is intentional.

RByers:
I should add that as far as I can see this property doesn't matter much one way or the other - there's not much value in reading the `button` value of a `pointermove` event (should probably be using `buttons` instead for any real use case).  So I'd personally also be happy to update the spec to match Edge's behavior and ask Mozilla to change.

Did you have a particular use case in mind?

bethge:
I have mulled it over some more and I feel like I now fully grasp the implications of this line:

> [...]. In order to facilitate differentiating button state transitions in any pointer event (and not just pointerdown and pointerup), the button property takes on a new value when no mouse buttons are depressed: [...]

[...]

PR #34, opened 2016-02-02
Clarify the button value for mouse drag

mustaqahmed:
The current spec wording suggests that the |button| value indicates "the last button changed" (like MouseEvent). On the other hand, the current Edge behavior suggests that |button| is "the button change that triggered the current event". The latter makes more sense since it makes it easier to differentiate between
- a pointermove fired through a button activity (chorded button press/release), and
- a pointermove fired through a non-button change (position, tilt, pressure, etc).

This patch proposes a spec wording change based on the Edge behavior.

Issue #33

### DIFF
--- a/index.html
+++ b/index.html
@@ -319,26 +319,46 @@ eventTarget.dispatchEvent(event);
 			<section>
 				<h3><dfn>Chorded Button Interactions</dfn></h3>
 				<div><p>Some pointer devices, such as mouse or pen, support multiple buttons. In the [[!DOM-LEVEL-3-EVENTS]] Mouse Event model, each button press produces a <code>mousedown</code> and <code>mouseup</code> event.  To better abstract this hardware difference and simplify cross-device input authoring, Pointer Events do not fire overlapping <code>pointerdown</code> and <code>pointerup</code> events for <a data-lt="Chorded Button Interactions">chorded button presses</a> (depressing an additional button while another button on the pointer device is already depressed).</p> 
-<p>Instead, chorded button presses can be detected by inspecting changes to the <code>button</code> and <code>buttons</code> properties. The <code>button</code> and <code>buttons</code> properties are inherited from the [[!DOM-LEVEL-3-EVENTS]] <code>MouseEvent</code> interface. In order to facilitate differentiating button state transitions in any pointer event (and not just <code>pointerdown</code> and <code>pointerup</code>), the <code>button</code> property takes on a new value when no mouse buttons are depressed:</p>
+<p>Instead, chorded button presses can be detected by inspecting changes to the <code>button</code> and <code>buttons</code> properties. The <code>button</code> and <code>buttons</code> properties are inherited from the [[!DOM-LEVEL-3-EVENTS]] <code>MouseEvent</code> interface, but with a change in semantics and values, as outlined in the following sections.</p></div>
+				<section class="note">The modifications to the <code>button</code> and <code>buttons</code> properties apply only to pointer events. For any <a>compatibility mouse events</a> the value of <code>button</code> and <code>buttons</code> should follow [[!DOM-LEVEL-3-EVENTS]].</section>
+			</section>
+			<section>
+				<h3>The <code>button</code> property</h3>
+				<div><p>To identify button state transitions in any pointer event (and not just <code>pointerdown</code> and <code>pointerup</code>), the <code>button</code> property indicates the device button whose state-change fired the event.</p></div>
 				<table class="simple">
-					<thead><tr><th>Device Button State</th><th><code>button</code></th><th><code>buttons</code></th></tr></thead>
+					<thead><tr><th>Device Button Changes</th><th><code>button</code></th></tr></thead>
 					<tbody>
-						<tr><td><b>Mouse move with no buttons pressed</b></td><td>-1</td><td>0</td></tr>
-						<tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact (with no modifier buttons pressed)</td><td>0</td><td>1</td></tr>
-						<tr><td>Middle Mouse</td><td>1</td><td>4</td></tr>
-						<tr><td>Right Mouse,<br>Pen contact with barrel button pressed</td><td>2</td><td>2</td></tr>
-						<tr><td>X1 (back) Mouse</td><td>3</td><td>8</td></tr>
-						<tr><td>X2 (forward) Mouse</td><td>4</td><td>16</td></tr>
-						<tr><td>Pen contact with eraser button pressed</td><td>5</td><td>32</td></tr>
+						<tr><td>Neither buttons nor touch/pen contact changed since last event</td><td>-1</td></tr>
+						<tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact (with no modifier buttons pressed)</td><td>0</td></tr>
+						<tr><td>Middle Mouse</td><td>1</td></tr>
+						<tr><td>Right Mouse,<br>Pen contact with barrel button pressed</td><td>2</td></tr>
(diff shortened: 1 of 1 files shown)


mustaqahmed:
@jacobrossi, please let me know if my proposed changes make the spec clearer or not.

index.html
```diff
@@ -319,26 +319,46 @@
...
-<p>Instead, chorded button presses can be detected by inspecting changes to the <code>button</code> and <code>buttons</code> properties. The <code>button</code> and <code>buttons</code> properties are inherited from the [[!DOM-LEVEL-3-EVENTS]] <code>MouseEvent</code> interface. In order to facilitate differentiating button state transitions in any pointer event (and not just <code>pointerdown</code> and <code>pointerup</code>), the <code>button</code> property takes on a new value when no mouse buttons are depressed:</p>
+<p>Instead, chorded button presses can be detected by inspecting changes to the <code>button</code> and <code>buttons</code> properties. The <code>button</code> and <code>buttons</code> properties are inherited from the [[!DOM-LEVEL-3-EVENTS]] <code>MouseEvent</code> interface with modified semantics and values; see the following sections for details.</p></div>
```
patrickhlauke: I'd suggest a subtle rewording from

> interface with modified semantics and values; see the following sections for details.

to

> interface, but with a change in semantics and values, as outlined in this section.

index.html
```diff
@@ -319,26 +319,46 @@
...
+				<section class="note">The modifications to the <code>button</code> and <code>buttons</code> properties apply only when firing pointer events. This specification does not alter the values of <code>button</code> or <code>buttons</code> used when firing mouse events. See [[DOM-LEVEL-3-EVENTS]] for the values when firing mouse events.</section>
```
patrickhlauke: I'd suggest changing

> properties apply only when firing pointer events.

to

> properties only apply to pointer events.

Also, I'd turn the following

> This specification does not alter the values of button or buttons used when firing mouse events. See [DOM-LEVEL-3-EVENTS] for the values when firing mouse events.

around to

> For any `<a>`compatibility mouse events`</a>` the value of `button` and `buttons` should follow [DOM-LEVEL-3-EVENTS]

index.html
```diff
@@ -319,26 +319,46 @@
...
+			</section>
+			<section>
+				<h3>The <code>button</code> property</h3>
+				<div><p>In order to facilitate differentiating button state transitions in any pointer event (and not just <code>pointerdown</code> and <code>pointerup</code>), the <code>button</code> property gives the device button whose state-change fired the event.</p></div>
```
patrickhlauke: > In order to facilitate differentiating

perhaps simpler

> To differentiate 

Also

> the button property gives

to

> the button property indicates

index.html
```diff
@@ -319,26 +319,46 @@
...
+						<tr><td>Neither buttons nor touch/pen contact changed since last event</td><td>-1</td></tr>
+						<tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact (with no modifier buttons pressed)</td><td>0</td></tr>
+						<tr><td>Middle Mouse</td><td>1</td></tr>
+						<tr><td>Right Mouse,<br>Pen contact with barrel button pressed</td><td>2</td></tr>
+						<tr><td>X1 (back) Mouse</td><td>3</td></tr>
+						<tr><td>X2 (forward) Mouse</td><td>4</td></tr>
+						<tr><td>Pen contact with eraser button pressed</td><td>5</td></tr>
+					</tbody>
+				</table>
+				<section class="note">During a mouse drag, the <code>button</code> property in a <code>pointermove</code> event has a value different from the same property in a <code>mousemove</code> event. For example, while moving the mouse with the right button pressed, the <code>pontermove</code> events will have <code>button = -1</code> but the <code>mousemove</code> events will have <code>button = 2</code>.</section>
```
patrickhlauke: > the button property in a pointermove event has a value different from the same property in a mousemove event

perhaps rearrange to

> the value of the button property in a pointermove event will be different from that in a mousemove event

and

> the pontermove events will have button = -1 but the mousemove events will have button = 2.

perhaps better as

> the pontermove events will have a button value of -1, while the mousemove events will have button value of 2.

index.html
```diff
@@ -319,26 +319,46 @@
...
+			</section>
+			<section>
+				<h3>The <code>buttons</code> property</h3>
+				<div><p>The <code>buttons</code> property gives the current state of the device buttons (same as in <code>MouseEvent</code> but with a bigger set of values).</p></div>
```
patrickhlauke: > same as in MouseEvent but with a bigger set of values

perhaps change to

> same as in MouseEvent, but with an expanded set of possible values

index.html
```diff
@@ -364,7 +384,7 @@
 		<h2>Pointer Event Types</h2>
 		<section>
 			<h3>Firing events using the <code>PointerEvent</code> interface</h3>
-			To <dfn>fire a pointer event name e</dfn> means to <dfn>fire an event named e</dfn> as defined in [[!DOM4]] with an event using the <a>PointerEvent</a> interface whose attributes are set as defined in <a href="#pointerevent-interface"><code>PointerEvent</code> Interface</a>.</p>
+			To <dfn>fire a pointer event name e</dfn> means to <dfn>fire an event named e</dfn> as defined in [[!DOM4]] with an event using the <a>PointerEvent</a> interface whose attributes are set as defined in <a href="#pointerevent-interface"><code>PointerEvent</code> Interface</a>.
```
patrickhlauke: for consistency with previous sections, maybe wrap this in a `<div><p>...</p></div>` as well?

patrickhlauke:
...with apologies for possibly bikeshedding some wording there. Apart from that, I think it brings the point we're trying to make across. Would it also be possible to squash this before committing, once we settled on final wordings?

mustaqahmed:
Hi Patrick:
Thanks for the careful look. Done with the suggested edits plus squashing, ptal.

index.html
```diff
@@ -319,26 +319,46 @@
...
-				</div>
 			</section>
 		</section>
 		<section>
 			<h2><dfn>The Primary Pointer</dfn></h2>
-			<div><p>In a multi-pointer (e.g. multi-touch) scenario, the <a href="#widl-PointerEvent-isPrimary"><code>isPrimary</code></a> property is used to identify a master pointer amongst the set of active pointers for each pointer type.  Only a primary pointer will produce <a>compatibility mouse events</a>. Authors who desire single-pointer interaction can achieve this by ignoring non-primary pointers (however, see the note below on <a href="#multiple-primary-pointers">multiple primary pointers</a>).
+			<div><p>In a multi-pointer (e.g. multi-touch) scenario, the <a href="#widl-PointerEvent-isPrimary"><code>isPrimary</code></a> property is used to identify a master pointer amongst the set of active pointers for each pointer type.  Only a primary pointer will produce <a>compatibility mouse events</a>. Authors who desire single-pointer interaction can achieve this by ignoring non-primary pointers (however, see the note below on <a href="#multiple-primary-pointers">multiple primary pointers</a>)</p></div>.
```
patrickhlauke: stray "." after the closing `</div>` at the end

patrickhlauke:
LGTM (noting that one tiny stray ".")

RByers:
Thanks for the review @patrickhlauke.  I finally made time to [take a look](https://rawgit.com/mustaqahmed/pointerevents/fix-button-desc/index.html#button-states) and it looks great to me (except for that trailing "." - otherwise I'd merge).

I'm out for a week and a half, Patrick can you please merge this once that last issue is fixed?

mustaqahmed:
Done moving the '.' to the para before.

mustaqahmed:
Fixed the second '.' outside the para, yikes! Sorry for the oversight!

patrickhlauke:
Lovely, thank you :+1:

PR #36, opened 2016-02-26
Add the missing pointerover/enter events

NavidZ:
In "Process Pending Pointer Capture" section, likewise the
pointerout/leave that is mentioned in this section in some
cases we need to send pointerover/enter when the capture
is lost and pointer is on another node that is not the
capturing node.

### DIFF
--- a/index.html
+++ b/index.html
@@ -421,7 +421,11 @@ these pointers will all produce <a>compatibility mouse events</a>.</div>
             <h4>Process Pending Pointer Capture</h4>
             <p>Whenever a user agent is to fire a Pointer Event that is not <code>gotpointercapture</code> or <code>lostpointercapture</code>, it must first run these steps:</p>
             <ol>
-                <li>If the <a>pointer capture target override</a> for this pointer is set and is not equal to the <a>pending pointer capture target override</a>, then fire a pointer event named <code>lostpointercapture</code> at the <a>pointer capture target override</a> node.</li>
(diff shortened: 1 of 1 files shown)


RByers:
Thanks Navid, this looks reasonable to me!  But this space is pretty complicated and I'm not an expert, so let's see if someone from Edge can also review the precise wording before we land it (eg. @teddink?).   

Presumably this was also a test hole in the tests, right?  Can you submit an update for that as well?

NavidZ:
I haven't looked at the tests for the pointer capturing coverage yet. I will keep this in mind when I'm reviewing and testing those on Chrome to add anything that's missing.

RByers:
Oh right, of course - we're starting capture implementation now so haven't tried to run those tests yet.  I knew that, sorry.

teddink:
The change sounds good to me as well, but like Rick, I am not yet an expert in the area so I am reaching out to folks internally like @jacobrossi

mustaqahmed:
One quick suggestion while we are into it: it seems to me that the section "Process Pending Pointer Capture" should really belong to the "Pointer Capture" section, perhaps between Sec10.2 & Sec10.3, because it is impossible to follow this section w/o skimming through 10.1 & 10.2 first. Only thing we need in the current section (Sec5.2.1) is a reference to the fact that we need to process pending capture before firing a "non-capture" PointerEvent.

jacobrossi:
LGTM and +1 to adding test coverage for this.  Regarding moving this section, I'm supportive but let's open a separate issue for that.

RByers:
Thanks everyone!

Issue #39, opened 2016-03-08
Incorrect order of the events in process pending pointer capture section


NavidZ:
@RByers, @mustaqahmed, @jacobrossi 

Looking at this section:

https://w3c.github.io/pointerevents/#process-pending-pointer-capture
- It seems weird that firing pointerleave/out happens after firing gotpointercapture. As the other places in the spec mentioned, if a pointer is captured by some element no other element should get those boundary transition events. The same thing happens for pointerenter/over.
- Also the conditions for when to send the out/leave/over/enter events only covers when the pointer is being captured from an un-captured state or vice versa. Meaning that at least one of the pending pointer capture target or pointer capture target should be clear for those transition events to be fired. I can imagine of some cases that this doesn't sound right when the element which has the capture set another element as the pointer capture.

This example shows both of the cases above:
https://output.jsbin.com/riyotu
1. Move the mouse onto the green div and press and keep the left key down.
2. Move the mouse a little bit in the green box (until you get more than ~10 pointermove) which then the pointer capture will be set to the blue div.
3. Move the mouse a little more until the blue div gets at least one pointer move.

Here is the list of events Edge prints:

1 green received pointermove 1
2 green received pointerover 1
3 green received pointerenter 1
4 green received pointerdown 1
======= Set Pointer Capture ======= 
5 green received gotpointercapture 1
6 green received pointermove 1
7 green received pointermove 1
8 green received pointermove 1
9 green received pointermove 1
10 green received pointermove 1
11 green received pointermove 1
12 green received pointermove 1
13 green received pointermove 1
14 green received pointermove 1
15 green received pointermove 1
16 green received pointermove 1
17 green received pointermove 1
======= Re-set Pointer Capture ======= 
18 green received lostpointercapture 1
19 blue received gotpointercapture 1
20 green received pointerout 1
21 green received pointerleave 1
22 blue received pointermove 1

Note that Edge does send pointerout/leave in steps 20 and 21 while neither of pending pointer capture target or pointer capture target were clear (i.e. pointer capture target was green div and pending pointer capture target was blue div) which spec doesn't cover in the conditions. Also I expected events in steps 20 and 21 happens at least before step 19 as per other parts of the spec. Further it feels more natural if they happen even before 18 but I don't have any arguments for this one.

p.s.: Note that step 1 also seems incorrect as pointermove happens before pointerover/enter but that is not part of this issue and it is probably just a bug in Edge.

mustaqahmed:
[...]

PR #40, opened 2016-03-08
Reorder the events in process pending pointer capture

NavidZ:
Following #39,
move the pointerleave/out events before the gotpointercapture
and also relax the condition for sending those as the current spec
does not consider the case that the capture is transferred from a
node to another.

### DIFF
--- a/index.html
+++ b/index.html
@@ -421,15 +421,13 @@ these pointers will all produce <a>compatibility mouse events</a>.</div>
             <h4>Process Pending Pointer Capture</h4>
             <p>Whenever a user agent is to fire a Pointer Event that is not <code>gotpointercapture</code> or <code>lostpointercapture</code>, it must first run these steps:</p>
             <ol>
-                <li>If the <a>pointer capture target override</a> for this pointer is set and is not equal to the <a>pending pointer capture target override</a>, then fire a pointer event named <code>lostpointercapture</code> at the <a>pointer capture target override</a> node.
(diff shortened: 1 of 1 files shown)


RByers:
This looks good to me and seems like a logical simplification.  This is still quite complicated, so I'd love to see some tests to go along with this change (but we don't necessarily need to block updating the spec).

patrickhlauke:
What was the status of this? Did we agree this was good to merge? If so, it'll need a rebase first

NavidZ:
No. I believe this is one of those issues that depends on whether we send boundary events while capturing or not. Do you want me to close this for now or is it better to leave it open?

patrickhlauke:
happy to leave it open for now until we figure out the bigger picture :)

mingchou:
About the last 'if' of the modifications, in the case that the pointer capture target override is not set and set pending pointer capture to any node other than the hit test node, shall we still fire pointerover and pointerenter to the hit test node?

NavidZ:
I believe we will get rid of all boundary events line in "Process Pending Pointer Capture" section as the cleaner way of defining it would be to change the hit-testing in that case and the boundary events will follow the UI Events spec. So I need to get rid of this PR and upload another one for that proposal.

NavidZ:
With Chrome going forward with change in the definition of hit-testing while a pointer is captured, I will close this PR and create another one which includes that proposal in one PR.

PR #43, opened 2016-03-14
clarification about dynamic touch-action changes

### LINKED ISSUE #42: clarify what happens if touch-action is changed dynamically
It's probably implicit in the prose for "9.1 The touch-action CSS property", but: what should happen if, on `pointerdown`, the `touch-action` for the element is changed dynamically (for instance, from `auto` to `none`, based on the document's scroll position, for a naive pull-to-refresh implementation)? I assume that at that point, it's already too late and the user agent determined what to do (whether to handle the pointer/gesture or not) based on what `touch-action` was the moment when `pointerdown` was fired...if so, would it be useful to clarify this in a note?

patrickhlauke:
Closes #42


RByers:
Looks good, thanks!

PR #44, opened 2016-03-14
Modify/clarify the example for directional touch-action values

### LINKED ISSUE #41: touch-action - incorrect values in note for pull-to-refresh?
Currently in "9.1 The touch-action CSS property" we have

> For example, to implement a simple pull-to-refresh effect the document's `touch-action` can be set to `pan-x pan-down` whenever the scroll position is 0 and `pan-x pan-y` otherwise. This allows pointer event handlers to define the behavior for upward scrolls that start from the top of the document.

Shouldn't the first value (for scroll position 0) be `pan-x pan-up`, as you'd want to handle the down movement yourself for the pull to refresh?

patrickhlauke:
Closes #41


RByers:
Thanks

patrickhlauke:
:+1:

Issue #48, opened 2016-03-24
What happens when an isPrimary pointer is removed?


patrickhlauke:
Currently the spec does not explain what would happen if an isPrimary pointer is removed.
The scenario would be: imagine more than one touch pointer in the document, one of which isPrimary; the user lifts the primary finger, and that pointer disappears. Should the touch that came after the isPrimary one be promoted to now being the new primary?
My hunch would be that no, as at the start of the gesture/interaction, no new primary should be promoted/elected. This is not, however, made explicit (or even implicit) in the spec to date.

NavidZ:
I agree that there is no explicit statement about this. However, this section
https://w3c.github.io/pointerevents/#dfn-the-primary-pointer

implies that for example for touch:
A pointer representing touch input is considered the primary touch input if its pointerdown event was dispatched when no other active pointers representing touch input existed.

which I guess prevents the promotion of any existing event to new primary. Also the last note in this section tries to bring up the case that user agent can fire pointer events in which no pointer is marked as a primary pointer. Maybe we can reword that note to add the scenario you are mentioning as well?

PR #50, opened 2016-03-24
Cover the case when primary pointer is removed

NavidZ:
To address #48:

Reword a note in primary pointer section to also cover the
case of primary pointer being removed under the scenarios
that user agent may fire pointer events with no primary one.


scottgonzalez:
Looks good.

patrickhlauke:
LGTM. Rick/Jacob?

index.html
```diff
@@ -375,7 +375,7 @@
...
+				<div class="note">In some cases, it is possible for the user agent to fire pointer events in which no pointer is marked as a primary pointer. For instance, when there are multiple active pointers of a particular type like multi-touch and the primary pointer is removed (e.g. it leaves the screen). Also on some platforms, the primary pointer is determined using all active pointers on the device, including those not targeted at the user agent (e.g. in another application). For example, if the first touch interaction is targeted outside the user agent and a secondary (multi-touch) touch interaction is targeted inside the user agent, then the user agent may fire pointer events for the second contact with a value of <code>false</code> for <code>isPrimary</code>.</div>
```
mustaqahmed: - "For instance, when there are multiple...screen), **there will be no primary pointers**."
- I think the use of "for example" within one case of "for instance" should be avoided for clarity, otherwise the second sentence ("For example, ...") in the second example can possibly appear as the overall third example. Should we use bullets? Or does the following sound clearer?
  "... Also on the platforms where the primary pointer is determined using all active pointers on the device (including those targeted at an application other than the user agent), if the first touch interaction is ..."

mustaqahmed:
LGTM.

RByers:
Sorry for the delay, looks good to me too - thanks!

PR #53, opened 2016-04-22
Rewrite of primary pointer section

### LINKED ISSUE #49: Determining the primary pointer - note about mice?
Currently https://w3c.github.io/pointerevents/#determining-the-primary-pointer makes an implicit assumption that there will never be more than one concurrent mouse input

> a pointer is considered primary if: [...] The pointer represents a mouse device.

This of course reflects the current reality where no OS allows for 2 or more mice to be active (in most cases, it seems the OS simply "combines" more than one mouse/trackpad - any relative movement on those multiple mice is simply applied to the one mouse pointer, buttons are treated the same across all mice [though i've not tested if chording works across multiple mice, i.e. if i press right mouse button on one and left mouse button on the other if they're then counted as a chorded left-and-right mouse button press]).

Is it worth adding an informative note about this somewhere in 5.1.2.1? Just so that the above assumption and explanation of why there's only ever one mouse and it will always be isPrimary? Further, do we think there's ever going to be the possibility that a device/OS will allow multiple concurrent mice, and if so should we generalise the bullet list to determine the primary mouse, but then note that in practice there's only ever one mouse pointer?

patrickhlauke:
This generalises the treatment of mouse inputs to allow for the possibility of systems that support more than one simultaneous and independent mouse device, while at the same time clarifying that currently most (all?) operating systems/user agents don't have a concept of multiple mouse pointers. 

It future-proofs the spec while also reflecting reality.

Closes #49

### DIFF
--- a/index.html
+++ b/index.html
@@ -359,28 +359,16 @@ eventTarget.dispatchEvent(event);
 		</section>
 		<section>
 			<h2><dfn>The Primary Pointer</dfn></h2>
-			<div><p>In a multi-pointer (e.g. multi-touch) scenario, the <a href="#widl-PointerEvent-isPrimary"><code>isPrimary</code></a> property is used to identify a master pointer amongst the set of active pointers for each pointer type.  Only a primary pointer will produce <a>compatibility mouse events</a>. Authors who desire single-pointer interaction can achieve this by ignoring non-primary pointers (however, see the note below on <a href="#multiple-primary-pointers">multiple primary pointers</a>).</p></div>
-			<section>
-				<h3>Determining the primary pointer</h3>
-				<div>When firing a pointer event, a pointer is considered primary if:
-					<ul>
-						<li>The pointer represents a primary mouse input.</li>
-						<li>The pointer represents a primary touch input.</li>
-						<li>The pointer represents a primary pen input.</li>
-					</ul>
-				</div>
-				<dl>
-					<dt>primary mouse input</dt><dd>A pointer representing mouse input is considered the <i>primary mouse input</i> if it dispatched any pointer events (such as <code>pointerdown</code> or <code>pointermove</code>) when no other active pointers representing a mouse input existed. However, see the note about <a href="#multiple-mouse-inputs">multiple mouse inputs</a>.</dd>
-					<dt>primary touch input</dt><dd>A pointer representing touch input is considered the <i>primary touch input</i> if its <code>pointerdown</code> event was dispatched when no other active pointers representing touch input existed.</dd>
-					<dt>primary pen input</dt><dd>A pointer representing pen input is considered the <i>primary pen input</i> if its <code>pointerdown</code> event was dispatched when no other active pointers representing pen input existed.</dd>
-				</dl>
-				<div class="note" id="multiple-mouse-inputs">Current operating systems and user agents don't usually have a concept of multiple mouse inputs. When more than one mouse device is present (for instance, on a laptop with both a trackpad and an external mouse), all mouse devices are generally treated as a single device - movements on any of the devices are translated to movement of a single mouse pointer, and there is no distinction between button presses on different mouse devices. For this reason, there will usually only be a single mouse pointer, and that pointer will be primary.</div>
-				<div class="note" id="multiple-primary-pointers">When two or more pointer device types are being used concurrently, multiple pointers (one for each <code>pointerType</code>) are considered primary. For example, a touch contact and a mouse cursor moved simultaneously will produce pointers that are both considered primary.</div>
-				<div class="note">In the case where there are multiple <a href="#the-primary-pointer">primary pointers</a>, 
-these pointers will all produce <a>compatibility mouse events</a>.</div>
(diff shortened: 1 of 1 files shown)


RByers:
This LGTM.  I also read over the compatibility mouse events section to ensure it was correct given these changes (definitely can't fire any mouse events for these weird non-primary mouse pointers), and it looks good.

RByers:
Just commenting on the precise wording though - I think we may want to make a couple tweaks...

index.html
```diff
@@ -364,15 +364,17 @@
 				<h3>Determining the primary pointer</h3>
 				<div>When firing a pointer event, a pointer is considered primary if:
 					<ul>
-						<li>The pointer represents a mouse device.</li>
+						<li>The pointer represents a primary mouse input.</li>
```
RByers: Seems like we can simplify this whole section now that the mouse, touch and stylus paragraphs are mostly redundant.  See below.

index.html
```diff
@@ -364,15 +364,17 @@
...
 						<li>The pointer represents a primary touch input.</li>
 						<li>The pointer represents a primary pen input.</li>
 					</ul>
 				</div>
 				<dl>
+					<dt>primary mouse input</dt><dd>A pointer representing mouse input is considered the <i>primary mouse input</i> if it dispatched any pointer events (such as <code>pointerdown</code> or <code>pointermove</code>) when no other active pointers representing a mouse input existed. However, see the note about <a href="#multiple-mouse-inputs">multiple mouse inputs</a>.</dd>
```
RByers: I think this (and the existing touch/pen sections below when considering hover-capable pointers) isn't quite right.  Eg. consider the case of two mice (or pens) both hovering without either being active.  One needs to be primary, right?  What happens in Edge today when a pen first becomes visibly by the digitizer - does it have `isPrimary=true` from the first `pointermove`, or only after the first `pointerdown` event?

I think the essential properties are that:
- At any given time exactly one pointer of each type is considered primary
- When the first pointer of a given type becomes active, that pointer also becomes the primary pointer

patrickhlauke:
@RByers 

> What happens in Edge today when a pen first becomes visibly by the digitizer - does it have isPrimary=true from the first pointermove, or only after the first pointerdown event?

testing on Surface 3 with Pen and Edge, the pen is primary as soon as it's recognized/hovering. so quite right, i think this isn't correct - but that's a separate issue (which i'll file later today)

patrickhlauke:
> When the first pointer of a given type becomes active, that pointer also becomes the primary pointer

agree - perhaps with link/pointer to the glossary definition of "active pointer" - this is the simplest amendment (which then can help simplify the entire prose for this section, i may actually work on this as one combined PR, but open a separate issue for clarity)

patrickhlauke:
Pushed a separate commit to address both #54 and #49 - drastically simplifies the definition and brings it in line with reality of implementation.

/cc @jacobrossi @RByers

index.html
```diff
@@ -359,28 +359,17 @@
...
-				
-			</section>
+			<div><p>In a multi-pointer (e.g. multi-touch) scenario, the <a href="#widl-PointerEvent-isPrimary"><code>isPrimary</code></a> property is used to identify a master pointer amongst the set of <a data-lt="active pointer">active pointers</a> for each pointer type.</p></div>
+			<ul>
+				<li>At any given time, there can only ever be exactly one primary pointer for each pointer type.</li>
```
mustaqahmed: Does
  "...there can be _at most one_ primary pointer..."
make it clearer? The phrase "can ever be exactly one" seems to exclude the following cases of no primary "touch" pointer:
- when there's no active touches, and
- when two fingers touch the screen in sequence, then the first one is removed.

index.html
```diff
@@ -359,28 +359,17 @@
...
+				<li>The first pointer to become active for a particular pointer type (e.g. the first finger to touch the screen in a multi-touch interaction) becomes the primary pointer for that pointer type.</li>
+				<li>Only a primary pointer will produce <a>compatibility mouse events</a>.</li>
+			</ul>
+			<div class="note">Authors who desire single-pointer interaction can achieve this by ignoring non-primary pointers (however, see the note below on <a href="#multiple-primary-pointers">multiple primary pointers</a>).</p></div>
```
mustaqahmed, line 368: Nit: Is there a way to avoid 5 Notes in a row? E.g. by combining 1 & 2, and/or dropping 3?

index.html
patrickhlauke: ah, good call. yes, that would make it clearer and more correct. i'll push an update

index.html
patrickhlauke, line 368: 3 could be added part of the second bullet above ("Only a primary pointer will product..."). that should lighten this up a bit. 1 & 2 are not directly related (one is about multiple pointers of the same type, the other about different pointer types and their primary pointers), so i'd leave them separate.

index.html
```diff
@@ -359,28 +359,16 @@
...
-				
-			</section>
+			<div><p>In a multi-pointer (e.g. multi-touch) scenario, the <a href="#widl-PointerEvent-isPrimary"><code>isPrimary</code></a> property is used to identify a master pointer amongst the set of <a data-lt="active pointer">active pointers</a> for each pointer type.</p></div>
+			<ul>
+				<li>At any given time, there can only ever be at most one primary pointer for each pointer type.</li>
+				<li>The first pointer to become active for a particular pointer type (e.g. the first finger to touch the screen in a multi-touch interaction) becomes the primary pointer for that pointer type.</li>
+				<li>Only a primary pointer will produce <a>compatibility mouse events</a>. In the case where there are multiple <a href="#the-primary-pointer">primary pointers</a>, these pointers will all produce <a>compatibility mouse events</a></li>
```
mustaqahmed: We are missing a '.' at the end.

mustaqahmed:
LGTM, thanks Patrick.

index.html
patrickhlauke: good spot. fixed

RByers:
LGTM, thanks!

RByers:
I squashed your two commits and merged them.  I hope that's what you intended (otherwise I can easily undo it).

patrickhlauke:
Yes i was relying on GitHub's new autosquashing feature :)

Issue #35, opened 2016-02-23
Compatibility mouse transition events should reflect a single logical mouse pointer


NavidZ:
@RByers @mustaqahmed 

Looking at this example:

https://output.jsbin.com/bupaluf

Step 1. Move the mouse into the green box and leave it there.
Step 2. Touch the blue box and release.
Step 3. Now slightly move the mouse.

Here is the sequence in case it is hard to test it on Edge:

---- Step 1 ----
pointerover (id=1) on green 
mouseover on green 
pointerenter (id=1) on green 
mouseenter on green 
---- Step 2 ----
mouseout on green 
pointerleave (id=134) on green 
mouseleave on green 
pointerover (id=134) on blue 
mouseover on blue 
pointerenter (id=134) on blue 
mouseenter on blue 
pointerout (id=134) on blue 
mouseout on blue 
pointerleave (id=134) on blue 
mouseleave on blue 
---- Step 3 ----
mouseover on green 
pointerenter (id=1) on green 
mouseenter on green 

This sequence of events seem incorrect to me. Particularly looking at pointerevent with id=1 you can see the inconsistent sequence of events for it (i.e. 2 enters for green). Also mouse events (that are actually fired by moving mouse device) do no fully match the pointerevents with id=1 (i.e. the last 3 events).

The problem I see here is not quite addressed in the spec either. So I was wondering if something is missing or I have misunderstood something. 

If there is actually a problem here, I can imagine the following solutions (which the first one looks cleaner to me but I don't know the impact).
1. Do not try to match pointerevents with mouse events at all. So basically we left mouse in the green box and we touched blue box and fired compatibility mouse events for the touch event. Now if we move the mouse since the pointerevent with id=1 is still in the green box we shouldn't send any transition events (i.e. pointerenter/over) however since the mouse (i.e. the compatibility mouse which might include the actual mouse, touch, or pen) is in the blue box we should now send the mouse transition events to both blue box (mouseout/leave) and green box (mouseover/enter). This solution breaks the consistency of pointerevent with id=1 with mouse events but tells a consistent story looking only at pointer events or only looking at the mouse events.
2. The second solution is when we touch somewhere and we fire compatibility mouse event we really move the mouse to there. So basically not only send the pointerevents with id=#RandomNumber for the touch event itself but also send the pointerevents with id=1 for the mouse. This solution makes sure the consistency of pointerevent with id=1 and the mouse events but in the cases like pen or touch events sends double pointer events which might not be desirable.

[...]

PR #56, opened 2016-04-28
Made mouseover/out/enter/leave event firing independent of corresponding PEs

mustaqahmed:
The spec currently suggests that compatibility mouse transition events (i.e. mouseover, mouseout, mouseenter and mouseleave) should be fired right after firing the corresponding PointerEvents, implying a 1:1 correspondence between the PointerEvents vs the compatibility MouseEvents. This 1:1 correspondence breaks the consistency of MouseEvent sequence itself when two or more primary pointers (of different types) are active at the same time. For example, mouseover events can be sent to two different targets without a mouseout in-between, which can possibly break the UIs based on legacy MouseEvents, e.g., by opening two menu items at the same time. (Issue #35.)

This patch fixes the problem by omitting the 1:1 correspondence altogether. More precisely, it suggests to fire the compatibility mouse transition events right before a pointerdown/up/move event.

In addition, this patch modifies the default actions of pointerover, pointerout, pointerenter and pointerleave events to "none". This keeps the spec effectively unchanged since the original default action was "the default action of corresponding MouseEvent" but the UI Event Spec defines those default actions as "none".


index.html
```diff
@@ -450,14 +450,14 @@
 						<td>Sync</td>
 						<td>Yes</td>
 						<td>Yes</td>
-						<td>Varies:  when the pointer is primary, all default actions of <code>mouseover</code></td>
+						<td>None</td>
 					</tr>
```
RByers, line 443: Please call out these changes in the description also (since they're only tangentially related).  I.e. that since UI Events says these mouse events have no default action, we can just say the same for the pointer events.

index.html
```diff
@@ -819,20 +819,26 @@
...
+			<div>
+			<p>While only <a href="#the-primary-pointer">primary pointers</a> can produce compatibility mouse events, <a href="#multiple-primary-pointers">multiple primary pointers</a> can be active simultaneously, each producing its own stream of compatibility mouse events.  Since all these compatibility events would appear to a legacy mouse-specific code to be coming from a single mouse hardware, user agents are encouraged to guarantee that the compatibility mouse events are consistent from a single device perspective.  For mouse transition events (i.e., <code>mouseover</code>, <code>mouseout</code>, <code>mouseenter</code> and <code>mouseleave</code>), this means the entry/exit state for every event target is valid as implied by [<cite><a class="bibref" href="#bib-DOM-LEVEL-3-EVENTS">DOM-LEVEL-3-EVENTS</a></cite>].  Users agents MAY guarantee this by maintaining the <dfn id="dfn-legacy-mouse-pos">effective position of legacy mouse pointer</dfn> in the document as follows.</p>
+			<p>Right before firing a <code>pointerdown</code>, <code>pointerup</code> or <code>pointermove</code> event, the user agent MAY run the following steps:</p>
```
RByers: The way to officially reference another spec normatively is `[[!DOM-LEVEL-3-EVENTS]]` - see the [ReSpec guide](https://www.w3.org/respec/guide.html#references).

However "DOM-LEVEL-3" is the old name for that spec and so kinda confusing now.  Would you mind updating all the various references to it to "UIEVENTS" instead?  Probably best to do that in a separate CL to avoid polluting this one if you don't mind.

index.html
```diff
@@ -819,20 +819,26 @@
...
+			<ol>
+				<li>Let <code>T</code> be the target of the event above.</li>
```
RByers: nit: "target of the pointer event being dispatched"?

index.html
```diff
@@ -819,20 +819,26 @@
 		</section>
 		</div>
 		<section>
+			<h2 id="tracking-legacy-mouse-pos">Tracking the effective position of legacy mouse pointer</h2>
+			<div>
+			<p>While only <a href="#the-primary-pointer">primary pointers</a> can produce compatibility mouse events, <a href="#multiple-primary-pointers">multiple primary pointers</a> can be active simultaneously, each producing its own stream of compatibility mouse events.  Since all these compatibility events would appear to a legacy mouse-specific code to be coming from a single mouse hardware, user agents are encouraged to guarantee that the compatibility mouse events are consistent from a single device perspective.  For mouse transition events (i.e., <code>mouseover</code>, <code>mouseout</code>, <code>mouseenter</code> and <code>mouseleave</code>), this means the entry/exit state for every event target is valid as implied by [<cite><a class="bibref" href="#bib-DOM-LEVEL-3-EVENTS">DOM-LEVEL-3-EVENTS</a></cite>].  Users agents MAY guarantee this by maintaining the <dfn id="dfn-legacy-mouse-pos">effective position of legacy mouse pointer</dfn> in the document as follows.</p>
```
RByers: Please use ReSpec's [definition and linking support](https://www.w3.org/respec/guide.html#definitions-and-linking) rather than writing your own `a href` links.

RByers: We should probably be using "SHOULD" instead of "MAY" here.  The section is optional, but if it's being used then it should be followed exactly.

index.html
```diff
@@ -819,20 +819,26 @@
 		</section>
 		</div>
 		<section>
+			<h2 id="tracking-legacy-mouse-pos">Tracking the effective position of legacy mouse pointer</h2>
```
RByers: nit: "of the legacy"

index.html
RByers: nit: s/"a legacy mouse-specific code"/"MouseEvent handling code"

RByers: nit: s/"a single mouse hardware"/"a single mouse device"

index.html
```diff
@@ -819,20 +819,26 @@
...
+			<p>Right before firing a <code>pointerdown</code>, <code>pointerup</code> or <code>pointermove</code> event, the user agent MAY run the following steps:</p>
+			<ol>
+				<li>Let <code>T</code> be the target of the event above.</li>
+				<li>If <code>T</code> is equal to current <a href="#dfn-legacy-mouse-pos" class="internalDFN">effective legacy mouse pointer position</a>, terminate these steps.</li>
```
RByers: This algorithm doesn't define what happens at the beginning (when the legacy-mouse-pos has not yet been defined).  I'm not sure exactly what do do about this - UI Events certainly doesn't seem very precise on this point to me.  Thoughts?

index.html
```diff
@@ -819,20 +819,26 @@
...
+				<li>Dispatch the same sequence of <code>mouseover</code>, <code>mouseout</code>, <code>mouseenter</code> and <code>mouseleave</code> events that would be dispatched as per [<cite><a class="bibref" href="#bib-DOM-LEVEL-3-EVENTS">DOM-LEVEL-3-EVENTS</a></cite>] if a (legacy) mouse pointer had moved from current <a href="#dfn-legacy-mouse-pos" class="internalDFN">effective legacy mouse pointer position</a> directly to <code>T</code>.</li>
```
RByers: nit: maybe rather than "dispatch the same sequence ... that would be dispatched ..." we can simplify this to: "dispatch ... events as per ... for a mouse moving moving from the legacy mouse position to T"?

RByers:
The basic approach here seems good to me.  @teddink any concerns?  I think this is closer (but still not exactly) to what Edge is doing today and makes more conceptual sense IMHO.

index.html
mustaqahmed, line 443: Done.

index.html
mustaqahmed: Done.

index.html
mustaqahmed: All done. Used "MouseEvent code" instead of "MouseEvent handling code".

index.html
mustaqahmed: Fixed my links. Filed #60 for the reference label change.

index.html
mustaqahmed: Done.

index.html
mustaqahmed: I was originally planning to say that it starts with a NULL value, meaning outside the document/window. This would be helpful to handle the pointer exit case too, by adding this:
"If event to be fired is a pointerleave at document/window, let T be NULL",
then add a few lines below to handle the NULL case.

Finally I thought: may be it is an implementation details and/or understood concept.

Do you find such details helpful?

index.html
mustaqahmed: Done.

index.html
```diff
@@ -819,20 +819,26 @@
 		</section>
 		</div>
 		<section>
+			<h2 id="tracking-legacy-mouse-pos">Tracking the effective position of the legacy mouse pointer</h2>
+			<div>
+			<p>While only <a href="#the-primary-pointer">primary pointers</a> can produce compatibility mouse events, <a href="#multiple-primary-pointers">multiple primary pointers</a> can be active simultaneously, each producing its own stream of compatibility mouse events.  Since all these compatibility events would appear to MouseEvent code to be coming from a single mouse device, user agents are encouraged to guarantee that the compatibility mouse events are consistent from a single device perspective.  For mouse transition events (i.e., <code>mouseover</code>, <code>mouseout</code>, <code>mouseenter</code> and <code>mouseleave</code>), this means the entry/exit state for every event target is valid as implied by [[!DOM-LEVEL-3-EVENTS]].  Users agents SHOULD guarantee this by maintaining the <dfn id="dfn-legacy-mouse-pos">effective position of the legacy mouse pointer</dfn> in the document as follows.</p>
```
RByers: sorry another for "each producing its own stream of compatibility mouse events" I'd suggest removing the "stream of".  "own stream" sounds like they're really independent streams.  The whole point of this section is to say that, although each device produces compat mouse events, it's a single logical stream.

index.html
```diff
@@ -819,20 +819,26 @@
...
+			<p>Right before firing a <code>pointerdown</code>, <code>pointerup</code> or <code>pointermove</code> event, the user agent MAY run the following steps:</p>
```
RByers: s/MAY/SHOULD/

index.html
RByers: Again you should avoid use any `href`s in your internal `<a>` links.  ReSpec handles this automatically for you.  For example see the various other places in the spec which do simply `<a>primary pointer</a>`.  When necessary you can use `data-lt` on the `a` tag to specify a definition name that's different from the anchor content (eg. see examples like `<a data-lt="Chorded Button Interactions">`).  I don't know why this is really any better than using `<a href=...>` but we should at least be consistent across the spec.

index.html
RByers: The way I see it there are two styles of spec writing - algorithm heavy and prose heavy.  The WhatWG DOM spec is algorithm heavy - very precisely defines what is supposed to happen but can be harder to follow.  The UI Events spec is prose heavy - you get the basic idea pretty easily but there can be ambiguity in the edge cases (though if that's resolved by precise tests then it should be fine IMHO).  I like that you're trying to define an algorithm here, but it's hard to do well when you're referring to stuff that's defined in a more handy-wavy prose style elsewhere.

So if we want to be consistent with the UI Event spec treatment, we could just remove the algorithm and definition and say (somewhat vaguely) that the transition events follow from the UI Events definition corresponding to a single mouse being moved around.

Or if you'd like to try to make it more precise, then yes I think you need some text around the default value, and unsetting it on pointerleave from the window etc.  That's a really going point actually - as written now it looks like it's illegal to send a mouseleave event when the mouse leaves the window (because there is no corresponding `pointermove` sent).  If you want to keep the algorithm style you'll need to expand the usage of this section to handle that special case.  Note that specs usually use the simpler concept of "is set" and "unset" rather than using a programming value like "NULL" (these aren't real variables afterall).

RByers: Oh and are you planning on writing some web-platform-tests for this?  I think that's the most valuable part to really making what you're describing concrete.  Eg. will we test anything about the interleaving of pointer and mouse transition events, or leave some flexibility there?  Eg. when entering a new element, are both of these legal event orderings or do you intend to require the latter?:
1. pointerover, mouseover, pointerenter, mouseenter, pointermove, mousemove
2. pointerover, pointerenter, mouseover, mouseenter, pointermove, mousemove

index.html
mustaqahmed: All done.

index.html
```diff
@@ -374,7 +374,7 @@
 					<dt>primary pen input</dt><dd>A pointer representing pen input is considered the <i>primary pen input</i> if its <code>pointerdown</code> event was dispatched when no other active pointers representing pen input existed.</dd>
 				</dl>
 				<div class="note" id="multiple-primary-pointers">When two or more pointer device types are being used concurrently, multiple pointers (one for each <code>pointerType</code>) are considered primary. For example, a touch contact and a mouse cursor moved simultaneously will produce pointers that are both considered primary.</div>
-				<div class="note">In the case where there are multiple <a href="#the-primary-pointer">primary pointers</a>, 
+				<div class="note">In the case where there are multiple <a data-lt="The Primary Pointer">primary pointers</a>,
```
RByers: Oh, it looks like the uses of `<a>primary pointer</a>` I was looking at are actually broken!  I think that's just because `<dfn>The Primary Pointer</dfn>` should be `The <dfn>Primary Pointer</dfn>`.  If you change that, then I think you can avoid the `data-lt` here and elsewhere.

RByers:
On [the call](https://www.w3.org/2016/05/04-pointerevents-minutes.html#item02) we agreed this was nearly good to merge, just blocked on some editorial feedback from @patrickhlauke.

patrickhlauke:
wondering if the entire paragraph

> The compatibility mapping with mouse events are an _optional_ feature of this specification. User agents are encouraged to support the feature for best compatibility with existing legacy content. User agents that do not support compatibility mouse events are still encouraged to support the `click` and `contextmenu` events (see the note below).

should be moved to the very start of the section, right after the "Compatibility Mapping with Mouse Events" heading. this would make it a bit more immediately clear right from the start when a reader delves into the section that this is optional, rather than burying the lede in between two notes.

Perhaps even wrapped as a note (though this would make it non-normative?) or some other form of styling change/block-out if available.

mustaqahmed:
Done.

patrickhlauke:
Sorry, you're right @mustaqahmed ... your update slipped me by.

At the risk of sounding nitpicky, I'd say that now the rationale for why there's even compatibility mapping is a bit buried. I'd suggest ripping out the first paragraph or the rationale and grafting it right at the start, so something like:

```
<h1><dfn lt="compatibility mouse events">Compatibility Mapping with Mouse Events</dfn></h1>
<p>The vast majority of web content existing today codes only to Mouse Events. The following describes an algorithm for how a user agent MAY map generic pointer input to mouse events for compatibility with this content. The compatibility mapping with mouse events are an OPTIONAL...
```

and then leaving the rest as you have it.

mustaqahmed:
ptal. Thanks for the suggestion, the rationale is clearer now.

patrickhlauke:
LGTM (once merge conflicts are resolved)

mustaqahmed:
Thanks Patrick. I will resolve the merge conflict after adding the boundary condition handling (see @RByers's comments on May 3).

mustaqahmed:
@RByers, I have updated the algorithm in Sec11.1 to handle unset values correctly.

RByers:
On today's call @teddink said it was their intention to have Edge's behavior match this PR but it'll take them some time to try coding it up and understand the full impact in practice.

RByers:
Sorry for the delay (fell off my notifications list somehow), looks good to me.  As discussed in the call today, let's merge (we can always continue to tweak the details as needed of course).

PR #57, opened 2016-04-28
Adding the scope of unique pointerId

### LINKED ISSUE #52: Pointer Id across different iframes
There is no indication of whether the user agent is allowed to use different ids for the same pointer across multiple iframe or not.
It seems that what Edge uses as pointer id somewhat comes from the OS it is kind of unique.

I was wondering if that was the intention and if so is it worth to add a note to make sure the id of the same pointer is preserved across different iframes?

NavidZ:
Defining the "top-level browsing context" as the scope of
pointerIds which closes #52 

I was not sure whether it is okay to add a link to another draft spec. Or maybe it is better not to have a link for the definition of top-level browsing context at all.

### DIFF
--- a/index.html
+++ b/index.html
@@ -213,7 +213,7 @@ eventTarget.dispatchEvent(event);
 						<li>If a touch contact or pen stylus is lifted beyond the range of the digitizer, then it is no longer considered active.</li>
 					</ul>
 					<div class="note">On some platforms, the set of active pointers includes all pointer input to the device, including any that are not targeted at the user agent (e.g. those targeted at other applications).</div>
-					<div class="note">Each active pointer should have the same id in the scope of at least the <a href="https://www.w3.org/TR/html51/browsers.html#top-level-browsing-context">top-level browsing context</a> (as defined by [[!HTML5]]).</div>
+					<div class="note">Each active pointer should have the same id in the scope of at least the <a href="https://www.w3.org/TR/html5/browsers.html#top-level-browsing-context">top-level browsing context</a> (as defined by [[!HTML5]]).</div>
 				</dd>
 			<dt><dfn>canceled event</dfn></dt>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -238,7 +239,7 @@
 			<dl class="idl" title="[Constructor(DOMString type, optional PointerEventInit eventInitDict)] interface PointerEvent : MouseEvent" data-merge="PointerEventInit">
 				<dt>readonly attribute long pointerId</dt>
 				<dd>
-					<p>A unique identifier for the pointer causing the event. This identifier MUST be unique from all other <a data-lt="active pointer">active pointers</a> at the time. A user agent MAY recycle previously retired values for <code>pointerId</code> from previous active pointers, if necessary.</p>
+					<p>A unique identifier for the pointer causing the event. This identifier MUST be unique from all other <a data-lt="active pointer">active pointers</a> in the <a href="https://www.w3.org/TR/html51/browsers.html#top-level-browsing-context">top-level browsing context</a> at the time. A user agent MAY recycle previously retired values for <code>pointerId</code> from previous active pointers, if necessary.</p>
```
RByers, line 243: In order to refer to another spec I believe we're supposed to add an entry to the "references" section (which we normally do with the `[[!HTML5]]` syntax.  We refer to W3C HTML5 all over the place, but not W3C HTML5.1.  I know I said it was fine to refer to the HTML 5.1 WD but I think it's probably simplest to just update this link to https://www.w3.org/TR/html5/browsers.html#top-level-browsing-context.

Alternately we could update all the references to HTML5 to point at HTML5.1 instead (this is all pretty silly IMHO - sorry).

RByers:
LGTM, thanks.  Will merge now, but if there's any additional feedback (eg. in tomorrow's call) we can always land follow-ups.

Issue #61, opened 2016-05-04
Should a captured pointer send boundary events by default?


RByers:
Edge and the spec send `pointerleave`/`pointerout` when a captured pointer moves outside the bounds of an element (and it's descendants), and `pointerenter`/`pointerover` whenever it moves back in. 

There's been some discussion in various places about the implications of this and whether it should change, so [I offered](https://www.w3.org/2016/05/04-pointerevents-minutes.html#item04) to file a dedicated issue to summarize the arguments to see if we can come up with a plan here.  In particular, this is a subset of Chrome's primary ship-blocking spec issue: #8.

I'll try to keep the above summary updated as we debate the details in the conversation below.

## Advantages of sending transition events during capture

1. Edge already ships this, we shouldn't consider changing without a really good reason.
2. This is useful for capture scenarios like buttons where you want the button style to "pop back out" when the cursor is dragged off the capturing element.
  * But in that case, should the button really even be capturing? It needs to also know to ignore `click` events seen after `pointerleave`, so it's mostly acting like an uncaptured control, right?

## Disadvantages of sending transition events during capture

1. It requires the UA to do hit-testing on each `pointermove` events during capture (#8)
  * Some fast paths may be possible in common cases, but detecting these cases is hard in itself.  Eg. if it's possible for a capturing element to be occluded by another element (possibly even via an accelerated animation), then a simple bounds check is insufficient.
2. It means the semantics of the transition events are fundamentally different during capture
  * Normally an element will not receive `pointermove` events after it sees a  `pointerleave`.  As currently defined, when an element has capture the transition events no longer say anything about the event stream.
  * `pointerover` and `pointerout` events normally apply only to the immediate bounds of the element, but under capture they (currently) behave like `pointerenter` and `pointerleave` where they reflect the bounds of the captured element and all it's descendants.  We're struggling to define this precisely (#63, #15) 
3. It means capture can't be defined simply as a modification of (the as yet unspecified) hit-testing algorithm.
  * A very conceptually simple model for capture would be to say that it's just an early-out from hit-testing - makes all pointer hit tests return the captured node.  This is consistent with how implicit touch capture works in browsers like Safari and Chrome today.
4. In drag and drop scenarios (a common use case for capture) the events are only a potential hinderance
  * Typically in drag and drop the application will rely on it's own intersection testing (does the dragged item overlap a drop target at all) rather than built-in hit testing
[...]

PR #63, opened 2016-05-06
Clarified definition of capturing element borders

### LINKED ISSUE #62: While pointer is captured the boundary of the capturing element should be itself and all its children
Looking at this example:
https://output.jsbin.com/xidupe

It shows that the hover state is active even if the mouse is over one of the children and not on the element itself. To mimic the same behavior while pointer is captured to an element if the mouse enters one of its children element it should be considered on the capturing element. Basically when the pointer is captured the meaning of pointerover/out will be the same as the meaning of pointerleave/enter instead of other way around.

This is the same behavior as Edge does but it is not clear from the spec. As we discussed during the meeting I'm going to send a small fix for this to clarify that.

NavidZ:
Closes w3c/pointerevents#62


index.html
```diff
@@ -773,7 +773,7 @@
...
 		</ol>
-		<div class="note">When pointer capture is set, <code>pointerover</code>, <code>pointerout</code>, <code>pointerenter</code>, and <code>pointerleave</code> events are only generated when crossing the boundary of the element that has capture as other elements can no longer be targeted by the pointer. This has the effect of suppressing these events on all other elements.</div>
+		<div class="note">When pointer capture is set, <code>pointerover</code>, <code>pointerout</code>, <code>pointerenter</code>, and <code>pointerleave</code> events are only generated when crossing the boundary of the element that has capture as other elements can no longer be targeted by the pointer. This has the effect of suppressing these events on all other elements. The region of the capturing element includes the element itself and all its children to mimic the hover state of the capturing element. Moving out of (into) this region causes <code>pointerout</code> and <code>pointerleave</code> (<code>pointerover</code> and <code>pointerenter</code>) to be sent to the capturing element and all its ancestors.</div>
```
scottgonzalez: Not all of its ancestors. `pointerenter` and `pointerleave` never bubble.

NavidZ: It doesn't bubble. But as far as I remember Edge does send it (and target it to them if I recall correctly) to all its ancestor one by one like the normal pointerenter/leave. It is as if that as soon as you are out of the capturing element you are out of the whole document.

NavidZ: But I just realized another mistake. Enter/leave should be sent (and retargeted) to all the ancestors one by one but over/out is only sent to the capturing element (and bubbles to the ancestors meaning that the target of all those ancestor events is the capturing element). Here is a demo:

https://output.jsbin.com/sopopo

Did you mean this?

scottgonzalez: > Enter/leave should be sent (and retargeted) to all the ancestors one by one

Nope. Enter the green div, then enter the blue div. You will only get `pointerenter` for the blue div at that point. If `pointerenter` were sent to the green div, you'd lose a lot of what `pointerenter` means in comparison to `pointerover`.

NavidZ: I understand that scenario. But when pointer is captured by the blue div if you leave blue div it is as if you left the whole document. Basically in that link it is the same as leaving blue div from the right hand side and leaving both green and blue div at the same time.

scottgonzalez: Hmm...You're correct that Edge does that, but I would not have expected that to happen.

I'm not sure that anyone would be relying on this behavior today. I'd like to have further discussion about this before documenting that behavior in the spec.

NavidZ: Sure. I posted a newer version of my patch which hopefully I explained the pointerover vs pointerenter behavior better. We can discuss it again in the next meeting. Basically there are two interpretation of the capturing I can imagine regarding the transition events.
1. The whole document and all the ancestors of the capturing element shrink to the borders of the capturing element.
2. The parent element of the capturing element expands to become the whole document and you can never leave it while pointer is captured. So you can only leave the capturing element but you always stay in your parent. This second approach is I believe what you think but looking at this then I kind of expect to get the pointer over for my parent as soon as I leave the capturing element while the other interpretation doesn't have this quirk.

scottgonzalez: Capturing retargets all events to the capturing element. So doesn't any event with a target other than the capturing node violate the spec?

NavidZ: Yup. That observation is correct too. I guess either we should go the second interpretation or if we go to the first one we should mention that capturing only suppresses the events on elements that are not in the ancestor tree of the capturing node.

Also I'd like to know what @jacobrossi and @teddink think and if they have a reason why Edge implements the first interpretation.

index.html
```diff
@@ -773,7 +773,7 @@
...
+		<div class="note">When pointer capture is set, <code>pointerover</code>, <code>pointerout</code>, <code>pointerenter</code>, and <code>pointerleave</code> events are only generated when crossing the boundary of the element that has capture as other elements can no longer be targeted by the pointer. This has the effect of suppressing these events on all other elements. The region of the capturing element includes the element itself and all its children to mimic the hover state of the capturing element. Moving out of this region dispatches <code>pointerout</code> only to the capturing element and <code>pointerleave</code> to the capturing element and all its ancestors as if the pointer has left the whole document. Moving into this region dispatches <code>pointerover</code> only to the capturing element and <code>pointerenter</code> to the capturing element and all its ancestors as if the pointer entered the capturing element from outside of the document.</div>
```
mustaqahmed, line 776: One problem with the current (as well as the original) wording is that it implies that we don't expect the transition events when the border is _not_ crossed: the triggering of transition events is tied to "... generated when crossing..." and "Moving out of this region...".

We need to cover the case when the pointer is captured while the pointer is away from the captured node. In this case, we expect pointerleave/pointerout even when the pointer is stationary (i.e., no boundary crossing). Edge does the right thing here (check [this demo](https://output.jsbin.com/rupulu)).

I think mentioning any of the "interpretations" Navid posted above would let us avoid enumerating all the triggering scenarios. I find the first interpretation pretty intuitive.

RByers:
Yeah firing `pointerenter` and `pointerleave` to the ancestors of the captured element seems strange to me too.  I'm already struggling to understand the model for transition events during capture (#61), I'm really having a hard time imaging the use case for this behavior.  Maybe that's just a bug in Edge?

NavidZ:
With Chrome going forward with change in the definition of hit-testing while a pointer is captured, I will close this PR and create another one which includes that proposal in one PR.

PR #66, opened 2016-05-12
Make it explicit that buttons is a bitmask.

### LINKED ISSUE #4: Value of button in a chorded pointerdown event
This Issue captures the following email from  Stuart P. Bentley stuart@testtrack4.com on 16 Mar 2015

[[
https://lists.w3.org/Archives/Public/public-pointer-events/2015JanMar/0029.html

It's not clear in the 24 February 2015 draft - what is the value of 
`button` supposed to be when multiple buttons are pressed in a 
`pointerdown` (or `pointerup`) event (as the current spec says only one 
event will be generated)?

(It's also not particularly clear in the draft that the `buttons` values 
are supposed to be a bitmask, which is only mentioned in the Mouse 
Events spec.)
]]

mustaqahmed:
Closes #4.

### DIFF
--- a/index.html
+++ b/index.html
@@ -343,7 +343,7 @@ eventTarget.dispatchEvent(event);
 			</section>
 			<section>
 				<h3>The <code>buttons</code> property</h3>
-				<div><p>The <code>buttons</code> property gives the current state of the device buttons (same as in <code>MouseEvent</code>, but with an expanded set of possible values).</p></div>
+				<div><p>The <code>buttons</code> property gives the current state of the device buttons as a bitmask (same as in <code>MouseEvent</code>, but with an expanded set of possible values).</p></div>
 				<table class="simple">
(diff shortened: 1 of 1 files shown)


scottgonzalez:
Looks good.

teddink:
Looks good to me as well - thanks for clarifying.

patrickhlauke:
LGTM. i'll be bold and merge this despite not being an editor, as this is a straightforward fix/addition.

RByers:
LGTM to me also.

> despite not being an editor

@patrickhlauke, want to fix that?  You've obviously been contributing a lot.  I spoke to the other editors and we all agree that we'd be happy to have you edit if you're interested and don't think it would interfere (eg. time commitments) with your role as chair.  If you're interested, go ahead and put up a PR and I'll merge it.

PR #67, opened 2016-05-12
Remove explicit "Status of this Document"

### LINKED ISSUE #45: Should the spec really reference tracker anymore?
The spec status currently says:

> Action items to be completed are tracked in Tracker.

Can we just officially say GitHub issue are our only source of tracking now?  If we want to do that, someone probably needs to go through the [open tracker items](https://www.w3.org/2012/pointerevents/track/) and make sure anything still relevant has a GitHub issue.

patrickhlauke:
This removes the extra reference to github and tracker - ReSpec does still add the more general status section, which should be sufficient.

Closes #45


RByers:
Thanks!

Issue #65, opened 2016-05-11
Default width/height attribute for mouse


NavidZ:
Spec is sort of silence about the default value of width and height property of devices that are not reporting the actual geometry of the pointer particularly mouse.

"The width (magnitude on the X axis), in CSS pixels (see [CSS21]), of the contact geometry of the pointer. This value may be updated on each event for a given pointer. For devices which have a contact geometry but the actual geometry is not reported by the hardware, a default value should be provided by the user agent to approximate the geometry typical of that pointer type. Otherwise, the value must be 0."

Edge reports width=height=1 for mouse or the normal pen. In our current Chrome implementation we  report 0. I wanted to match the behavior of Edge but I thought If we are matching the behavior in this level is it worth it to maybe add a note of some sort to also mention this in the spec or not. WDYT?

scottgonzalez:
I would expect the value to be 1 for mice.

teddink:
Adding a note is fine - like Scott I would expect the value to be one for mice and pen and trackpad, unless of course the HW device reports something more specific and granular.

patrickhlauke:
Not sure where we'd put the note though. As part of the definition of `width`, of `height`, of both, or - very disconnected - after all the interface definitions?

Personally, I think authors would simply expect a value of 1 for `width` and `height` in the case of a mouse and stylus...not sure it needs spelling out explicitly.

mustaqahmed:
I agree that adding a note could be a bit tricky here. But what about changing the default value to 1 (from 0)?

A device which doesn't know about the contact geometry still must report a pointer coordinate that essentially "points" to a CSS pixel---I guess this is why we easily agreed that mouse/pen/trackpad should have default 1.

patrickhlauke:
I guess the difference here is: if I query the property and get 0, this clearly means that the hardware/OS/UA had no idea what the width/height were and did not even attempt to make a guess (which, depending on what pointer type this came from, can tell the author that they need to do their own guessing/choosing a value for whatever their code does). If we standardise to 1, it may be a device that truly has a 1x1 px "tip" or it may be the former case.

NavidZ:
So with this discussion should we maybe set the width and height of mouse pointers to 0 as well and let the author decide what to do with that?

patrickhlauke:
[...]

PR #69, opened 2016-05-20
Make width/height default to 1, remove UA "guessing"/faking geometry

patrickhlauke:
Closes https://github.com/w3c/pointerevents/issues/65

### DIFF
--- a/index.html
+++ b/index.html
@@ -244,11 +244,11 @@ eventTarget.dispatchEvent(event);
 				</dd>
 				<dt>readonly attribute double width</dt>
 				<dd>
-					The width (magnitude on the X axis), in CSS pixels (see [[CSS21]]), of the <a>contact geometry</a> of the pointer.  This value MAY be updated on each event for a given pointer. For devices which have a contact geometry but the actual geometry is not reported by the hardware, a default value SHOULD be provided by the <a>user agent</a> to approximate the geometry typical of that pointer type. Otherwise, the value MUST be 1.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -244,11 +244,11 @@
 				</dd>
 				<dt>readonly attribute double width</dt>
 				<dd>
-					The width (magnitude on the X axis), in CSS pixels (see [[CSS21]]), of the <a>contact geometry</a> of the pointer.  This value MAY be updated on each event for a given pointer. For devices which have a contact geometry but the actual geometry is not reported by the hardware, a default value SHOULD be provided by the <a>user agent</a> to approximate the geometry typical of that pointer type. Otherwise, the value MUST be 0.
+					The width (magnitude on the X axis), in CSS pixels (see [[CSS21]]), of the <a>contact geometry</a> of the pointer.  This value MAY be updated on each event for a given pointer. For inputs that typically lack contact geometry (such as a traditional mouse or stylus), and in cases where the actual geometry of the input is not detected by the hardware, the <a>user agent</a> MUST return a default value of 1.
```
mustaqahmed: May we possibly omit "stylus"? Stylus "contact area" could be considered equivalent to "having a geometry".

teddink:
Change looks good to me. I am fine either with or without the reference to stylus.

index.html
patrickhlauke: good point...in hindsight, i completely agree that a stylus has a contact geometry. going to make the change and force-push

RByers:
LGTM

patrickhlauke:
Merging, as no objections. https://www.w3.org/2016/05/25-pointerevents-minutes.html

PR #76, opened 2016-06-02
A possible immediate pointer capture processing

### LINKED ISSUE #32: When are lost/gotpointercapture delayed until the next pointer event?
Looking at this 
https://www.w3.org/TR/pointerevents/#process-pending-pointer-capture
and 
https://www.w3.org/TR/pointerevents/#setting-pointer-capture

It seems that user agent should not fire gotpointercapture right after setPointerCapture and instead fire it right before the next pointer event with the captured id. But in this example I can see Edge doesn't do that and instead fires gotpointercapture right after pointerdown for the green div.

https://jsbin.com/hivude/edit?html,output

NavidZ:
closes #32 

This is one possible way of doing the immediate pointer capture processing. As far as I tested Edge it conforms with it.

@jacobrossi @teddink I'm not sure how much this PR can be help but could this be something that Edge does?


teddink:
> I now understand the behavior that Edge (and IE11) currently has. Edge will immediately fire `gotpointercapture` ONLY when `setpointercapture' was called as a result of`pointerdown`. The code does NOT do anything special for`lostpointercapture`- so`lostpointercapture` continues to follow the spec as written. I will also comment on PR #76 by @NavidZ.

As I read this PR, it does not match the Edge behavior as implemented today.
- The changes made to section "10.1 Setting Pointer Capture" look good.
- The changes made to section "10.2 Releasing Pointer Capture" are not needed.
- I am not sure about the changes made to the "Process Pending Pointer Capture" paragraph. I am not sure that we need the `temporaryPendingPointerCaptureTargetOverride` when the `pendingPointerCaptureTargetOverride` seems to be sufficient.

NavidZ:
The temporary variable is needed I believe as the handlers between the first step and the last step of "Processing Pending Pointer Capture" routine might set the pending pointer capture but at the end of that routine we should use the original value and set it in pointer capture. Right?

NavidZ:
I didn't see anyone talking about this issue in the last call.
@teddink just wanted to follow up on this to see if you guys were okay with this proposal.

teddink:
Reviewing one more time with my dev team to ensure that we agree with the wording.

NavidZ:
Just to make sure the difference is noted. This is slightly different from what Edge does. However, in a more generic way it covers the case that you said you added the extra pointermove event to prevent delaying the gotpointercapture.

index.html
```diff
@@ -409,8 +409,9 @@
 				<li>Otherwise, fire the event to the object returned by normal hit test mechanisms (out of scope for this specification).</li>
 			</ul>
             <h4>Process Pending Pointer Capture</h4>
-            <p>Whenever a user agent is to fire a Pointer Event that is not <code>gotpointercapture</code> or <code>lostpointercapture</code>, it must first run these steps:</p>
+            <p>Whenever a user agent is to fire a Pointer Event that is not <code>gotpointercapture</code> or <code>lostpointercapture</code>, it must first run these steps for the corresponding <code>pointerId</code>:</p>
```
mustaqahmed: The opening para need to include the immediate processing during set/release capture, something like this:

> Calls to <code>setPointerCapture</code> or <code>releasePointerCapture</code> may also trigger these steps, see Sec10.1 and 10.2 for details.

NavidZ:
I rebased this PR and applied @mustaqahmed comment. Note the addition of settings attributes of got/lostpointercapture as that will be needed if we can send those events irrespective of an input event.

NavidZ:
This also closes #113 with solution B proposed there.

NavidZ:
As discussed in the Hackathon, we stick to the delayed model in all the scenarios. I'll close this PR for now and create another one for the attributes of got/lostpointercapture which is briefly mentioned in this PR.

Issue #25, opened 2015-10-28
Add additional digitizer/pen attributes: twist (rotation)


skibulk:
Brush and Marker rotation is used by painters, calligraphers, and comic artists to achieve a variety of strokes. Stylus rotation is the digital equivalent, and is enabled by Wacom products, Adobe Photoshop. There are already numerous web apps which support drawing and painting. Lets extend their capabilities.

https://www.youtube.com/watch?v=2gm9lM7svaY
https://www.youtube.com/watch?v=ukXJQAY0caU

RByers:
Thanks for the suggestion.  Can you point me at the existing native APIs for determining rotation?

On Windows 8+ I see [PointerPointProperties.Orientation](https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.input.pointerpointproperties.orientation) and [PointerPointProperties.Twist](https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.input.pointerpointproperties.twist).  It's not clear to me what the difference is between these.

If someone can point to a native demo app (on Windows or otherwise) and specific device where rotation is supported, it would probably help.  I have a Surface Pro 3, and AFAIK that pen supports tilt but not rotation (but I'm not completely sure).

patrickhlauke:
Don't have a device at hand to test right now, but isn't a barrel rotation going to already result in the tiltX/tiltY changing (as I'm assuming the tilt will reference some idealised "zero position" of the pen, so rotating it will result in a different set of angles)?

patrickhlauke:
Also, Surface pen doesn't support tilt as far as I know, only pressure (and no, nothing specific about rotation either)

RByers:
> Don't have a device at hand to test right now, but isn't a barrel rotation going to already result in the tiltX/tiltY changing (as I'm assuming the tilt will reference some idealised "zero position" of the pen, so rotating it will result in a different set of angles)?

I don't think so.  There's three axis of movement possible here.  Eg. imagine the pen pointing straight up and down so tiltX and tiltY are fixed.  Now roll the pen between your fingers - rotation changes without changing either tilt.

> Also, Surface pen doesn't support tilt as far as I know, only pressure (and no, nothing specific about rotation either)

Oh sorry, I was thinking your stylus video (showing tilt) was with a Surface.  What device was that again?  I guess I better get one too.

patrickhlauke:
[...]

PR #79, opened 2016-06-07
Add digitizer/pen twist

patrickhlauke:
Closes https://github.com/w3c/pointerevents/issues/25

### DIFF
--- a/index.html
+++ b/index.html
@@ -274,6 +274,10 @@ eventTarget.dispatchEvent(event);
 						<figcaption>Negative <code>tiltY</code>.</figcaption>
 					</figure>
 				</dd>
+        <dt>readonly attribute long twist</dt>
+        <dd>
+          The clockwise rotation (in degrees, in the range of [0,359]) of a transducer (e.g. pen stylus) around its own major axis. For devices that do not report twist, the value MUST be 0.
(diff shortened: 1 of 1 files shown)


patrickhlauke:
@dfleck one minor question: you mentioned that the value is 0-360. https://msdn.microsoft.com/en-us/library/windows/apps/windows.ui.input.pointerpointproperties.twist?f=255&MSPPError=-2147217396 mentions 0.0 - 359.0. I assume to be precise, it's 0.0 - 359.9999999999999999 (or whatever the precision is). does it make sense to just say 0-360 for simplicity's sake?

RByers:
Should we just reference the [USB HID Usage](http://www.usb.org/developers/hidpage/Hut1_12v2.pdf)?  I.e. Usage ID 0x41?  Though I still don't see any definition in that spec of the allowable values.  Still saying simply 0-360 seems reasonable to me (we can write it `[0-360)` if we want to be precise that it's exclusive of 360).

patrickhlauke:
Agreed in today's call https://www.w3.org/2016/06/08-pointerevents-minutes.html to merge this as is - as the property is of type `long`, [0,359] is appropriate and less confusing than [0,360)

Issue #70, opened 2016-05-23
Add additional digitizer/pen attributes: barrel pressure (tangential pressure)


patrickhlauke:
this issue splits out the discussion around https://github.com/w3c/pointerevents/issues/25#issuecomment-193898076 - adding a barrel pressure / tangential pressure attribute (implemented as a scroll-wheel or similar on airbrush digitizers), maps to HID digitizer 0x31

patrickhlauke:
In the call today (https://www.w3.org/2016/05/25-pointerevents-minutes.html) the question about actual devices that support rotation came up. This Wacom airbrush pen (compatible with Intuos Pro) https://us-store.wacom.com/Product/Airbrush-Pen-black-S01 supports barrel pressure.

RByers:
Looks like MacOS has a standard API for this: [NSEvent tangentialPressure](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSEvent_Class/#//apple_ref/occ/instp/NSEvent/tangentialPressure).  This passes my "at least one major OS must have built-in API support test", so I support adding this.

PR #80, opened 2016-06-07
Add digitizer/pen tangential (barrel) pressure

patrickhlauke:
Closes https://github.com/w3c/pointerevents/issues/70

### DIFF
--- a/index.html
+++ b/index.html
@@ -258,6 +258,10 @@ eventTarget.dispatchEvent(event);
 				<dd>
 					The normalized pressure of the pointer input in the range of [0,1], where 0 and 1 represent the minimum and maximum pressure the hardware is capable of detecting, respectively. For hardware that does not support pressure, the value MUST be 0.5 when in the <a>active buttons state</a> and 0 otherwise.
 				</dd>
+        <dt>readonly attribute float barrelPressure</dt>
+        <dd>
+          The normalized barrel pressure (set by an additional control, e.g. a finger wheel on a stylus) of the pointer input in the range of [0,1], where 0 and 1 represent the minimum and maximum pressure the hardware is capable of detecting, respectively. For hardware that does not support barrel pressure, the value MUST be 0.
+        </dd>
 				<dt>readonly attribute long tiltX</dt>
 				<dd>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
As the feature is not as common, I crowbarred an explanation in brackets in there to mention finger wheels on stylus. Does that work?

patrickhlauke:
Btw compared to twist, I can sort of see the argument that this might well be more niche

RByers:
Seems fine to me, LGTM

patrickhlauke:
Discussed in today's call https://www.w3.org/2016/06/08-pointerevents-minutes.html - need to discuss further, find out if there's any existing implementations / whether it's exposed in a native API (not just as custom property when a specific hardware driver is installed). /cc @teddink 
Potential to add to spec, note that it's "at risk" because currently not exposed as part of standard system API (but would give more incentive to OS developers to include it)

RByers:
See [my comment on the issue](https://github.com/w3c/pointerevents/issues/70#issuecomment-224750781).  I don't think this needs to be "at risk".  We'd expose MacOS's `tangentialPressure` property in this field in Chrome.

Note however that it's [defined](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSEvent_Class/#//apple_ref/occ/instp/NSEvent/tangentialPressure) to be in the range [-1.0,1.0].  I suggest we just use the same range here (even though most devices may send only non-negative values)

patrickhlauke:
Seems QT also implements tangential pressure, again in the [-1.0, 1.0] range http://doc.qt.io/qt-5/qtabletevent.html#tangentialPressure.

As "tangential pressure" seems to be the more common term (just doing a very unscientific search for that term brings up more relevant results than "barrel pressure"), i'm wondering if we should go for that term instead (and yes, go for the [-1.0, 1.0] range).

patrickhlauke:
Will rebase this before merging, but in the first instance interested in any feedback on the latest proposed change.

teddink:
Looks good to me.

RByers:
LGTM

patrickhlauke:
Having trouble rebasing for some reason...will start with a separate fresh PR from current gh-pages head...

PR #81, opened 2016-06-07
Add the note to define the scope of pointerId more precisely

### LINKED ISSUE #59: Should pointer IDs be unique even across different top-level browsing contexts?
In Edge the pointer ID is provided by the OS and so is global across the entire system.  In #52 we agreed to require the ID to be unique the top-level browsing context, but it's possible an application could depend on it being unique even across different top-level browsing contexts.  Eg. two instances of an app (in different windows) could maybe be communicating via a server and supporting drag-and-drop / pointer-handoff between them.  This would be hard to spec, and possibly impossible to implement in some OSes.  To what extent to we think this might matter for interop and useful functionality?

@teddink said he'd look into what exactly is possible on Windows today.  Eg. when dragging from one window to another, is the consistency of pointer IDs observable?

NavidZ:
closes #59 .


patrickhlauke:
LGTM

teddink:
Agreed - LGTM2

RByers:
LGTM also, thanks!

Issue #82, opened 2016-06-08
tabs vs spaces (sorry)


patrickhlauke:
At the risk of starting a holy war...current spec markup uses both, and it's confusing. Can we normalize this to either tabs or spaces? I'm not particularly fussed about which one (if you have strong opinions, voice them here now...)

RByers:
Yes please.  Spaces offer the least opportunity for confusion IMHO.

PR #84, opened 2016-06-08
Markup cleanup (tabs vs spaces, indentation, validation errors)

patrickhlauke:
Closes https://github.com/w3c/pointerevents/issues/82

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
   <head>
     <title>Pointer Events - Level 2</title>
     <meta http-equiv='Content-Type' content='text/html;charset=utf-8'/>
-	<meta http-equiv="X-UA-Compatible" content="IE=Edge">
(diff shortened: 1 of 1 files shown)


patrickhlauke:
(with apologies for the major reworking - due mostly to re-indentation - which makes this probably quite difficult to diff)

patrickhlauke:
suffice to say that (with one minor exception - namely the fact that "Firing events using the PointerEvent interface" is NOT numbered by respec) the overall structure and TOC remain the same, but more importantly the validation errors (with the exception of respec's custom `lt` attribute on `<dfn>`) are now gone

RByers:
I did a quick skim locally (using `git diff -b --word-diff=color`), LGTM.

RByers:
I merged instead of squashing to keep the mechanical commit separate (may make tracing history slightly easier).

PR #96, opened 2016-06-23
Removed "pen contact" condition on button/buttons.

### LINKED ISSUE #14: Support hovering button states
[Originating thread](http://lists.w3.org/Archives/Public/public-pointer-events/2015AprJun/0080.html)

The latest proposal looks something like this:
- Allow `buttons` to change while hovering
- Restrict pointerdown/up to only be for left buttton, touch, and stylus tip on screen (e.g. `button` value of 0)

We should also consider:
- should `click` fire for hovering button presses?
- should there be a new event, e.g. `pointerbuttonschange` that fires in this case?

Note that the latter point could also be useful for chorded button scenarios where you want an easy way to know another button changed even if the pointer is down.

mustaqahmed:
To support button presses on a hovering stylus, button and buttons shouldn't require a pen contact when barrel/eraser buttons are pressed. This change makes pen buttons independent of pen contact.

Note that contact independent button/buttons values cause subtle changes in active button state.

Closes #14.

### DIFF
--- a/index.html
+++ b/index.html
@@ -211,7 +211,7 @@ eventTarget.dispatchEvent(event);
         <h1>Glossary</h1>
         <dl>
             <dt><dfn>active buttons state</dfn></dt>
-                <dd>The condition when a pointer has a non-zero value for the <code>buttons</code> property. For mouse, this is when the device has at least one button depressed. For touch, this is when there is physical contact with the digitizer. For pen, this is when the pen either has physical contact with the digitizer, or has at least one button depressed while hovering.</dd>
+                <dd>The condition when a pointer has a non-zero value for the <code>buttons</code> property. For mouse, this is when the device has at least one button depressed. For touch, this is when there is physical contact with the digitizer. For pen, this is when either the pen has physical contact with the digitizer, or at least one button is depressed while hovering.</dd>
             <dt><dfn>active pointer</dfn></dt>
                 <dd>Any touch contact, pen stylus, mouse cursor, or other pointer that can produce events.  If it is possible for a given pointer (identified by a unique <code>pointerId</code>) to produce additional events within the document, then that pointer is still considered active. Examples:
                     <ul>
@@ -350,10 +350,10 @@ eventTarget.dispatchEvent(event);
                             <tr><td>Neither buttons nor touch/pen contact changed since last event</td><td>-1</td></tr>
                             <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact</td><td>0</td></tr>
                             <tr><td>Middle Mouse</td><td>1</td></tr>
-                            <tr><td>Right Mouse,<br>Pen barrel button pressed</td><td>2</td></tr>
+                            <tr><td>Right Mouse,<br>Pen barrel button</td><td>2</td></tr>
                             <tr><td>X1 (back) Mouse</td><td>3</td></tr>
                             <tr><td>X2 (forward) Mouse</td><td>4</td></tr>
-                            <tr><td>Pen eraser button pressed</td><td>5</td></tr>
+                            <tr><td>Pen eraser button</td><td>5</td></tr>
                         </tbody>
                     </table>
                     <div class="note">During a mouse drag, the value of the <code>button</code> property in a <code>pointermove</code> event will be different from that in a <code>mousemove</code> event. For example, while moving the mouse with the right button pressed, the <code>pontermove</code> events will have the <code>button</code> value -1, but the <code>mousemove</code> events will have the <code>button</code> value 2.</div>
@@ -364,13 +364,13 @@ eventTarget.dispatchEvent(event);
                     <table class="simple">
                         <thead><tr><th>Current state of device buttons</th><th><code>buttons</code></th></tr></thead>
                         <tbody>
-                            <tr><td><b>Mouse moved with no buttons pressed</b><br>, Pen moved without changing its <a>active buttons state</a></td><td>0</td></tr>
+                            <tr><td><b>Mouse moved with no buttons pressed</b>,<br> Pen moved while hovering with no buttons pressed</td><td>0</td></tr>
                             <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact</td><td>1</td></tr>
                             <tr><td>Middle Mouse</td><td>4</td></tr>
-                            <tr><td>Right Mouse,<br>Pen barrel button pressed</td><td>2</td></tr>
+                            <tr><td>Right Mouse,<br>Pen barrel button</td><td>2</td></tr>
                             <tr><td>X1 (back) Mouse</td><td>8</td></tr>
                             <tr><td>X2 (forward) Mouse</td><td>16</td></tr>
-                            <tr><td>Pen eraser button pressed</td><td>32</td></tr>
+                            <tr><td>Pen eraser button</td><td>32</td></tr>
                         </tbody>
                     </table>
                 </section>
--- a/index.html
+++ b/index.html
@@ -211,7 +211,7 @@ eventTarget.dispatchEvent(event);
         <h1>Glossary</h1>
         <dl>
             <dt><dfn>active buttons state</dfn></dt>
-                <dd>The condition when a pointer has a non-zero value for the <code>buttons</code> property. For mouse, this is when the device has at least one button depressed. For touch, this is when there is physical contact with the digitizer. For pen, this is when the pen has physical contact with the digitizer.</dd> 
+                <dd>The condition when a pointer has a non-zero value for the <code>buttons</code> property. For mouse, this is when the device has at least one button depressed. For touch, this is when there is physical contact with the digitizer. For pen, this is when the pen either has physical contact with the digitizer, or has at least one button depressed while hovering.</dd>
             <dt><dfn>active pointer</dfn></dt>
                 <dd>Any touch contact, pen stylus, mouse cursor, or other pointer that can produce events.  If it is possible for a given pointer (identified by a unique <code>pointerId</code>) to produce additional events within the document, then that pointer is still considered active. Examples:
                     <ul>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -364,7 +364,7 @@
                     <table class="simple">
                         <thead><tr><th>Current state of device buttons</th><th><code>buttons</code></th></tr></thead>
                         <tbody>
-                            <tr><td><b>Mouse move with no buttons pressed</b></td><td>0</td></tr>
+                            <tr><td><b>Mouse moved with no buttons pressed</b><br>, Pen moved without changing its <a>active buttons state</a></td><td>0</td></tr>
```
mustaqahmed: A quick thought: this seems more correct to replace the mouse/pen specific lines with:
"No change in active buttons state   =>   0"

index.html
```diff
@@ -211,7 +211,7 @@
         <h1>Glossary</h1>
         <dl>
             <dt><dfn>active buttons state</dfn></dt>
-                <dd>The condition when a pointer has a non-zero value for the <code>buttons</code> property. For mouse, this is when the device has at least one button depressed. For touch, this is when there is physical contact with the digitizer. For pen, this is when the pen has physical contact with the digitizer.</dd> 
+                <dd>The condition when a pointer has a non-zero value for the <code>buttons</code> property. For mouse, this is when the device has at least one button depressed. For touch, this is when there is physical contact with the digitizer. For pen, this is when the pen either has physical contact with the digitizer, or has at least one button depressed while hovering.</dd>
```
patrickhlauke: "...or at least one button _is_ depressed while hovering"

It may also need an extra qualifier at the end of that sentence, along the lines of "(for hardware that supports this)"? Or do all digitizers with pens that have buttons intrinsically able to recognize these buttons? (the answer here may be yes, though perhaps even then the qualifier may be useful to have)

index.html
```diff
@@ -348,12 +348,12 @@
...
                             <tr><td>Neither buttons nor touch/pen contact changed since last event</td><td>-1</td></tr>
-                            <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact (with no modifier buttons pressed)</td><td>0</td></tr>
+                            <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact</td><td>0</td></tr>
                             <tr><td>Middle Mouse</td><td>1</td></tr>
-                            <tr><td>Right Mouse,<br>Pen contact with barrel button pressed</td><td>2</td></tr>
+                            <tr><td>Right Mouse,<br>Pen barrel button pressed</td><td>2</td></tr>
                             <tr><td>X1 (back) Mouse</td><td>3</td></tr>
                             <tr><td>X2 (forward) Mouse</td><td>4</td></tr>
-                            <tr><td>Pen contact with eraser button pressed</td><td>5</td></tr>
+                            <tr><td>Pen eraser button pressed</td><td>5</td></tr>
```
patrickhlauke: Maybe for consistency (with the likes of "Left Mouse", "Right Mouse" etc) remove "pressed"?

index.html
```diff
@@ -348,12 +348,12 @@
                         <thead><tr><th>Device Button Changes</th><th><code>button</code></th></tr></thead>
                         <tbody>
                             <tr><td>Neither buttons nor touch/pen contact changed since last event</td><td>-1</td></tr>
-                            <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact (with no modifier buttons pressed)</td><td>0</td></tr>
+                            <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact</td><td>0</td></tr>
                             <tr><td>Middle Mouse</td><td>1</td></tr>
-                            <tr><td>Right Mouse,<br>Pen contact with barrel button pressed</td><td>2</td></tr>
+                            <tr><td>Right Mouse,<br>Pen barrel button pressed</td><td>2</td></tr>
```
patrickhlauke: Remove "pressed"? (see below)

index.html
```diff
@@ -364,13 +364,13 @@
                     <table class="simple">
                         <thead><tr><th>Current state of device buttons</th><th><code>buttons</code></th></tr></thead>
                         <tbody>
-                            <tr><td><b>Mouse move with no buttons pressed</b></td><td>0</td></tr>
-                            <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact (with no modifier buttons pressed)</td><td>1</td></tr>
+                            <tr><td><b>Mouse moved with no buttons pressed</b><br>, Pen moved without changing its <a>active buttons state</a></td><td>0</td></tr>
+                            <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact</td><td>1</td></tr>
                             <tr><td>Middle Mouse</td><td>4</td></tr>
-                            <tr><td>Right Mouse,<br>Pen contact with barrel button pressed</td><td>2</td></tr>
+                            <tr><td>Right Mouse,<br>Pen barrel button pressed</td><td>2</td></tr>
```
patrickhlauke: Remove "pressed"

index.html
```diff
@@ -364,13 +364,13 @@
...
                             <tr><td>X1 (back) Mouse</td><td>8</td></tr>
                             <tr><td>X2 (forward) Mouse</td><td>16</td></tr>
-                            <tr><td>Pen contact with eraser button pressed</td><td>32</td></tr>
+                            <tr><td>Pen eraser button pressed</td><td>32</td></tr>
```
patrickhlauke: Remove "pressed"

index.html
```diff
@@ -348,12 +348,12 @@
                         <thead><tr><th>Device Button Changes</th><th><code>button</code></th></tr></thead>
                         <tbody>
                             <tr><td>Neither buttons nor touch/pen contact changed since last event</td><td>-1</td></tr>
-                            <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact (with no modifier buttons pressed)</td><td>0</td></tr>
+                            <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact</td><td>0</td></tr>
```
patrickhlauke, line 351: Random question: are there pens that have a (redundant/additional) "left click" button which would allow a hovering click? If so, we'd need to probably add that here too, unless that makes it too confusing?

index.html
```diff
@@ -364,13 +364,13 @@
                     <table class="simple">
                         <thead><tr><th>Current state of device buttons</th><th><code>buttons</code></th></tr></thead>
                         <tbody>
-                            <tr><td><b>Mouse move with no buttons pressed</b></td><td>0</td></tr>
-                            <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact (with no modifier buttons pressed)</td><td>1</td></tr>
+                            <tr><td><b>Mouse moved with no buttons pressed</b><br>, Pen moved without changing its <a>active buttons state</a></td><td>0</td></tr>
```
patrickhlauke: Unless I'm misunderstanding this, would this not mean that for a hovering pen, the first time the user presses one of the buttons, they'd get a non-zero `buttons`, but on subsequent move they'd get 0 regardless of what buttons are pressed as it's not "transitioning" per the new active buttons definition? So `buttons` wouldn't reflect the state, but just a change, and only once?

index.html
scottgonzalez: I don't think it needs an extra qualifier. If the device doesn't have a button, then that portion of the conditional will just never be met, but it's an or so it doesn't matter.

index.html
scottgonzalez, line 351: That seems like it'd be pretty awkward to use, so there probably aren't any.

index.html
scottgonzalez: This should be hovering pen moved with no buttons pressed. The active buttons state shouldn't come into play.

index.html
patrickhlauke, line 351: Actually, just tested this and Wacom Intuos Pro supports hovering left click as an option to be mapped to the two barrel buttons, in addition to the tip. By default the two buttons are set to right click and double click, but it IS possible (and Wacom Intuos Pro not that uncommon a device) 
wacom-intuos-pro

scottgonzalez, line 351: In that case, we should probably mention it.

patrickhlauke, line 351: related, see https://github.com/w3c/pointerevents/issues/97 - this may be a quirk of Wacom Intuos Pro and its driver. Perhaps it's best to leave this out for now, under the implied assumption that in this case the pen/driver will internally treat the hovering click button press as a "contact", which means it's covered adequately already by having `Pen contact` in the text there

index.html
mustaqahmed: I have removed all the "pressed" you suggested,

index.html
mustaqahmed: Good catch, thanks. Hope this is more precise now.

index.html
mustaqahmed, line 351: I also think that Wacom's button swapping is a low (driver) level setting that is beyond the scope of this spec. A similar case: mouse L/R button swapping is possible is in many (all?) OSes, and quite logically UI event spec is silent about it.

patrickhlauke:
LGTM (and ignore the remaining note/discussion about hovering click, as I think that's sufficiently addressed in current wording and not exactly standard anyway).

index.html
```diff
@@ -364,13 +364,13 @@
                     <table class="simple">
                         <thead><tr><th>Current state of device buttons</th><th><code>buttons</code></th></tr></thead>
                         <tbody>
-                            <tr><td><b>Mouse move with no buttons pressed</b></td><td>0</td></tr>
-                            <tr><td>Left Mouse,<br>Touch Contact,<br>Pen contact (with no modifier buttons pressed)</td><td>1</td></tr>
+                            <tr><td><b>Mouse moved with no buttons pressed</b>,<br> Pen moved while hovering with no buttons pressed</td><td>0</td></tr>
```
mustaqahmed, line 367: > Pen _moved_ while hovering with no buttons pressed

Note that "moved" here includes the cases when the "footprint" of the pen is unmoved: e.g. a change in <code>twist</code>. I think the pen still must move physically to generate an event, so this is fine.

Do we need a note?

RByers, line 367: I think it's fine to treat that as implied.

RByers:
LGTM.  Not seeing any other objections to the wording, let's land this for now - we can always continue to iterate.

RByers:
Oh except this is substantial enough that we probably need an entry in the changlog for it (just link to the PR).

PR #99, opened 2016-07-06
Add pinch-zoom token to touch-action

### LINKED ISSUE #29: Add explicit control over pinch-zoom to touch-action
Consider a common scenario like a long web page with an image carousel .  Our guidance to developers would be to put something like 'touch-action: pan-y' on the carousel to permit custom horizontal event handling without disabling scrolling.  But this can be an accessibility issue - if the user can pinch-zoom elsewhere in the document, shouldn't they be able to pinch-zoom on the carousel as well?

Edge supports this with the `pinch-zoom` [touch-action token](https://msdn.microsoft.com/en-ca/library/windows/apps/hh767313.aspx) but we omitted that from the spec due to "pinch" being out of scope.  But then we ended up referring to "continuous zooming" in normative text anyway.  So why not just add a `continuous-zoom` token too?

### LINKED ISSUE #565: Add explicit control over pinch-zoom to touch-action
Consider a common scenario like a long web page with an image carousel .  Our guidance to developers would be to put something like 'touch-action: pan-y' on the carousel to permit custom horizontal event handling without disabling scrolling.  But this can be an accessibility issue - if the user can pinch-zoom elsewhere in the document, shouldn't they be able to pinch-zoom on the carousel as well?

Edge supports this with the `pinch-zoom` [touch-action token](https://msdn.microsoft.com/en-ca/library/windows/apps/hh767313.aspx) but we omitted that from the spec due to "pinch" being out of scope.  But then we ended up referring to "continuous zooming" in normative text anyway.  So why not just add a `continuous-zoom` token too?

RByers:
Based on the [existing Microsoft implementation](https://msdn.microsoft.com/en-ca/library/windows/apps/hh767313.aspx) but defined to be gesture-agnostic.
Fixes w3c/pointerevents#565

### DIFF
--- a/index.html
+++ b/index.html
@@ -700,7 +700,8 @@ eventTarget.dispatchEvent(event);
             <h2>The <code>touch-action</code> CSS property</h2>
             <table class="simple">
                 <tr><th>Name:</th><td><code>touch-action</code></td></tr>
-                <tr><th>Value:</th><td><code>auto</code> | <code>none</code> | [ [ <code>pan-x</code> | <code>pan-left</code> | <code>pan-right</code> ] || [ <code>pan-y</code> | <code>pan-up</code> | <code>pan-down</code> ] ] | <code>manipulation</code></td></tr>
+                <tr><th>Value:</th><td><code>auto</code> | <code>none</code> | <code>manipulation</code> | <br>
+                 [ [ <code>pan-x</code> | <code>pan-left</code> | <code>pan-right</code> ] || [ <code>pan-y</code> | <code>pan-up</code> | <code>pan-down</code> ] || <code>pinch-zoom</code> ] </td></tr>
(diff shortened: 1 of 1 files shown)


RByers:
@teddink @jacobrossi WDYT?  Preview [here](https://rawgit.com/RByers/w3c-pointerevents/touch-action-pinch-zoom/index.html#the-touch-action-css-property).

patrickhlauke:
@rbyers is this still active/relevant? @teddink any news on whether the esquires are happy?

jpmedley:
This is confirmed as being in Chrome 56. Will it make it into the actual spec?

https://bugs.chromium.org/p/chromium/issues/detail?id=626086

RByers:
See w3c/pointerevents#565 - now spec'd [here](https://compat.spec.whatwg.org/#touch-action).

PR #101, opened 2016-07-07
convert deprecated <dl> based WebIDL to Contiguous IDL

### LINKED ISSUE #58: Convert spec IDL to "contiguous" style
ReSpec now generates a warning saying:

> Defining WebIDL in `dl` elements is deprecated. Please use Contiguous IDL instead: https://www.w3.org/respec/guide.html#contiguous-idl

Perhaps it's time to finally rewrite using the preferred syntax?  Note that we looked at this for TE at one point: w3c/touch-events#31.

/cc @NavidZ @mustaqahmed in case either of you are interested in picking this up.

patrickhlauke:
https://www.w3.org/respec/examples/webidl-contiguous.html

closes https://github.com/w3c/pointerevents/issues/58

### DIFF
--- a/index.html
+++ b/index.html
@@ -245,30 +245,58 @@ eventTarget.dispatchEvent(event);
         <section>
             <h2><code>PointerEvent</code> Interface</h2>
             <div>
-                <dl class="idl" title="[Constructor(DOMString type, optional PointerEventInit eventInitDict)] interface PointerEvent : MouseEvent" data-merge="PointerEventInit">
-                    <dt>readonly attribute long pointerId</dt>
+              <pre class="idl">
+dictionary PointerEventInit : MouseEventInit {
+    long        pointerId = 0;
+    double      width = 1;
+    double      height = 1;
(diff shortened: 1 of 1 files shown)


patrickhlauke:
note that as part of this, I removed the "Dictionary PointerEventInit Members" ... not sure it was adding anything that's not already in the actual WebIDL (unless I'm missing the finer point, in which case happy to add it back in and hook it up properly with the required `<dfn>` elements). Thoughts?

patrickhlauke:
also, this new WebIDL doesn't seem to give the nice auttomatic split of "Attributes" and "Methods" (see for instance https://w3c.github.io/pointerevents/#extensions-to-the-element-interface) - I take it this needs to be done manually in markup now? Again, happy to add this in before merging...

RByers:
Awesome, this looks great!  I did a side-by-side comparison and it looks better to be in every instance where things are different.  Let's land this baby now!

Issue #102, opened 2016-07-11
Current draft: the definitions of tiltY and twist is not visible"


mustaqahmed:
See the definitions of <code>tiltY</code> and <code>twist</code> around Figure 3 [here](https://w3c.github.io/pointerevents/#pointerevent-interface).

I suspect some problem with HTML tags.

patrickhlauke:
urgh, well spotted. yes, must be some problem in the markup which re-spec then mangles further. i'll have a look

PR #104, opened 2016-07-11
Fix broken WebIDL <dl>

patrickhlauke:
Missing explicit <p> resulted in Re-Spec completely ignoring/hiding the
actual text of the definition

Closes https://github.com/w3c/pointerevents/issues/102


RByers:
LGTM, thanks!

PR #108, opened 2016-07-19
fixing typo

dret:


patrickhlauke:
crikey...embarassing.

dret:
thanks for merging. also, what about making the abstract text-only? one of my pet peeves are the abstracts containing cross-references, and even worse cross-references that need to be read as text.

PR #116, opened 2016-07-25
Add detail=0 for pointerevents

### LINKED ISSUE #98: What should be the 'detail' property of pointer events?
From the UI Event spec, mouse events should have 'detail' property set to the click count. Do we expect the same for the pointer events?
Currently both Edge and Chrome set this property to 0 all the time.
I believe either way we need a note to define the value of this property somewhere.

NavidZ:
closes #98.

As discussed in one of the meeting I added a paragraph specifying detail should be 0 for all pointerevents defined in one of the sections.


RByers:
LGTM, thanks!
Before marking the bug as fixed, can you please also update some tests for this?

patrickhlauke:
LGTM too...but remove the `.index.html.swp` file that snuck into the commit

PR #119, opened 2016-07-27
Reference TouchEvents spec for compat mouse events

### LINKED ISSUE #7: Mouse Event compatibility model for touch is incompatible with current practice
Existing web code expects mouse events for touch only after the end of the event sequence ("gesture based mouse event compatibility").  The model defined by pointer events is incompatible with this.  This is especially problematic on a page that has components which both use touch and pointer events.

IE is already exploring doing this.  See "simplify mouse events for touch" in this [IE blog post](http://blogs.msdn.com/b/ie/archive/2015/02/24/pointer-events-w3c-recommendation-interoperable-touch-and-removing-the-dreaded-300ms-tap-delay.aspx).

NavidZ:
closes #7

### DIFF
--- a/index.html
+++ b/index.html
@@ -927,7 +927,7 @@ partial interface Navigator {
                 <li>If the pointer event dispatched was <code>pointerup</code> or <code>pointercancel</code>, clear the <i>PREVENT MOUSE EVENT</i> flag for this <code>pointerType</code>.</li>
             </ol>
             <div class="note">
-                <p>If the user agent supports both Touch Events (as defined in [[TOUCH-EVENTS]]) and Pointer Events, generation of compatibility mouse events as described in this section is not very web compatible. In this scenario gesture based generation of compatiblity mouse events for primary PointerEvents is better. For more information see [[TOUCH-EVENTS]].</p>
+                <p>If the user agent supports both Touch Events (as defined in [[TOUCH-EVENTS]]) and Pointer Events, the user agent SHOULD NOT generate compatibility mouse events as described in this section as it is likely to introduce compatibility problems for sites that expect mouse events to be generated in accordance with the <a href="https://www.w3.org/TR/touch-events/#mouse-events">model</a> outlined in [[TOUCH-EVENTS]].</p>
             </div>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -927,6 +927,9 @@
                 <li>If the pointer event dispatched was <code>pointerup</code> or <code>pointercancel</code>, clear the <i>PREVENT MOUSE EVENT</i> flag for this <code>pointerType</code>.</li>
             </ol>
             <div class="note">
+                <p>If the user agent supports both Touch Events (as defined in [[TOUCH-EVENTS]]) and Pointer Events, generation of compatibility mouse events as described in this section is not very web compatible. In this scenario gesture based generation of compatiblity mouse events for primary PointerEvents is better. For more information see [[TOUCH-EVENTS]].</p>
```
patrickhlauke: assuming this is aimed at UA developers, rather than content developers? "generation of compatibility mouse events as described in this section is not very web compatible" sounds a little off stylistically, and doesn't give me quite a clear indication of what we're trying to say. something more like "the user agent SHOULD NOT generate compatibility mouse events as described in this section, as this is likely to introduce compatibility problems for sites that expect mouse events to be generated in accordance with the model outlined in [[TOUCH-EVENTS]]." and then drop the last "For more info..." sentence?

patrickhlauke: "In this scenario gesture based generation of compatiblity mouse events for primary PointerEvents is better." do we actually define what "gesture-based generation" (I'd add a hyphen) is anywhere?

NavidZ: The problem is that the gesture is not defined anywhere and I believe it is out of the scope of this spec to define it either. So I believe we agreed on a handwavy definition of sending events based on gestures and just a link to the touch event spec is okay. I used the sentence you suggested and just got rid of the rest of it. That first sentence seem to capture the essence of the note.

patrickhlauke:
thanks for making the suggested edits @NavidZ - LGTM at this stage.

RByers:
Looks good, just updated the commit description.  Thanks!

PR #122, opened 2016-07-28
got/lostpointercapture events: attributes and immediate firing on implicit release

### LINKED ISSUE #113: Attributes of lost/gotpointercapture events fired immediately on set/release capture
This is blocked on #32. I am creating this new issue because @NavidZ's [comment](https://github.com/w3c/pointerevents/issues/32#issuecomment-232687047) there needs a bit more attention.

Since Edge is not delaying <code>lostpointercapture</code>/<code>gotpointercapture</code> in some cases and Chrome will be following a similar approach soon, the attribute values for, say, <code>button</code> or <code>pressure</code> seems to become irrelevant for these <code>lostpointercapture</code> and <code>gotpointercapture</code> events (i.e. for the ones fired immediately after a call to <code>setPointerCapture</code> or <code>releasePointerCapture</code>) because there is no hardware events to provide us the values here.

I see three possible ways to spec this:
A. Don't say anything, and expect UAs to pick up a consistent behavior (or expect developers rely on only the meaningful attributes).
B. These events set only the meaningful attributes (only <code>pointerId</code>?). All other attributes are either default or unknown.
C. These events always carry the attributes of the last known event for the <code>pointerId</code>, for both immediate firing and lazy firing in response to a real event.

Comments?

NavidZ:
Closes #113 and #32.

### DIFF
--- a/index.html
+++ b/index.html
@@ -400,9 +400,7 @@ interface PointerEvent : MouseEvent {
             <section>
                 <h2>Firing events using the <code>PointerEvent</code> interface</h2>
                 <p>To <dfn>fire a pointer event name e</dfn> means to <dfn>fire an event named e</dfn> as defined in [[!DOM4]] with an event using the <a>PointerEvent</a> interface whose attributes are set as defined in <a href="#pointerevent-interface"><code>PointerEvent</code> Interface</a>.</p>
-
-                <p>For any pointer event initialize the <code>composed</code> ([[!WHATWG-DOM]]) attribute to <code>true</code> and <a href="https://www.w3.org/TR/uievents/#widl-UIEvent-detail"><code>detail</code></a> [[!DOM-LEVEL-3-EVENTS]] attribute to 0.</p>
-                <p>Initialize the <code>bubbles</code> and <code>cancelable</code> attributes for the event according to the <a href="#pointer-event-type-table">Pointer Events table</a>.</p>
+                <p>For any pointer event initialize the attributes of the event according to the <a href="#h-pointer-event-types">Pointer Event types</a> section.</p>
                 <p>For <code>gotpointercapture</code> and <code>lostpointercapture</code> all the attributes except the ones defined in the table above should be the same as the Pointer Event that caused the user agent to run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> and fire the <code>gotpointercapture</code> and <code>lostpointercapture</code> events.</p>
  
                 <p>If the event is not <code>gotpointercapture</code> or <code>lostpointercapture</code>, run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps for this <code>PointerEvent</code>.
@@ -519,6 +517,7 @@ interface PointerEvent : MouseEvent {
                     </tr>
                 </tbody>
             </table>
+            <p>For all pointer events in the table above, <code>composed</code> ([[!WHATWG-DOM]]) attribute SHOULD be <code>true</code> and <a href="https://www.w3.org/TR/uievents/#widl-UIEvent-detail"><code>detail</code></a> [[!DOM-LEVEL-3-EVENTS]] attribute SHOULD be 0.</p>
             <p>In the case of the <a>primary pointer</a>, these events (with the exception of <code>gotpointercapture</code> and <code>lostpointercapture</code>) may also fire <a>compatibility mouse events</a>.</p>
             <section>
                 <h3><dfn>The <code>pointerover</code> event</dfn></h3>
--- a/index.html
+++ b/index.html
@@ -401,8 +401,6 @@ interface PointerEvent : MouseEvent {
                 <h2>Firing events using the <code>PointerEvent</code> interface</h2>
                 <p>To <dfn>fire a pointer event name e</dfn> means to <dfn>fire an event named e</dfn> as defined in [[!DOM4]] with an event using the <a>PointerEvent</a> interface whose attributes are set as defined in <a href="#pointerevent-interface"><code>PointerEvent</code> Interface</a>.</p>
                 <p>For any pointer event initialize the attributes of the event according to the <a href="#h-pointer-event-types">Pointer Event types</a> section.</p>
-                <p>For <code>gotpointercapture</code> and <code>lostpointercapture</code> all the attributes except the ones defined in the table above should be the same as the Pointer Event that caused the user agent to run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> and fire the <code>gotpointercapture</code> and <code>lostpointercapture</code> events.</p>
- 
                 <p>If the event is not <code>gotpointercapture</code> or <code>lostpointercapture</code>, run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps for this <code>PointerEvent</code>.
 
                 <p>The target object at which the event is fired is determined as follows:
@@ -518,6 +516,7 @@ interface PointerEvent : MouseEvent {
                 </tbody>
             </table>
             <p>For all pointer events in the table above, <code>composed</code> ([[!WHATWG-DOM]]) attribute SHOULD be <code>true</code> and <a href="https://www.w3.org/TR/uievents/#widl-UIEvent-detail"><code>detail</code></a> [[!DOM-LEVEL-3-EVENTS]] attribute SHOULD be 0.</p>
+                <p>For <code>gotpointercapture</code> and <code>lostpointercapture</code> all the attributes except the ones defined in the table above should be the same as the Pointer Event that caused the user agent to run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> and fire the <code>gotpointercapture</code> and <code>lostpointercapture</code> events.</p>
             <p>In the case of the <a>primary pointer</a>, these events (with the exception of <code>gotpointercapture</code> and <code>lostpointercapture</code>) may also fire <a>compatibility mouse events</a>.</p>
             <section>
                 <h3><dfn>The <code>pointerover</code> event</dfn></h3>
--- a/index.html
+++ b/index.html
@@ -519,7 +519,7 @@ interface PointerEvent : MouseEvent {
                     </tr>
                 </tbody>
             </table>
-            <p>In the case of the <a>primary pointer</a>, these events (with the exception of <code>gotpointercapture</code>, and <code>lostpointercapture</code>) may also fire <a>compatibility mouse events</a>.</p>
+            <p>In the case of the <a>primary pointer</a>, these events (with the exception of <code>gotpointercapture</code> and <code>lostpointercapture</code>) may also fire <a>compatibility mouse events</a>.</p>
             <section>
                 <h3><dfn>The <code>pointerover</code> event</dfn></h3>
                 <p>A user agent MUST <a href="#firing-events-using-the-pointerevent-interface">fire a pointer event</a> named <code>pointerover</code> when a pointing device is moved into the hit test boundaries of an element. A user agent MUST also fire this event prior to firing a <code>pointerdown</code> event for <a href=#mapping-for-devices-that-do-not-support-hover>devices that do not support hover</a> (see <code><a href="#the-pointerdown-event">pointerdown</a></code>).</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -423,6 +423,8 @@
                     <li><code>pointerout</code></li>
                 </ul>
                 
+                <p>Run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps for this <code>PointerEvent</code> if it is not <code>gotpointercapture</code> or <code>lostpointercapture</code>.
```
RByers: nit: reverse the sentence "If the event is not ..., run ..."

index.html
```diff
@@ -546,7 +548,7 @@
                 </tbody>
             </table>
             <p>In the case of the <a>primary pointer</a>, these events (with the exception of <code>gotpointercapture</code>, and <code>lostpointercapture</code>) may also fire <a>compatibility mouse events</a>.</p>
-            <p>For all PointerEvents in the table above, <a href="https://www.w3.org/TR/uievents/#widl-UIEvent-detail">detail</a> attribute defined in [[!DOM-LEVEL-3-EVENTS]] SHOULD be 0.</p>
+            <p>For all PointerEvents in the table above, <a href="https://www.w3.org/TR/uievents/#widl-UIEvent-detail"><code>detail</code></a> attribute defined in [[!DOM-LEVEL-3-EVENTS]] SHOULD be 0. For <code>gotpointercapture</code> and <code>lostpointercapture</code> all the attributes except the ones defined in the table above should be the same as the Pointer Event that causes the user agent to run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> and fire <code>gotpointercapture</code> and <code>lostpointercapture</code> events.</p>
```
RByers: I realize now that this bit about `detail` really fits better in the "firing events" section (since that's what's describing how events get fired, including setting other properties like `bubbles` and `cancelable`.  Mind moving it?

Then I think copying the properties for `got/lostpointercapture` make the most sense in the "process pending pointer capture" section.  In fact I think the property copying also applies to all the other events fired there (over,enter,leave,out) too, right?  So maybe just one sentence at the top of 5.1.3.1 indicating that this is done only in response to some other pointer event, and (unless indicated otherwise) all events fired in that section get their property values from the original event.  WDYT?

index.html
```diff
@@ -854,7 +856,7 @@
...
             <section>
                 <h3>Implicit Release of Pointer Capture</h3>
-                <p>Immediately after firing the <code>pointerup</code> or <code>pointercancel</code> events, a user agent MUST run the steps as if the <code>releasePointerCapture()</code> method has been called with an argument equal to the <code>pointerId</code> property of the <code>pointerup</code> or <code>pointercancel</code> event just dispatched.</p>
+                <p>Immediately after firing the <code>pointerup</code> or <code>pointercancel</code> events, a user agent MUST run step 4 in <a href="#releasing-pointer-capture">Releasing Pointer Capture</a> with the <code>pointerId</code> property of the <code>pointerup</code> or <code>pointercancel</code> event just dispatched and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code> if necessary.</p>
```
RByers: Rather than say "step 4" I think you can continue to say "run the steps", right?  It looks like they're written to handle  this case (it's kind of weird to jump into the middle of the steps, and the step numbers may change).  Alternately you can just say "clear the pending pointer capture override for the specified pointerId" if you think that's simpler.

RByers:
Big picture looks good to me, thanks!  Just a couple minor style / editorial suggestions.

index.html
patrickhlauke: nit: "that causes the user agent to run " > "that _caused_ the user agent to run"; "and fire &lt;code>gotpointercapture&lt;/code>" > "and fire _the_ &lt;code>gotpointercapture&lt;/code>"

index.html
patrickhlauke: nit: "event just dispatched and then run" > "event _that was_ just dispatched, and then run" (note comma)

NavidZ: The problem with those steps is that they expect an element to be passed to release pointer capture api. But previously it was not specified on which element release pointer capture steps should be run. Since the rest of the steps have no effect in this case I just mentioned step 4. I can certainly change it to the text if you think that is okay too.

NavidZ:
I moved the table to the other section and removed the "Composed" column from it as we do mention setting composed attribute to true for all the pointer events at first.
Regarding the attributes of got/lostpointercapture, I'm not sure if we split the attributes values into two sections makes things clearer. Right now they are all in this "Firing event.." section. Regarding that boundary events I believe we would be able to remove them all from that section in future PRs. Right?

index.html
RByers: You could just address that by augmenting "as if the releasePointerCapture() method has been called with an argument equal to the pointerId property of the pointerup or pointercancel event just dispatched." to include "on the pending pointer capture override element".  Either way is OK with me.

RByers:
Personally I think the table makes more sense in 5.2 as a summary of that section (listing all the events).  Having a line in the firing algorithm saying "initialize the properties according to this table" (link) seems adequate to me.  But I don't have a strong opinion, this is all just editorial.  @patrickhlauke  WDYT?

patrickhlauke:
> Personally I think the table makes more sense in 5.2 as a summary of that section (listing all the events). Having a line in the firing algorithm saying "initialize the properties according to this table" (link) seems adequate to me.

yeah, i'd agree with that

index.html
```diff
@@ -494,62 +473,53 @@
...
                         <td><code>lostpointercapture</code></td>
                         <td>Sync/Async</td>
                         <td>Yes</td>
                         <td>No</td>
-                        <td>Yes</td>
                         <td>None</td>
                     </tr>
                 </tbody>
             </table>
             <p>In the case of the <a>primary pointer</a>, these events (with the exception of <code>gotpointercapture</code>, and <code>lostpointercapture</code>) may also fire <a>compatibility mouse events</a>.</p>
```
mustaqahmed: Please get rid of the comma before "and lostpointercamture...".

index.html
```diff
@@ -401,29 +401,11 @@
...
+                <p>For any pointer event initialize the <code>composed</code> ([[!WHATWG-DOM]]) attribute to <code>true</code> and <a href="https://www.w3.org/TR/uievents/#widl-UIEvent-detail"><code>detail</code></a> [[!DOM-LEVEL-3-EVENTS]] attribute to 0.</p>
+                <p>Initialize the <code>bubbles</code> and <code>cancelable</code> attributes for the event according to the <a href="#pointer-event-type-table">Pointer Events table</a>.</p>
+                <p>For <code>gotpointercapture</code> and <code>lostpointercapture</code> all the attributes except the ones defined in the table above should be the same as the Pointer Event that caused the user agent to run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> and fire the <code>gotpointercapture</code> and <code>lostpointercapture</code> events.</p>
+ 
+                <p>If the event is not <code>gotpointercapture</code> or <code>lostpointercapture</code>, run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps for this <code>PointerEvent</code>.
```
mustaqahmed, line 430: This para seems odd given that this section talks mostly about setting attributes. I suggest getting rid of this para, instead grouping all paras except the first one (the definition of "firing") in this block into a new section "5.1.3.1 Initializing attributes". The  "Process Pending Pointer Capture" section will then become 5.1.3.2---the obvious next step.

index.html
```diff
@@ -401,29 +401,11 @@
...
-                    <li><code>pointerout</code></li>
-                </ul>
-                
-                <p>Run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps for this <code>PointerEvent</code> if it is not <code>gotpointercapture</code> or <code>lostpointercapture</code>.
+                <p>For any pointer event initialize the <code>composed</code> ([[!WHATWG-DOM]]) attribute to <code>true</code> and <a href="https://www.w3.org/TR/uievents/#widl-UIEvent-detail"><code>detail</code></a> [[!DOM-LEVEL-3-EVENTS]] attribute to 0.</p>
+                <p>Initialize the <code>bubbles</code> and <code>cancelable</code> attributes for the event according to the <a href="#pointer-event-type-table">Pointer Events table</a>.</p>
+                <p>For <code>gotpointercapture</code> and <code>lostpointercapture</code> all the attributes except the ones defined in the table above should be the same as the Pointer Event that caused the user agent to run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> and fire the <code>gotpointercapture</code> and <code>lostpointercapture</code> events.</p>
```
mustaqahmed: Replace "table above" with the link too. On my first reading, I scrolled up to find the buttons table!

NavidZ:
Anyone else has any comment about this PR?

mustaqahmed:
Am I the only one to think that the mix of "attribute setting" and "event firing" between section headers 5.1.3 and 5.1.3.1 [could be made cleaner](https://github.com/w3c/pointerevents/pull/122#discussion_r73200829)?

NavidZ:
I personally agree with you. Do you think just moving all attribute values to 5.2 along with that table and reference that section for all attributes be better? Regarding firing events in 5.1.3 vs 5.1.3.1, they are a bit different I believe. They are 2 different routines that call into each other in some situations. WDTY?

mustaqahmed:
> I personally agree with you. Do you think just moving all attribute values to 5.2 along with that table and reference that section for all attributes be better?

Yes, that should be cleaner than referencing that table twice before it appears.

> Regarding firing events in 5.1.3 vs 5.1.3.1, they are a bit different I believe. They are 2 different routines that call into each other in some situations. WDTY?

The mix of "setting attributes" and "firing details" looked a bit odd to me. Your suggestion above would reduce the "setting" part in 5.1.3 into a one-line reference to 5.2, which seems better. So, please don't split 5.1.3.

NavidZ:
Does this look okay?

index.html
```diff
@@ -400,9 +400,7 @@
...
-
-                <p>For any pointer event initialize the <code>composed</code> ([[!WHATWG-DOM]]) attribute to <code>true</code> and <a href="https://www.w3.org/TR/uievents/#widl-UIEvent-detail"><code>detail</code></a> [[!DOM-LEVEL-3-EVENTS]] attribute to 0.</p>
-                <p>Initialize the <code>bubbles</code> and <code>cancelable</code> attributes for the event according to the <a href="#pointer-event-type-table">Pointer Events table</a>.</p>
+                <p>For any pointer event initialize the attributes of the event according to the <a href="#h-pointer-event-types">Pointer Event types</a> section.</p>
                 <p>For <code>gotpointercapture</code> and <code>lostpointercapture</code> all the attributes except the ones defined in the table above should be the same as the Pointer Event that caused the user agent to run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> and fire the <code>gotpointercapture</code> and <code>lostpointercapture</code> events.</p>
```
mustaqahmed: This looks cleaner now, thanks. LGTM except for one last comment: please fix the "table above" here by moving this para to somewhere around Line 520.

NavidZ: Oops. Thanks for the catch. I missed this one. Done.

mustaqahmed:
LGTM.

index.html
```diff
@@ -400,29 +400,9 @@
             <section>
                 <h2>Firing events using the <code>PointerEvent</code> interface</h2>
                 <p>To <dfn>fire a pointer event name e</dfn> means to <dfn>fire an event named e</dfn> as defined in [[!DOM4]] with an event using the <a>PointerEvent</a> interface whose attributes are set as defined in <a href="#pointerevent-interface"><code>PointerEvent</code> Interface</a>.</p>
+                <p>For any pointer event initialize the attributes of the event according to the <a href="#h-pointer-event-types">Pointer Event types</a> section.</p>
```
RByers: I think would be clearer if merged the previous paragraph since it's really a continuation of the theme of initializing the attributes.  Maybe: "... whose attributes are set as defined in PointerEvent interface, and in Pointer Event types for the type e.".

Also while you're editing that line, can you fix the typo: "fire a pointer event name e" -> "fire a pointer event named e"?

RByers:
LGTM

RByers:
Oh one more thing I almost forgot again before merging - add a ChangeLog entry with a link to this PR.

RByers:
Great except now you have a merge conflict - need to rebase your branch

PR #127, opened 2016-08-01
Add hasPointerCapture

### LINKED ISSUE #121: Add Element.hasPointerCapture API
As [discussed at the hackathon](https://docs.google.com/document/d/1b8aJOJcGXakstFJslKl87QwvlFHhHwGbQfJwnUTBBHg/edit#), if  we end up going with the reduced hit-test model (#8) then there may be scenarios where some library will want behave differently depending on whether the pointer has been captured, without actually changing the capture itself.

For example, some analytics libraries track "user attention" by recording a log of the position and interesting underlying elements of every pointer.  If touch starts capturing implicitly, it's possible the stats from such a library could be broken.  To mitigate this they might want to do their own `elementFromPoint` hit-test on `pointermove` without actually doing a `releasePointerCapture`.

Really this is a more generic problem that in composition scenarios one library may want to know if the pointer is currently captured or not.  It's not clear how important this is, but it seems to make a lot of sense to have an API like `Element.hasPointerCapture(pointerId)` which pointer event listeners could check on the target of the event.

RByers:
Fixes #121

Also cleans up a couple related definitions for easier linking.

### DIFF
--- a/index.html
+++ b/index.html
@@ -993,6 +993,7 @@ partial interface Navigator {
         <p>The following is an informative summary of substantial and major editorial changes between publications of this specification, relative to the first [[PointerEvents]] specification. See the <a href="https://github.com/w3c/pointerevents/commits">complete revision history of the Editor's Drafts of this specification</a>.</p>
         <h3>Changes since the 24 February 2015 Recommendation</h3>
         <ul>
+            <li><a href="https://github.com/w3c/pointerevents/pull/127">Add hasPointerCapture</a></li>
             <li><a href="https://github.com/w3c/pointerevents/pull/96">Removed "pen contact" condition on button/buttons</a></li>
             <li><a href="https://github.com/w3c/pointerevents/pull/92">Make all pointer events composed events</a></li>
             <li><a href="https://github.com/w3c/pointerevents/pull/87">Add digitizer/pen tangential (barrel) pressure</a></li>
@@ -1006,7 +1007,6 @@ partial interface Navigator {
             <li><a href="https://github.com/w3c/pointerevents/pull/34">Clarify the button value for mouse drag</a></li>
             <li><a href="https://github.com/w3c/pointerevents/pull/24">Fix the touch-action processing model for zoom scenarios</a></li>
             <li><a href="https://github.com/w3c/pointerevents/pull/13">Add direction-specific touch-action values</a> (pan-left, pan-right, pan-up, pan-down) and clarified behavior of existing pan-x and pan-y values.</li>
-            <li><a href="https://github.com/w3c/pointerevents/pull/127">Add hasPointerCapture</a></li>
         </ul>
     </section>
     <!-- appendix -->
--- a/index.html
+++ b/index.html
@@ -652,7 +652,7 @@ partial interface Element {
                 <dt><dfn>hasPointerCapture</dfn></dt>
                 <dd>
                     <p>Indicates whether the element on which this method is invoked has <a>pointer capture</a> for the pointer identified by the argument <code>pointerId</code>.  In particular, returns <code>true</code> if the <a>pending pointer capture target override</a> for <code>pointerId</code> is set to the element on which this method is invoked, and <code>false</code> otherwise.</p>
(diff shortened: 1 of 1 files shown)


RByers:
[Preview](https://rawgit.com/RByers/w3c-pointerevents/hasPointerCapture/index.html#dom-element-haspointercapture).
@patrickhlauke @NavidZ @scottgonzalez @teddink any feedback?

index.html
```diff
@@ -1002,6 +1006,7 @@
             <li><a href="https://github.com/w3c/pointerevents/pull/34">Clarify the button value for mouse drag</a></li>
             <li><a href="https://github.com/w3c/pointerevents/pull/24">Fix the touch-action processing model for zoom scenarios</a></li>
             <li><a href="https://github.com/w3c/pointerevents/pull/13">Add direction-specific touch-action values</a> (pan-left, pan-right, pan-up, pan-down) and clarified behavior of existing pan-x and pan-y values.</li>
+            <li><a href="https://github.com/w3c/pointerevents/pull/127">Add hasPointerCapture</a></li>
```
patrickhlauke: the revision history is in reverse chronological order, so this should be at the very top of the list

index.html
```diff
@@ -549,33 +549,33 @@
             <p>For all PointerEvents in the table above, <a href="https://www.w3.org/TR/uievents/#widl-UIEvent-detail">detail</a> attribute defined in [[!DOM-LEVEL-3-EVENTS]] SHOULD be 0.</p>
 
             <section>
-                <h3><dfn>The <code>pointerover</code> event</dfn></h3>
+                <h3>The <dfn><code>pointerover</code> event</dfn></h3>
```
patrickhlauke, line 552: nit: if it's not too much trouble, possible to split out these fixes to `<dfn>` into a separate standalone PR or commit?

patrickhlauke:
makes sense to me in principle. random pondering: would we also want a function that lets me query which pointer(s) the element i'm calling the function on has capture for? or is that getting a bit too convoluted?

index.html
```diff
@@ -642,13 +643,16 @@
...
+                </dd>
+                <dt><dfn>hasPointerCapture</dfn></dt>
+                <dd>
+                    <p>Indicates whether the element on which this method is invoked has <a>pointer capture</a> for the pointer identified by the argument <code>pointerId</code>.  In particular, returns <code>true</code> if the <a>pending pointer capture target override</a> for <code>pointerId</code> is set to the element on which this method is invoked, and <code>false</code> otherwise.</p>
+                    <div class="note">This method will return true immediately after a call to <a>setPointerCapture</a>, even though that element will not yet have received a <a>gotpointercapture event</a>.  As a result it can be useful to detecting <a>implicit pointer capture</a> from inside of a <a>pointerdown event</a> listener.</div>
```
scottgonzalez: "useful to detecting" -> "useful _for_ detecting" or "useful to _detect_"

scottgonzalez:
> makes sense to me in principle. random pondering: would we also want a function that lets me query which pointer(s) the element i'm calling the function on has capture for? or is that getting a bit too convoluted?

Do you have a scenario in mind where you would want to know this? We expect this would be used inside an event and you'd be querying against the single pointer ID for that event.

patrickhlauke:
> Do you have a scenario in mind 

admittedly no, i don't have a concrete use case in mind - which is probably a good indicator that this would be superfluous.

RByers:
> > Do you have a scenario in mind
> 
> admittedly no, i don't have a concrete use case in mind - which is probably a good indicator that this would be superfluous.

Yeah I think the use cases even for this simple API are pretty limited.  I think we should add it mainly as a hedge - it makes conceptual sense, has nice symmetry with the existing APIs and so adds very little overall complexity, and there's some reason to believe there may be non-zero value.  But for anything more complex I think we should wait for more compelling use cases.  Easy to add later though!

index.html
RByers: Whoops, thanks.

index.html
RByers, line 552: Done in #128

index.html
RByers: Done, thanks.

RByers:
Suggestions applied, any other feedback?

patrickhlauke:
Looking good to me, fella

RByers:
Ok, thanks!  Let's land this now then (so I can more easily do the next step) - happy to take additional feedback after the fact.

PR #128, opened 2016-08-02
Cleanup event definitions for easier linking

RByers:
As discussed [here](https://github.com/w3c/pointerevents/pull/127#discussion_r73036237)


RByers:
I'm just gonna merge this trivial editorial change to unblock #127.  Note that I didn't bother adding new links wherever an event name is referenced, we can do that separately when there's value.

patrickhlauke:
perfect thanks

Issue #8, opened 2015-04-17
Pointermove should not require a hit-test by default for touch


RByers:
From https://lists.w3.org/Archives/Public/public-pointer-events/2015JanMar/0041.html:

>  One of the [main concerns I cited](https://lists.w3.org/Archives/Public/public-pointer-events/2014JulSep/0051.html)
> with our decision not to implement is that the uncaptured-by-default model
> for touch input places a performance burden on the engine.  Since the
> dominant input APIs on touch devices (Touch Events, Android and iOS native
> APIs) don’t have this property, we are unwilling to adopt an API with this
> disadvantage, no matter how small.  In addition to the performance
> concerns, we believe an implicit-capture model encourages a style of UX we
> feel is most appropriate for direct-manipulation input devices - where, by
> default, the user is manipulating a specific object, not touching a layer
> of glass above all objects.
> 
> I feel the simplest way to address these concerns would be to make the
> following two changes to the API.  First, allow an implementation to
> implicitly capture touch (and perhaps pen) input to the target node on
> pointerdown (note that the spec already technically permits the browser to
> choose to capture implicitly since IE does this for button elements for
> example).  Developers can still use the explicit capture APIs to recapture
> elsewhere (as is often necessary to achieve the desired UX), or return to
> uncaptured events.  Second, add an optional parameter (false if
> unspecified) to setPointerCapture which indicates whether
> pointerover/out/enter/leave events are desired during capture.  When false,
> an implementation can avoid hid-testing for every move during capture.
> 
> Of course these are breaking changes, and it’s likely they could be a
> significant compatibility problem for existing websites.  I’m interested in
> implementing this model behind a flag and collecting data about the
> compatibility implications.  Perhaps it’s possible that Chrome could ship
> this with acceptable compat pain and move the web to using explicit APIs to
> indicate developer intent.  Then other PE implementations could update
> without much trouble.  I’m also willing to explore other ideas for
> mitigating the compatibility risk within our requirement that developers
> will be unlikely to accidentally incur per-move hit-test costs without
> explicitly opting into them.

scottgonzalez:
This would only apply while a pointing device is down, correct? In the case of touch, you don't get move events unless there's a down action, so this would cover the existing Touch Events API scenarios. However, when using a mouse, or hovering with a stylus, you need hit tests to determine enters and leaves.

patrickhlauke:
[...]

PR #129, opened 2016-08-02
Add initial implicit capture proposal

RByers:
Adds an initial description of chromium's "implicit capture" behavior as discussed in w3c/pointerevents#8.


RByers:
Note that this is a pull request against a new "reduce-hit-tests" branch as discussed in w3c/pointerevents#8.  I imagine we'll want to continue to iterate on related changes in the branch (for example, this doesn't yet contain the fix for w3c/pointerevents#566).  So I don't think we need to worry about this being perfect from the start, I just want to get an initial framework in place for describing the chromium implementation so that folks can continue to iterate on it (eg. while I'm on vacation until Aug 22nd).

@NavidZ @scottgonzalez @patrickhlauke @teddink WDYT, does this look like a reasonable place to start on the branch?

patrickhlauke:
Looks reasonable to me, yes

NavidZ:
lgtm too.

index.html
```diff
@@ -856,13 +858,16 @@
...
+            <p>Direct manipulation devices should behave exactly as if <a data-lt="Element.setPointerCapture">setPointerCapture</a> was called on the target element just before the invocation of any <code>pointerdown</code> listeners.  The <a data-lt="Element.hasPointerCapture">hasPointerCapture</a> API may be used (eg. within any <code>pointerdown</code> listener) to determine whether this has occurred.  If <a data-lt="Element.releasePointerCapture">releasePointerCapture</a> is not called for the pointer before the next pointer event is fired, then a <a>gotpointercapture event</a> will be dispatched to the target (as normal) indicating that capture is active.
```
scottgonzalez: These two paragraphs are missing closing tags.

scottgonzalez:
One minor nit, but otherwise this looks good.

index.html
RByers: Thanks!  Fixed.

Issue #94, opened 2016-06-22
Add feature detection section


RByers:
Discussed on the call today that we should add a non-normative section about suggested feature detection.  Perhaps something like this?

``` javascript
  if (window.PointerEvent) {
    // Listen only for pointer events when they're supported.
    // Don't bother listening for mouse/touch as they're redundant and will only cause confusion.
    target.addEventListener("pointerdown", ondown);
    ...
  } else {
    // Listen for both touch and (non-touch-derived) mouse events.  No need for feature detection.
    target.addEventListener("touchstart", ondown);
    target.addEventListener("mousedown", function(e) {
      if (!e.sourceCapabilities || !e.sourceCapabilities.firesTouchEvents)
        ondown(e);
      });
    ...
  }
```

@patrickhlauke says he'll write up a PR to discuss further.

NekR:
If v2 will be different than v1 (incompatibilities), how can one detect the new version without UA checks?

patrickhlauke:
As v1 is only really implemented in IE11 and early versions of Edge, and v2 is effectively a realignment of what Edge does now and what Chrome and Mozilla are targetting, I don't think this will be a huge problem.

patrickhlauke:
(sorry, accidentally closed this issue by accident)

NekR:
Okay, just to clear some things.
Imagine my website has support of both mouse and touch events. Now I wan't to add Pointer Events (v2) support because it's better, simpler, etc. So I am doing this:

``` js
if (window.PointerEvent) {
  initNewInputSystem();
} else {
  initOldInputSystem();
}
```

Question is: will my Pointer Events v2 implementation be _fully_ compatible with old Pointer Events v1 (IE11/earl Edge) without any hacks/changes?

patrickhlauke:
As Edge is evergreen, it likely then comes down to: will my code work in IE11. In which case: if the code does rely on anything substantively different as per https://w3c.github.io/pointerevents/#revision-history then you will likely have issues (unless it's one of the changes to the spec that are more clarifications of things that even IE11 already did differently from the nominal Level 1 spec). @teddink may be able to clarify if the behavior of PE in IE11 may still receive updates or not. @rbyers may have further thoughts on whether some method to explicitly distinguish between Level 1 and Level 2 support is required or not.

RByers:
[...]

PR #130, opened 2016-08-04
add feature detection example, clean up code examples

patrickhlauke:
Closes https://github.com/w3c/pointerevents/issues/94

### DIFF
--- a/index.html
+++ b/index.html
@@ -132,40 +132,34 @@
     <section class="informative">
         <h1>Examples</h1>
         <p>The following are example author code that demonstrates how the APIs in this specification might be used.</p>
-<pre id="example_1" class="example" title="Basic HTML5 Canvas painting application">
-&lt;style&gt;
-  /* Disable intrinsic user agent touch behaviors (such as panning or zooming) so 
-  that all events on the canvas element are given to the application instead. */
-  
-  canvas { 
-    touch-action: none; 
-  }
-&lt;/style&gt;
+<pre id="example_1" class="example" title="Feature detection and event binding">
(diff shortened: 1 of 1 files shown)


patrickhlauke:
also adds a "don't forget about keyboard users" bit to the canvas drawing example (doesn't go into details, but as a gentle reminder of the note about not forgetting kbd users in the spec's intro)

index.html
```diff
@@ -132,79 +132,106 @@
...
+    ...
+    target.addEventListener('mousedown', ...);
+    ...
+}
+
+// additional event listeners for keyboard handling
+...</code>
+</pre>
+<pre id="example_2" class="example" title="Detecting the type of input from a user">
+<code>window.addEventListener("pointerdown", detectInputType, false);
```
NavidZ: I can see this is the same as before. But do we need this "false" as the third parameter here at all? Just to make this line simpler I guess maybe we can get rid of it and not giving the false impression of this having anything to do with pointer types? WDYT?
This also appeared in a few other examples.

NavidZ:
Looks good to me overall.

index.html
patrickhlauke: agree, this somehow snuck into the original and stayed around. unnecessary in modern browsers...i'll nuke it (and any other instances i find). good catch

patrickhlauke:
merging this later today, unless @RByers @scottgonzalez @teddink have any comments?

teddink:
LGTM - thanks Patrick!

RByers:
Thanks for this - retroactive post-vacation LGTM

PR #137, opened 2016-08-29
Fix the wording in releasePointerCapture to use hasPointerCapture

### LINKED ISSUE #136: releasePointerCapture right after setPointerCapture is no-op
setPointerCapture sets 'pending pointer capture target override', but
releasePointerCapture doesn't clear it if 'pointer capture target override' isn't set.

NavidZ:
closes #136


smaug----:
looks ok to me

mustaqahmed:
LGTM.

patrickhlauke:
LGTM thanks

Issue #22, opened 2015-09-18
Add an API to get pre-coalesced points


RByers:
When a web app is slow, most browsers will coalesces multiple movement events into a single one (with the latest position only).  This is perfect for scenarios that care most about minimizing latency, but unfortunate for scenarios (like art applications) where maximum motion detail is required.  Similarly, many OSes will artificially align input to v-sync for maximum motion smoothness, but for drawing scenarios it's really the raw points the app wants.

Android [MotionEvents](http://developer.android.com/reference/android/view/MotionEvent.html) are similarly batched and expose a history API (eg. [getHistoricalX](http://developer.android.com/reference/android/view/MotionEvent.html#getHistoricalX%28int,%20int%29)) to get all the original raw points.  iOS 9 adds a `coalescedTouchesForTouch` API for this ([sample code](https://developer.apple.com/library/prerelease/ios/samplecode/TouchCanvas/Listings/TouchCanvas_CanvasView_swift.html#//apple_ref/doc/uid/TP40016561-TouchCanvas_CanvasView_swift-DontLinkElementID_6)).

Perhaps `PointerEvent` should have a similar API?  Eg. `getCoalescedEvents` that returns an array of `PointerEvent` instances for the same pointerId?

I don't think this is particularly urgent (the art scenarios are fairly niche), but may become so if we want to block #9 on this.

patrickhlauke:
Purely from a layman's "I like to make funky things with JS" point of view, I'm intrigued by this. Presumably, I'd be easy to then iterate over the array in sequence (i.e. the array has a defined chronological order for each `PointerEvent` contained inside)?

RByers:
Yep, roughly instead of:

``` javascript
canvas.addEventListener('pointermove', addPoint)
```

You'd have something like:

``` javascript
canvas.addEventListener('pointermove', e => {
  for(let point of e.getCoalescedPoints()) {
    addPoint(point);
  }
}
```

On a device with a 120hz touch scan rate (eg. Nexus 5) you'd get lines that have twice as many points.

RByers:
One risk here is that people would end up using this for other (more common) scenarios where it adds only wasted work (eg. dragging something around) and may decrease smoothness.  Perhaps we'd want to choose the API name carefully to help reduce the risk of this (maybe Android's "historical" term is better for this than iOS's more technically descriptive "coalesced").

RByers:
[...]

PR #149, opened 2016-10-05
Initial draft of getCoalescedEvents API

NavidZ:
Addresses #22


index.html
```diff
@@ -233,6 +233,16 @@
   });
 eventTarget.dispatchEvent(event);</code>
 </pre>
+<pre id="example_6" class="example" title="Drawing using coalesced events">
+<code>window.addEventListener("pointermove", function(event) {
+  var listOfAllEvents = event.getCoaloescedEvents();
```
scottgonzalez: Typo in method name.

index.html
```diff
@@ -233,6 +233,16 @@
...
+  listOfAllEvents.push(event);
+  for (var e of listOfAllEvents) {
```
scottgonzalez: If we're using `for of` should we use `let`?

index.html
```diff
@@ -363,6 +375,10 @@
                         <dd>
                             <p>Indicates if the pointer represents the <a>primary pointer</a> of this pointer type.</p>
                         </dd>
+                    <dt><dfn>getCoalescedEvents</dfn></dt>
+                        <dd>
+                            <p>Returns a sequence of all <code>PointerEvents</code> that were coalesced into the event. The coalesced events will have the same value as the event in attributes like <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>, and particularly <code>target</code>. But other attributes might be different. The event itself will have the greatest <a href="https://dom.spec.whatwg.org/#dom-event-timestamp">timeStamp</a> ([[!WHATWG-DOM]]) among the events in the sequence and the first event in the sequence will have the smallest timeStamp. None of the events in the returned sequence will have coalesced events. When an event has no coalesed event this function returns an empty sequence.</p>
```
scottgonzalez: Suggested wording change from "The coalesced events will have the same value as the event in attributes like ..." to "The following attributes of the coalesced events will always have the same value as the event: ..."

I wonder if there's a better way to refer to the primary event object to make "the event" less confusing when used in the same sentence as "coalesced events."

scottgonzalez (changes requested):

index.html
NavidZ: Yeah. I didn't like "the event" myself either. Is "the original event" any better?

scottgonzalez: I think that's better.

scottgonzalez (approved):
Looks good.

index.html
```diff
@@ -377,7 +377,7 @@
...
+                            <p>Returns a sequence of all <code>PointerEvents</code> that were coalesced into the (original) event. The following attributes of the coalesced events will always have the same value as the original event: <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>, and particularly <code>target</code>. But other attributes might be different. The events in the sequence will have increasing <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a> ([[!WHATWG-DOM]]). So the first event will have the smallest <code>timeStamp</code>. Also the original event's <code>timeStamp</code> will be greater than or equal to the last event's <code>timeStamp</code> in the sequence. None of the events in the sequence will have coalesced events. When an event has no coalesed event this function returns an empty sequence.</p>
```
mustaqahmed: Sorry for missing this before: the word "origin" is synonymous to "historically first", so I found it a bit misleading. Does "received event" avoid this ambiguity?

mustaqahmed: Nit: the last sentence could have a comma before "this function".

mustaqahmed: I think we should clarify the "global" event ordering too. Without any coalescing, it is assumed that event listeners are fired in the order input events occur. Coalescing adds buffering to it, breaking the assumption. The spec should clarify that despite the buffering, the ordering the "pointermove" handlers are triggered after the corresponding "pointerup" and before the corresponding "pointerdown"/"pointercancel".

I believe all we need is adding a normative test somewhere saying that received events' timestamps are non-decreasing across all pointers. Please correct me if I missed the related spec---I couldn't find any.

mustaqahmed: Another thought: do we need a note that two events for different pointerIds could contain interleaved historical events? E.g. the input sequence:
  [p1-move1, p2-move1, p2-move2, p1-move2]
could be received as:
  [p2-move2 (containing coalesced p2-move1), p1-move2 (containing p1-move1)].
This may not be obvious from the current text.

NavidZ:
I created another html as per suggestion in the mailing list for this extension. I also added an entry to PointerEventInit to enable javascript to create PointerEvents with coalesced events. Let me know what you think and whether you have better names in mind for some of the fields.

extension.html
```diff
@@ -0,0 +1,119 @@
...
+    </section>
+    <section dfn-for="extensions-to-pointerevents-interface">
+      <h2>Extensions to the <code>PointerEvent</code> interface</h2>
+      <p>
+        The following section describes extensions to the existing <a href="https://w3c.github.io/pointerevents/index.html#pointerevent-interface"><code>PointerEvent</code> interface</a>.
+      </p>
+      <div>
+          <pre class="idl">
+partial dictionary PointerEventInit {
+    sequence&lt;PointerEventInit> coalescedEventInits = [];
```
RByers: I can see how this coalescedEventInits pattern would be convenient for developers writing events from scratch, but I don't think it's worth worrying about that too much (it's a very niche scenario).  I think it would be unusual for a constructor to involve logic like this, normally they just populate properties.  So I suggest we change this to instead be `sequence<PointerEvent> coalescedEvents = [];`

extension.html
```diff
@@ -0,0 +1,119 @@
...
+      </div>
+    </section>
+
+    <section class="informative">
+      <h2>Examples</h2>
+      <p>The following are examples that demonstrate how the APIs in this specification might
+      be used</p>
+<pre id="example_1" class="example" title="Creating an untrusted pointer event with coalesced events from script">
+<code>
+var event = new PointerEvent("pointerover",
```
RByers: I'd suggest showing an example where you re-use an init dict, modifying it slightly to construct each event in the sequence and then the final event.

extension.html
```diff
@@ -0,0 +1,119 @@
...
+// event.getCoalescedEvents()[1].clientX is 305
+// event.getCoalescedEvents()[1].clientY is 505
+</code>
+</pre>
+<pre id="example_2" class="example" title="Drawing using coalesced events">
+<code>window.addEventListener("pointermove", function(event) {
+  var listOfAllEvents = event.getCoalescedEvents();
+  listOfAllEvents.push(event);
+  for (let e of listOfAllEvents) {
+    // Draw point for event e
```
RByers: nit, maybe make this more concrete with code like `drawPoint(e.pageX, e.pageY)`.

extension.html
```diff
@@ -0,0 +1,119 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset='utf-8'>
+    <title>PointerEvents - Level 2 - Extensions</title>
+    <link rel="stylesheet" type="text/css" href="https://w3c.github.io/uievents/spec-WD.css">
+    <script src='https://www.w3.org/Tools/respec/respec-w3c-common.js' class='remove'></script>
+    <script class='remove'>
+      var respecConfig = {
+        specStatus: "CG-DRAFT",
```
RByers, line 10: I think you're missing something in the metadata here.  The [status](https://rawgit.com/NavidZ/pointerevents/b9f39bd6a3d1011436a45745035a2d0d54be3838/extension.html#sotd) says "This specification was published by the .".  You can use "Pointer Events Working Group" for now (eventually we'll have a CG to put here).

extension.html
```diff
@@ -0,0 +1,119 @@
...
+  <body>
+    <section id='abstract'>
+      <p>
+The features in this specification extend or modify those found in <a href="https://w3c.github.io/pointerevents/index.html">Pointer Events Level 2</a> that describes events and related interfaces for handling hardware agnostic pointer input from devices including a mouse, pen, touchscreen, etc.
+      </p>
+    </section>
+
+    <section id='sotd'>
+      <p>
+        This proposal was first introduced in this <a href="https://discourse.wicg.io/t/new-event-for-non-primary-button-click/1527">discourse</a>.
```
RByers: This is from `auxclick` - remove.

extension.html
```diff
@@ -0,0 +1,119 @@
...
+      </p>
+    </section>
+
+    <section id='intro' class='informative'>
+      <h2>Introduction</h2>
+      <p>
+        This document mainly contains the new APIs that may not make it to Pointer Event Level 2 spec. So we iterate over these new APIs to make them ready for the next version.
```
RByers: How about:
"This document reflects incubation for APIs which are intended to be part of a future version of the Pointer Events specification.  They are not yet broadly supported by implementations.".

extension.html
```diff
@@ -0,0 +1,119 @@
...
+      </p>
+      <p>
+	One of the APIs that are added in this document exposes all the events that were coalesced into a single event. Coalescing events into a single event enables user agents to prevent processing of the events that don’t get a chance to update the frame content visually. Basically if the javascript handlers process the events and update the screen contents faster than refresh rate only the last update will be rendered and shown in the next frame. However, some applications like drawing apps might want to take advantage of the a precise history of events as user moves the pointer on the screen to be able to draw smoother and better looking curves. So we needed an API to expose the coalesced events along with the event that was dispatched to the javascript.
```
RByers: nit: s/one of the APIs that are added in this document/The getCoalescedEvents API/

extension.html
```diff
@@ -0,0 +1,119 @@
...
+              </dd>
+          </dl>
+          <p>The <code>PointerEventInit</code> dictionary is used by the <code>PointerEvent</code> interface's constructor to provide a mechanism by which to construct untrusted (synthetic) pointer events. It inherits from the <code>MouseEventInit</code> dictionary defined in [[!DOM-LEVEL-3-EVENTS]]. The steps for constructing an event are defined in [[!DOM4]]. The <coalescedEventInits> will be used to create coalesced PointerEvents of the generated untrusted synthetic pointer event. The number of coalesced events will be the same as <code>coalescedEventInits</code> list and all of their attributes will be have the same values as the generated event unstrusted synthetic pointer event except those that have a different value in their coalescedEventInits entry. See the <a href="#examples" title="examples">examples</a> for sample code demonstrating how to fire an untrusted pointer event.</p>
+          </p>
+      </div>
```
RByers, line 79: Do you think it's worth having a note somewhere that links to documentation for the equivalent APIs in Windows, iOS and Android?  I think that could be helpful in understanding the context for this API.

RByers (commented):
Looks good overall, just one non-trivial suggestion.

extension.html
```diff
@@ -0,0 +1,119 @@
...
+    </section>
+
```
mustaqahmed, line 81: Here is my suggestion on how to define the "global" event timing after coalescing. The main idea is that the UA can't hold back a particular event stream (say for a given pointerId) indefinitely to fire a coalesced event later on (see my old comments [here](https://github.com/w3c/pointerevents/pull/149#discussion_r83735181) and [here](https://github.com/w3c/pointerevents/pull/149#discussion_r83735563).

Add a new section here (before Examples), something like "Timing of coalesced event firing", covering the following points:

A. For a given pointerId, the ordering of the fired events is consistent with respect to the event-types. For example, when the UA decides to fire a "pointerup" event, it should first fire the "pointerdown" if still pending, then coalesce & send all pending "pointermove" events before firing the "pointerdown" event.

B. If the UA decides to coalesce & fire one pending event sequence for some reason (e.g. for A above or before frame update), it MUST do so for all pending event sequences (i.e. across pointerIds & types).

extension.html
```diff
@@ -0,0 +1,119 @@
...
+};
+
+partial interface PointerEvent {
+    sequence&lt;PointerEvent> getCoalescedEvents();
+};
+          </pre>
+          <dl dfn-for="PointerEvent" link-for="PointerEvent">
+              <dt><dfn>getCoalescedEvents</dfn></dt>
+              <dd>
+                  <p> Returns a sequence of all <code>PointerEvents</code> that were coalesced into the received event. The following attributes of the coalesced events will always have the same value as the received event: <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>, and particularly <code>target</code>. But other attributes might be different. The events in the sequence will have increasing <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a> ([[!WHATWG-DOM]]). So the first event will have the smallest <code>timeStamp</code>. Also the received event's <code>timeStamp</code> will be greater than or equal to the last event's <code>timeStamp</code> in the sequence. None of the events in the sequence will have coalesced events. When an event has no coalesed event, this function returns an empty sequence.</p>
```
mustaqahmed: @RByers, @dtapuska: With coalescing, the time-stamps of the received events may not reflect the times when the events occurred. Is it okay? Consider a single pointermove that occurred at time T, which the browser held back until the next frame timing at T+t1. Assuming for simplicity that there is no other held-back events, is it really okay (or possible) that the single pointermove event fired at time T+t1 would have a timestamp T?

mustaqahmed (commented):
Have a question & a suggestion on event timing.

extension.html
NavidZ: I guess I missed the same thing in auxclick spec as well then as I just copy pasted from that :). I fixed it here.

extension.html
NavidZ: done

extension.html
NavidZ: done

extension.html
NavidZ: done

extension.html
NavidZ: Makes sense.

extension.html
NavidZ: Is this good enough?

extension.html
NavidZ: done

extension.html
NavidZ: I added a link to those APIs in the introduction. I couldn't find any link for Windows one though.

extension.html
NavidZ: I added a new section about it. Let me know what you think.

NavidZ (commented):
I also updated the fact that all pointermove events will have at least one coalesced event and the other types will have no coalesced event.
The other solution for pointermove events was that the getCoalescedEvents size is either zero or >=2 which seemed weird to me.

patrickhlauke:
LGTM in principle. @rbyers we good to merge?

extension.html
```diff
@@ -0,0 +1,148 @@
...
... always have the same value as the received event: <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>, and particularly <code>target</code>. But other attributes might be different. The events in the sequence will have increasing <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a> ([[!WHATWG-DOM]]). So the first event will have the smallest <code>timeStamp</code>. The received event's attributes will be initalized in a way that is best representative of all the coalesced events. For example It's <code>timeStamp</code> will be equal to the last event's <code>timeStamp</code> in the sequence and it's <a href="https://www.w3.org/TR/pointerlock/#widl-MouseEvent-movementX">movementX</a> and <a href="https://www.w3.org/TR/pointerlock/#widl-MouseEvent-movementY">movementY</a> ([[!POINTERLOCK]]) will be the sum of that of all the coalesced events. None of the events in the sequence will have coalesced events and <code>getCoalescedEvents</code> returns an empty sequence for them. This API always returns at least one coalesced event for <code>pointermove</code> events and an empty list for other types of <code>PointerEvents</code>.</p>
```
mustaqahmed: Please clarify the nesting case: "None of the events in the sequence will have (nested) coalesced events, so getCoalescedEvents returns...".

Another nit: "For example _its_ timestamp ... and _its_...".

extension.html
```diff
@@ -0,0 +1,148 @@
...
+              </dd>
+          </dl>
+      </div>
+    </section>
+
+    <section>
+      <h2>Timing of coalesced event firing</h2>
+      <p>Coalescing events only happens for <code>pointermove</code> events. All the coalesced events should be fired before the next frame or before any <code>pointerup</code>, <code>pointerdown</code>, <code>pointercancel</code> events. Note that user agents don't need to hit-test all the coalesced events as all of them will have the same target as the dispatched event.</p>
```
mustaqahmed, line 84: Nit to make consistent with last section: "the same target as the _received_ event".

Alternatively if you prefer, replace all "received" with "dispatched". I have no strong opinion.

extension.html
```diff
@@ -0,0 +1,148 @@
...
+      <p>When the user agent dispatches any <code>PointerEvent</code> it MUST dispatch an event for every <code>pointerId</code> that has any coalesced events so far at that point with the correct relative ordering.</p>
```
mustaqahmed: Here "has any coalesced events" may seem to include only pointermoves. Also "an event for every pointerId" is not a perfect description because your example below (which is great btw) has two events with id=1. What about this: "...it MUST dispatch all events held back so far for coalescing, regardless of their pointerIds. The ordering of all these dispatched events ...

mustaqahmed (changes requested):

extension.html
NavidZ: I reworded this. Let me know if this looks okay.

extension.html
```diff
@@ -0,0 +1,148 @@
...
+              </dd>
+          </dl>
+      </div>
+    </section>
+
+    <section>
+      <h2>Timing of coalesced event firing</h2>
+      <p>Coalescing events only happens for <code>pointermove</code> events. All the coalesced events should be fired before the next frame or before any <code>pointerup</code>, <code>pointerdown</code>, <code>pointercancel</code> events. Note that user agents don't need to hit-test all the coalesced events as all of them will have the same target as the dispatched event.</p>
+      <p>When the user agent dispatches any <code>PointerEvent</code> it MUST dispatches all the events held back so far for coalescing, regardless of their <code>pointerIds</code>. The ordering of all these dispatched events should be in a way that resemebles the most with the actual events' ordering. For example if a <code>pointerdown</code> event causes the dispatch for the coalesced <code>pointermove<code> events the user agent SHOULD first dispatch one <code>pointermove</code> event with all those coalesced events of a <code>pointerId</code> followed by the <code>pointerdown</code> event.</p>
```
RByers: typo: s/dispatches/dispatch/

RByers: typo: `<code>pointermove<code>` -> `<code>pointermove</code>`

extension.html
RByers: Looks good.  Maybe worth adding / updating a timestamp for completeness?  Then again, that often won't matter so maybe it's not worth complicating the example.  Up to you.

extension.html
RByers: Yes I think that's what we want.  the timestamp represents (roughly) when the finger was at the supplied position, nothing about when the event was actually dispatched.

extension.html
NavidZ: done

NavidZ: done

extension.html
NavidZ: I used the dispatched everywhere.

extension.html
RByers: This makes sense, and is an extra developer complexity I didn't appreciate.  Android and iOS don't have this problem since they bundle the pointers together.  At least I think it would be very rare that an app would want to BOTH reason about the relationship between different fingers AND look at historical events (maybe a drawing app that uses the distance between fingers to influence the stroke width or something).

RByers (changes requested):
Just a couple tiny tweeks

RByers (approved):

RByers:
LGTM.  @mustaqahmed let us know if you have any other feedback

extension.html
```diff
@@ -0,0 +1,148 @@
...
+      <h2>Timing of coalesced event firing</h2>
+      <p>Coalescing events only happens for <code>pointermove</code> events. All the coalesced events should be fired before the next frame or before any <code>pointerup</code>, <code>pointerdown</code>, <code>pointercancel</code> events. Note that user agents don't need to hit-test all the coalesced events as all of them will have the same target as the dispatched event.</p>
+      <p>When the user agent dispatches any <code>PointerEvent</code> it MUST dispatch all the events held back so far for coalescing, regardless of their <code>pointerIds</code>. The ordering of all these dispatched events should be in a way that resemebles the most with the actual events' ordering. For example if a <code>pointerdown</code> event causes the dispatch for the coalesced <code>pointermove</code> events the user agent SHOULD first dispatch one <code>pointermove</code> event with all those coalesced events of a <code>pointerId</code> followed by the <code>pointerdown</code> event.</p>
+      <p>Here is an example of the actual events happening in one frame with increasing <code>timestamps</code> and the dispatched events by the user agent:
```
mustaqahmed: Nit: s/dispatched events by/events dispatched by/.

extension.html
```diff
@@ -0,0 +1,148 @@
...
+            </td></tr>
+          </tbody>
+        </table>
+      </p>
+    </section>
+
+    <section class="informative">
+      <h2>Examples</h2>
+      <p>The following are examples that demonstrate how the APIs in this specification might
+      be used</p>
```
mustaqahmed: Nit: The line needs an ending colon or period.

extension.html
mustaqahmed: Thanks for switching to "dispatch", much clearer now.

extension.html
```diff
@@ -0,0 +1,148 @@
...
...e the same value as the dispatched event: <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>, and particularly <code>target</code>. But other attributes might be different. The events in the sequence will have increasing <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a> ([[!WHATWG-DOM]]). So the first event will have the smallest <code>timeStamp</code>. The dispatched event's attributes will be initalized in a way that is best representative of all the coalesced events. For example its <code>timeStamp</code> will be equal to the last event's <code>timeStamp</code> in the sequence and its <a href="https://www.w3.org/TR/pointerlock/#widl-MouseEvent-movementX">movementX</a> and <a href="https://www.w3.org/TR/pointerlock/#widl-MouseEvent-movementY">movementY</a> ([[!POINTERLOCK]]) will be the sum of that of all the coalesced events. None of the events in the sequence will have (nested) coalesced events, so <code>getCoalescedEvents</code> returns an empty sequence for them. This API always returns at least one coalesced event for <code>pointermove</code> events and an empty list for other types of <code>PointerEvents</code>.</p>
```
mustaqahmed: s/sum of that of/sum of those of/.

mustaqahmed (approved):
Sorry for missing the review over the week. This looks great, has all ordering details I wanted to see in a well-organised manner.

LGTM with a few optional nits.

NavidZ:
Thanks for catching all the mistakes. They should be all fixed now.

PR #153, opened 2016-11-08
Call out pressure=0 for pointerup events

### LINKED ISSUE #146: pointerup pressure should always be 0
`pointerup` is defined to be when a {finger,stylus} loses contact with the screen or when mouse buttons are no longer being pushed.  Therefore, I'd expect the pressure value of every `pointerup` event to be 0.  Moreover, the spec requires pressure to be 0 if the hardware doesn't support it and no buttons are active (that is, the pointer is up).

Maddeningly, its actual value is different across implementations.  Here's the current breakdown:

| Platform | Browser | Pointer | pointerup pressure is 0 | pointerup pressure is whatever the last pointermove recorded |
| --- | --- | --- | --- | --- |
| Android | Chrome | touch |  | :ballot_box_with_check: |
| iOS | Safari via PEP | touch |  | :ballot_box_with_check: |
| Mac | Chrome | trackpad | :ballot_box_with_check: |  |
| Mac | Chrome | Wacom stylus | :ballot_box_with_check: |  |
| Mac | Chrome | Wacom touch |  | :ballot_box_with_check: (always 1, spec says it should be .5) |
| Mac | Firefox | trackpad | :ballot_box_with_check: |  |
| Mac | Firefox | Wacom stylus | :ballot_box_with_check: |  |
| Mac | Firefox | Wacom touch | :ballot_box_with_check: |  |
| Mac | Safari via PEP | trackpad | :ballot_box_with_check: |  |
| Mac | Safari via PEP | Wacom stylus | :ballot_box_with_check: |  |
| Mac | Safari via PEP | Wacom touch | :ballot_box_with_check: |  |
| ChromeOS | Chrome | touch |  | :ballot_box_with_check: |
| ChromeOS | Chrome | trackpad | :ballot_box_with_check: |  |
| Windows | Chrome or Edge | touch |  | :ballot_box_with_check: |
| Windows | Chrome or Edge | trackpad | :ballot_box_with_check: |  |
| Windows | Chrome or Edge | stylus | :ballot_box_with_check: |  |
| Windows | Firefox | touch | :ballot_box_with_check: |  |
| Windows | Firefox | trackpad | :ballot_box_with_check: |  |
| Windows | Firefox | stylus | :ballot_box_with_check: |  |

And here's [the test](https://codepen.io/appsforartists/full/gwLWVE/).  If the background turns green on `pointerup`, `pressure` is 0.  If it turns red, `pressure` is using a stale value from `pointermove`.

**_tl;dr:_**
- Firefox always sets pressure to 0 in `pointerup`.
- The other desktop browsers use 0 for stylus and trackpad, but the last `pointermove` value for touch.
- iOS and Android both use the last pressure value from `pointermove`.

**Can we please explicitly declare in the spec that pressure should be 0 in every `pointerup`?**  This would be consistent with the currently specified behavior for pointers that don't granularly report pressure.

Here are the relevant bugs:
- [Chrome](https://crbug.com/648777)
- [PEP](https://github.com/jquery/PEP/issues/180/)

NavidZ:
closes #146

### DIFF
--- a/index.html
+++ b/index.html
@@ -317,7 +317,7 @@ interface PointerEvent : MouseEvent {
                         </dd>
                     <dt><dfn>pressure</dfn></dt>
                         <dd>
-                            <p>The normalized pressure of the pointer input in the range of [0,1], where 0 and 1 represent the minimum and maximum pressure the hardware is capable of detecting, respectively. For hardware that does not support pressure, the value MUST be 0.5 when in the <a>active buttons state</a> and 0 otherwise. Particularly as a result, all <code>pointerup</code> events will have pressure 0.</p>
+                            <p>The normalized pressure of the pointer input in the range of [0,1], where 0 and 1 represent the minimum and maximum pressure the hardware is capable of detecting, respectively. For hardware that does not support pressure, the value MUST be 0.5 when in the <a>active buttons state</a> and 0 otherwise. Note: all <code>pointerup</code> events will have pressure 0.</p>
                         </dd>
                     <dt><dfn>tangentialPressure</dfn></dt>
                         <dd>
--- a/index.html
+++ b/index.html
@@ -317,7 +317,7 @@ interface PointerEvent : MouseEvent {
                         </dd>
                     <dt><dfn>pressure</dfn></dt>
                         <dd>
-                            <p>The normalized pressure of the pointer input in the range of [0,1], where 0 and 1 represent the minimum and maximum pressure the hardware is capable of detecting, respectively. For hardware that does not support pressure, the value MUST be 0.5 when in the <a>active buttons state</a> and 0 otherwise.</p>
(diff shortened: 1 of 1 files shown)


scottgonzalez:
This looks fine to me.

index.html
```diff
@@ -317,7 +317,7 @@
                         </dd>
                     <dt><dfn>pressure</dfn></dt>
                         <dd>
-                            <p>The normalized pressure of the pointer input in the range of [0,1], where 0 and 1 represent the minimum and maximum pressure the hardware is capable of detecting, respectively. For hardware that does not support pressure, the value MUST be 0.5 when in the <a>active buttons state</a> and 0 otherwise.</p>
+                            <p>The normalized pressure of the pointer input in the range of [0,1], where 0 and 1 represent the minimum and maximum pressure the hardware is capable of detecting, respectively. For hardware that does not support pressure, the value MUST be 0.5 when in the <a>active buttons state</a> and 0 otherwise. Particularly as a result, all <code>pointerup</code> events will have pressure 0.</p>
```
patrickhlauke: I think we can get away with dropping the "Particularly as a result" (which reads a bit funny) and instead change this to, say, "Note: all <code>pointerup</code> ..."

NavidZ: done

patrickhlauke:
LGTM, merged

PR #162, opened 2016-12-20
Reference compat spec touch-action entry

RByers:
Add a hack to work around the long-standing problem in w3c/pointerevents#565.

@staktrace @smaug---- @dtapuska WDYT?


dtapuska:
lgtm

smaug----:
Looks reasonable, for now at least.

RByers:
Thanks!

PR #179, opened 2017-02-22
Define coalesced events target, currentTarget, and eventPhase

NavidZ:
@dtapuska @RByers @smaug----


dtapuska:
There is a typo "the" should be them. 

As my feedback delivered verbally before this was written was whether we should include text that indicates that calling any methods shouldn't have an effect on the current dispatch. For example if I call event.getCoalescedEvent()[0].stopPropogation() shouldn't have an effect on the current dispatch (unless  of course event.getCoalescedEvent()[0] === event i guess)

NavidZ:
I don't think the current spec allows any coalesced event ever be the same as the actual event as they have different values for some of the attributes like cancelable.
So I added those functions in the pool as well. Is that what you meant?

NavidZ:
@dtapuska does this look good now?

dtapuska:
lgtm

PR #182, opened 2017-03-01
Expand list of aknowledgements

patrickhlauke:
Also rewording, so not just "group members"


patrickhlauke:
@RByers does this look ok? did I miss anybody out? (sorry, doing some boring housekeeping on the spec)

RByers:
LGTM, thanks!  I don't see anyone obviously missing from the Google side.

PR #186, opened 2017-03-07
Clear the definition of boudnary events in capturing scenarios

### LINKED ISSUE #185: The spec is unclear whether boundary events should fire while pointer is captured to a certain target
setPointerCapture says: "For subsequent events of the pointer, the capturing target will substitute the normal hit testing result as if the pointer is always over the capturing target"

pointerleave (as an example): "A user agent MUST fire a pointer event named pointerleave when a pointing device is moved out of the hit test boundaries of an element and all of its descendants, including as a result of a pointerup and pointercancel events from a device that does not support hover (see pointerup and pointercancel). Note that setPointerCapture or releasePointerCapture might have changed the hit test target."

So, if pointer is moved outside the capturing target, per spec pointerleave should fire, since, well pointing device is outside hit test boundaries (note, moving outside the boundaries doesn't necessarily mean hit testing). But then setPointerCapture says hit testing doesn't happen and I guess it tries to say pointerleave shouldn't be dispatched.

Could this be clarified?

NavidZ:
@smaug---- Does this address the confusion?

Closes #185


index.html
```diff
@@ -537,11 +537,11 @@
...
+                <p>A user agent MUST <a href="#firing-events-using-the-pointerevent-interface">fire a pointer event</a> named <code>pointerover</code> when a pointing device is moved into the hit test boundaries of an element. Note that <code>setPointerCapture</code> or <code>releasePointerCapture</code> might have changed the hit test target and while a pointer is captured it is considerd to be always inside the boundaries of the capturing element for the purpose of firing boundary events. A user agent MUST also fire this event prior to firing a <code>pointerdown</code> event for <a href=#mapping-for-devices-that-do-not-support-hover>devices that do not support hover</a> (see <code><a href="#the-pointerdown-event">pointerdown</a></code>).</p>
```
scottgonzalez: typo: considerd

Appears in all four sentences.

smaug----:
looks reasonable

smaug---- (approved):

PR #188, opened 2017-03-09
Add notes about touch-action and direct/indirect manipulation and stylus

patrickhlauke:
It's too late of course to change "touch-action" and to generalise the
language in this whole section, but...with these notes, certain
scenarios such as Samsung Galaxy Note (where you CAN pan with a stylus,
and where touch-action DOES have the correct effect the same way as for
touch) are more explicitly covered.

Conversely, on a Surface 3, you CAN'T pan with the Surface Pen - so
touch-action does NOT have any effect here (since the browser doesn't
have any default behavior here that needs to be dealt with/suppressed)

### DIFF
--- a/index.html
+++ b/index.html
@@ -967,18 +967,23 @@ partial interface Navigator {
     </section>
     <section class="appendix">
         <h1>Acknowledgments</h1>
-        <p>Many thanks to lots of people for their proposals and recommendations, some of which are incorporated into this document. The group's Chair acknowledges contributions from the following group members:
+        <p>Many thanks to lots of people for their proposals and recommendations, some of which are incorporated into this document. The group's Chair acknowledges contributions from the following past and present group members and participants:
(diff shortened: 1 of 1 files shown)


patrickhlauke:
(apologies for the multiple commits...at some point i must have messed up my sync with the upstream, and am too much of a git noob to work out how to undo that mess...will squash for the merge)

patrickhlauke:
adding for clarification: i've left the language (and mention of "direct or indirect") fairly open-ended, as i can envisage user agents also doing something for mouse (some UA deciding at some point that the act of clicking-then-dragging the mouse should be used to scroll), and then `touch-action` should also be followed.

RByers:
Note we had some discussion about this PR [on Twitter](https://twitter.com/patrick_h_lauke/status/841028283533799425) when I didn't have my key handy to auth to GitHub.

PR #189, opened 2017-03-09
Set isTrusted for coalesced events

### LINKED ISSUE #187: Set coalesced events isTrusted flag to false
@dtapuska @smaug---- @RByers 
I was wondering if it is alright to leave the isTrusted flag to be false instead of setting it from the original event. I believe since these coalesced events are not the events to be fired or maybe handled by themselves having the isTrusted bit doesn't convey much of a meaning.

NavidZ:
closes #187 
@dtapuska @RByers @smaug---- looks reasonable?


smaug---- (commented):
Could we please make pointer events spec source code
easier to review, so not super long lines. TabAtkins said there is some tool for that.

UA creating non-trusted events feels wrong.

NavidZ:
How about this? I basically specifically mentioned the guarantee of those attributes including isTrusted being the same as the parent event holds only when the event is created by the browser.

NavidZ:
@TabAtkins is there a tool so that makes the spec text more structured and easier to review?

tabatkins:
There's no "tool", just a useful technique I and @fantasai use, and have been spreading to others, called [semantic linefeeds](http://rhodesmill.org/brandon/2012/one-sentence-per-line/).  Using this, you get most of the benefits of an "80 col" or "100 col" line-wrapping limit, but without it causing extra re-wrapping when you make an edit early in a paragraph.

The end result is that the line-wrapping is easy to apply while you're writing (you quickly develop a reflex for pressing Enter when you start a new phrase), lines are short and easy to view in PRs, and diffs are localized to just the bits of the sentences that actually changed, so it's easy to find what was altered.

NavidZ:
@smaug---- do you want me to split up this paragraph in this review? or does it look good as it is?
I don't know if there is a tool that does it for the whole document. If so it might make more sense to do only that it in another review.

smaug----:
The change looks reasonable.

Better to change "coding style" of the spec in some other pr.

RByers:
Reformatting to a "semantic linefeeds" style (in it's own PR) SGTM (I agree the git diff would be MUCH easier to work with).  This no-wrapping style dates back to Jacob's original commit - IIRC, I said I could live with it if it was what he preferred :-).

PR #190, opened 2017-03-09
Clarify the relatedTarget definition for pointerevents

### LINKED ISSUE #184: The spec doesn't define the values for relatedTarget
Per current spec relatedTarget should be null (since nothing sets it), but implementations sure are setting it to some value with pointerover/out etc.

NavidZ:
closes #184


NavidZ:
@smaug---- does this look good?

smaug---- (approved):
Given that hit testing and all that is undefined in general,  this should be good enough for now.

index.html
```diff
@@ -514,6 +514,7 @@
                         </table>
 
                         <p>For all pointer events in the table above, <code>composed</code> ([[!WHATWG-DOM]]) attribute SHOULD be <code>true</code> and <a href="https://www.w3.org/TR/uievents/#widl-UIEvent-detail"><code>detail</code></a> [[!DOM-LEVEL-3-EVENTS]] attribute SHOULD be 0.</p>
+                        <p>Similar to the <a href="https://www.w3.org/TR/uievents/#interface-mouseevent">MouseEvents</a> [[!DOM-LEVEL-3-EVENTS]] the <code>relatedTarget</code> should be initialized to the element whose bounds the pointer just left (in the case of a <code>pointerover</code> or <code>pointerenter</code> event) or the element whose bounds the pointer is entering (in the case of a <code>pointerout</code> or <code>pointerleave</code>). For other pointer events, this value need not be assigned (and will default to null). Note that when an element receives the pointer capture all the following events for that pointer are considered to be inside the boundary of the capturing element.</p>
```
mustaqahmed: Let's be more precise for "other pointer events" and say that "this value will be null" (to rule out any implementation not going for the default value).

mustaqahmed (approved):
Thanks, LGTM.

RByers:
Fixes #184

Issue #183, opened 2017-03-01
ReSpec warning on "Privacy and Security"


patrickhlauke:
New ReSpec warning "This specification doesn't appear to have any 'Privacy' or 'Security' considerations sections. Please consider adding one, see https://w3ctag.github.io/security-questionnaire/"

marcoscaceres:
> On 4 Apr 2017, at 7:04 pm, Patrick H. Lauke <notifications@github.com> wrote:
> 
> /cc @marcoscaceres any good example out in the wild of a spec that has such a section?
> 

I cant think of any really good ones ... which is sad 😭 

> i'm guessing (please anybody else chime in if you think i'm talking rubbish) that this spec does not have any such potential concerns, so am i right in thinking we'd just need a section with the correct heading text and then a short para basically saying that there's nothing to be concerned about here in this respect?
> 

Go through: 

https://www.w3.org/TR/security-privacy-questionnaire/

It's actually super enlightening. You should be able to construct a paragraph from answering those questions.

patrickhlauke:
@marcoscaceres well i gave it a shot https://github.com/w3c/pointerevents/pull/193 ;)

PR #193, opened 2017-04-04
Add security/privacy considerations

patrickhlauke:
Closes https://github.com/w3c/pointerevents/issues/183

### DIFF
--- a/index.html
+++ b/index.html
@@ -988,6 +988,10 @@ partial interface Navigator {
         </p>
         <p>Special thanks to those that helped pioneer the first edition of this model, including especially: Charu Chandiram, Peter Freiling, Nathan Furtwangler, Thomas Olsen, Matt Rakow, Ramu Ramanathan, Justin Rogers, Jacob Rossi, Reed Townsend and Steve Wright.</p>
     </section>
+    <section class="appendix">
+        <h1>Security and Privacy Considerations</h1>
+        <p>There are no known security or privacy impacts resulting from the features outlined in this specification.</p>
+    </section>
     <section class="appendix informative">
         <h1>Revision History</h1>
         <p>The following is an informative summary of substantial and major editorial changes between publications of this specification, relative to the first [[PointerEvents]] specification. See the <a href="https://github.com/w3c/pointerevents/commits">complete revision history of the Editor's Drafts of this specification</a>.</p>


patrickhlauke:
Just adding reviewers to make sure I've not missed anything about potential privacy/security issues. To my naive developer eyes there are none, but I'd rather get it confirmed before merging this and then having to revert...

index.html
```diff
@@ -988,6 +988,10 @@
         </p>
         <p>Special thanks to those that helped pioneer the first edition of this model, including especially: Charu Chandiram, Peter Freiling, Nathan Furtwangler, Thomas Olsen, Matt Rakow, Ramu Ramanathan, Justin Rogers, Jacob Rossi, Reed Townsend and Steve Wright.</p>
     </section>
+    <section class="appendix">
+        <h1>Security and Privacy Considerations</h1>
+        <p>There are no known security or privacy impacts resulting from the features outlined in this specification.</p>
```
marcoscaceres, line 993: I was expecting:

The working group has undertaken the Privacy and Security questionnaire [[!CITATION HERE]]. From the review, the working group believes that this API: 

 * Does not  expose personally-identifiable information.
 * Does not deal with high-value data.
 * Does not introduce new state for an origin that persists across browsing sessions.
 * Does not expose persistent, cross-origin state to the web. 
 * Does not expose any other data to an origin that it doesn’t currently have access to.
 * Does not enable new script execution/loading mechanisms.
 * Does not allow an origin access to a user’s location.
 * Does not require any special handling when the user agent is in "incognito" mode.
* Does not allow an origin access to other devices.
* Does not allow an origin control over a user agent’s native UI.
* Does not expose temporary identifiers to the web.
* Does not distinguish between behavior in first-party and third-party contexts.
* Does not persist data to a user’s local device.
* Does not allow downgrading default security characteristics.

However, ANSWER THESE please: 

* Does not allow an origin access to sensors on a user’s device?
* Does not allow an origin access to aspects of a user’s local computing environment?

marcoscaceres (changes requested):

index.html
marcoscaceres: You may choose to flip the order. And you should probably consider how pointer events can be used to track what the user does on a website (is it different from mouse event? etc.).

marcoscaceres: I'll note that: 
> Does not expose personally-identifiable information. 

It might be possible to identify a user by the way they use an input device - as they may generate a particular usage signature over time.

patrickhlauke: I'll note again that without any good examples out there, it's difficult for a layperson like me to even understand what this security/privacy section should look like. If the expectation is to have something along the lines you mentioned there, with detailed bullet points, then it would be ace to see that as an actual example - it may be clear once reading the whole of https://www.w3.org/TR/security-privacy-questionnaire/, but on skim-reading it it wasn't apparent ;)

annevk: I don't think a list of bullet points is required, but certainly there's some amount of fingerprinting possible by exposing touch events.

tomrittervg:
This draft does have privacy considerations. It exposes pointerType, which reveals whether the user is using a mouse, pen, or touch input. It also exposes detailed information about the geometry of the input (size of the thing pointing, pressure, tilt, twist). All of these are more detailed information than websites currently receiving, meaning that this can be used as a mechanism for fingerprinting (and tracking) users.

As mentioned in https://www.w3.org/2001/tag/doc/unsanctioned-tracking/

> [the TAG b]elieves that, because combatting fingerprinting is difficult, new Web specifications should take reasonable measures to avoid adding unneeded fingerprinting surface area. However, added surface area should not be a primary factor in determining whether to add a new feature.
> Asserts that when a new feature does add fingerprinting surface area, it should be documented as such.

Therefore I strongly believe that these concerns should be documented in the spec.

patrickhlauke:
superceded by https://github.com/w3c/pointerevents/pull/235

PR #196, opened 2017-04-12
Clarify pointer capture behavior when pointer lock is applied

### LINKED ISSUE #176: How should PointerLock vs setPointerCapture work together?
This issue covers the PointerLock part of #135. Here is the original comment in that bug (by @mingchou):

> One element setPointerCapture and the other requestPointerLock. Please ignore this case if it will be covered by issue#131. On Canary, requestPointerLock works but setPointerCapture not. On Edge, the result is opposite.

In Chrome, pointer-capture comes after pointer-lock in the event flow just because we chose not to worry about both together in our initial implementation. We need to fix it for sure.

As for Edge, @NavidZ observed that it doesn't allow a requestPointerLock() after a setPointerCapture() but a setPointerCapture() after a requestPointerLock() doesn't work reasonably: PointerLock overrides the effect of capturing while the capturing target still gets got/lostpointercapture events.

What Edge does in the first case is the most reasonable solution to me: don't allow both of these together. The PointerLock spec [seems to suggest](https://w3c.github.io/pointerlock/#why-not-merge-with-mouse-capture-setcapture) these two can be active together (for the similar but non-standard setCapture() API instead), but I don't think the text there gives a definitive answer.

LanWei22:
closes #176


index.html
```diff
@@ -828,6 +828,7 @@
             <ol>
                 <li>If the <code>pointerId</code> provided as the method's argument does not match any of the <a data-lt="active pointer">active pointers</a>, then throw a <code>DOMException</code> with the name <code>InvalidPointerId</code>.</li>
                 <li>If the <code>Element</code> on which this method is invoked does not participate in its <code>ownerDocument</code>'s tree, throw an exception with the name <code>InvalidStateError</code>.</li>
+                <li>If this method is invoked while the page has a locked element, throw an exception with the name <code>InvalidStateError</code>.</li>
```
NavidZ: Can we link the pointerlock spec here like this "[[!PointerLock]]" ?

index.html
```diff
@@ -854,6 +855,7 @@
...
             <p>If the user agent supports firing the <code>click</code> event (see <a title="compatibility mouse events" href="#dfn-compatibility-mouse-events">compatibility mouse events</a>) and if in an implicit release scenario both <code>click</code> and <code>lostpointercapture</code> events are fired, <code>click</code> should be fired before <code>lostpointercapture</code>.</p>
             <p>When the <a>pointer capture target override</a> is removed from its <code>ownerDocument</code>'s tree, clear the <a>pending pointer capture target override</a> and <a>pointer capture target override</a> nodes and fire a PointerEvent named <code>lostpointercapture</code> at the document.</p>
+            <p>When a pointer lock is successfully applied on an element, a user agent MUST run the steps as if the releasePointerCapture() method has been called if any element is set to be captured or pending to be captured.
```
NavidZ, line 867: This line is missing running the "Process Pending Pointer Capture" steps. I believe the plan is to send the `lostpointercapture` right before `pointerlockchange`. Right? Then we need to maybe mention that here as well.

NavidZ (changes requested):

NavidZ:
@smaug---- @mingchou @patrickkettner does the change and the proposed solution sound reasonable to you guys as well?

patrickkettner:
lgtm

index.html
```diff
@@ -855,7 +855,7 @@
...
             <p>When the <a>pointer capture target override</a> is removed from its <code>ownerDocument</code>'s tree, clear the <a>pending pointer capture target override</a> and <a>pointer capture target override</a> nodes and fire a PointerEvent named <code>lostpointercapture</code> at the document.</p>
-            <p>When a pointer lock is successfully applied on an element, a user agent MUST run the steps as if the releasePointerCapture() method has been called if any element is set to be captured or pending to be captured.
+            <p>When a pointer lock is successfully applied on an element, a user agent MUST run the steps as if the releasePointerCapture() method has been called if any element is set to be captured or pending to be captured and run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code>.
```
NavidZ: We already talked offline but to mention for others. I was wrong here. Since we do not have any pointer event object at the time of lock change for `Process Pending Pointer Capture` to use then I guess we should delay calling that process until the next pointer event we fire for mouse which is already covered in another part of this spec. So I guess having this paragraph the way it was is correct.

NavidZ:
Thanks. LGTM

NavidZ (approved):

PR #200, opened 2017-04-26
Clarified the touch-action values for transformed elements.

### LINKED ISSUE #115: Clarify what touch-action allows with rotated scrollable elements
I think the spec is unclear on what the intended behaviour is for elements which have rotation transforms on them. For example on [this test page](https://people.mozilla.org/~kgupta/examples/w3c-pe-114.html) Chrome allows scrolling while your finger is moving in the "x" direction, even though that means the element is scrolling vertically (the scrollTop value is changing). While that's a reasonable thing to do, our implementation in Firefox does the opposite (allows scrolling such that scrollLeft changes).

mustaqahmed:
Also rearranged the touch-action section for better readability.

(This is an updated version of my last (reverted) pull request.)

Related: #115.

### DIFF
--- a/index.html
+++ b/index.html
@@ -734,43 +734,53 @@ partial interface Navigator {
                 <tr><th>Computed value:</th><td>Same as specified value.</td></tr>
             </table>
 
-            <p>The <code>touch-action</code> CSS property determines whether touch input MAY  trigger default behavior supplied by user agent.  This includes, but is not limited to, behaviors such as panning or zooming.</p>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
Changed "Closes" to "Related" in your message above, since you wanted to keep that one open (saves you re-reopening it

RByers (commented):
LGTM also, thanks!

Left some additional thoughts in #115.

Issue #194, opened 2017-04-04
Should DnD release pointer capture?


NavidZ:
Following #135 I believe there was an agreement that when DnD starts it should release the pointer capture of the pointer that initiated the DnD. I faced another question while implementing this in Chrome.

**Should the lostpointercapture be sent after dragstart event or right before?**

Testing [this example](http://jsbin.com/giqopi/edit?html,output) on Edge it seems that if fires the lostpointercapture sometime after dragstart. Edge may also send a few drag events before firing lostpointercapture depending on how much you move the mouse.

I believe it might be a little inconsistent with what the [DnD](http://w3c.github.io/html/editing.html#drag-and-drop) spec says:
_From the moment that the user agent is to initiate the drag-and-drop operation, until the end of the drag-and-drop operation, device input events (e.g., mouse and keyboard events) must be suppressed._

However, the spec didn't define what exactly drag-and-drop operation is. For example if js calls preventDefault in dragstart event handler the drag doesn't continue (or not even start I guess).

**Having said all these, and to also make it consistent across the board, I suggest sending lostpointercapture right before dragstart event if no one has any point against it.**

NavidZ:
@patrickkettner is this alright from your side in Edge?
@mingchou does this sound reasonable since you are implementing it in Firefox?

NavidZ:
@patrickkettner does the proposal sound reasonable?

patrickkettner:
lgtm

NavidZ:
Note that following the discussion on another issue #205 we went with sending pointercancel (and if necessary the lostpointercapture) after the drag operation starts which could be prevented by calling preventDefault on dragstart.
The test [here](https://github.com/w3c/web-platform-tests/blob/master/pointerevents/html/pointerevent_drag_interaction-manual.html) expects this behavior.

PR #201, opened 2017-04-27
Define the interation of drag operation vs capture

### LINKED ISSUE #205: Add OS image dragging to non-normative reasons for pointercancel
The nonnormative reasons for `pointercancel` include a bunch of scenarios where a browser default action (like panning) took over in the middle of a gesture.  A common, but unmentioned, scenario is browser/OS-level drag-and-drop.  Like when panning, I believe this should cause a `pointercancel` to be dispatched when the browser stops sending PointerEvents.  Unfortunately, this doesn't seem to currently be specified; and none of {Chrome, Edge, Firefox} behave that way.

Here's a bug to fix in Chrome:
https://crbug.com/720201

and a reduction:
https://codepen.io/appsforartists/pen/NjybQX?editors=1011

NavidZ:
Release the pointer capture right before
the drag operation starts. In other words,
if a capturing pointer caused the drag,
before the dragstart event is fired to the page
lostpointercapture event should be sent for
that pointer.
fixes #205  and #194

### DIFF
--- a/index.html
+++ b/index.html
@@ -861,10 +861,12 @@ partial interface Navigator {
         <section>
             <h3>Implicit Release of Pointer Capture</h3>
             <p>Immediately after firing the <code>pointerup</code> or <code>pointercancel</code> events, a user agent MUST clear the <a>pending pointer capture target override</a> for the <code>pointerId</code> of the <code>pointerup</code> or <code>pointercancel</code> event that was just dispatched, and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code> if necessary.</p>
-            <p>Immediately before firing <code>dragstart</code> event [[!HTML]],
-a user agent MUST clear the <a>pending pointer capture target override</a> for the pointer that caused the drag operation,
-and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code>
-before the <code>dragstart</code> event if necessary.</p>
+            <p>Immediately before <a href="https://html.spec.whatwg.org/multipage/dnd.html#initiate-the-drag-and-drop-operation">drag operation starts</a> [[!HTML]],
+a user agent MUST send a <code>pointercancel</code> for the pointer that caused the drag operation.
+This <code>pointercancel</code> will follow the paragraph above regarding releasing the capture.</p>
+<div class="note">If the drag operation initiation is prevented through any means
+(e.g. calling <code<preventDefault</code> the <code>dragstart</code> event)
+there will be no need for the <code>pointercancel</code>.</div>
             <p>If the user agent supports firing the <code>click</code> event (see <a title="compatibility mouse events" href="#dfn-compatibility-mouse-events">compatibility mouse events</a>) and if in an implicit release scenario both <code>click</code> and <code>lostpointercapture</code> events are fired, <code>click</code> should be fired before <code>lostpointercapture</code>.</p>
             <p>When the <a>pointer capture target override</a> is removed from its <code>ownerDocument</code>'s tree, clear the <a>pending pointer capture target override</a> and <a>pointer capture target override</a> nodes and fire a PointerEvent named <code>lostpointercapture</code> at the document.</p>
         </section>
--- a/index.html
+++ b/index.html
@@ -567,11 +567,11 @@ interface PointerEvent : MouseEvent {
                     <li>The user agent has determined that a pointer is unlikely to continue to produce events (for example, because of a hardware event).</li>
                     <li>After having fired the <code>pointerdown</code> event, if the pointer is subsequently used to manipulate the page viewport (e.g. panning or zooming).</li>
                     <li>Immediately before <a href="https://html.spec.whatwg.org/multipage/dnd.html#initiate-the-drag-and-drop-operation">drag operation starts</a> [[!HTML]],
-                     for the pointer that caused the drag operation.</li>
+                     for the pointer that caused the drag operation.
+                     <div class="note">If the start of the drag operation is prevented through any means
+                     (e.g. through calling <code>preventDefault</code> on the <code>dragstart</code> event)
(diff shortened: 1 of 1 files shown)


NavidZ:
Any comments on this? Does this look good? @patrickkettner @smaug----

index.html
```diff
@@ -861,6 +861,10 @@
         <section>
             <h3>Implicit Release of Pointer Capture</h3>
             <p>Immediately after firing the <code>pointerup</code> or <code>pointercancel</code> events, a user agent MUST clear the <a>pending pointer capture target override</a> for the <code>pointerId</code> of the <code>pointerup</code> or <code>pointercancel</code> event that was just dispatched, and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code> if necessary.</p>
+            <p>Immediately before firing <code>dragstart</code> event [[!HTML]],
+a user agent MUST clear the <a>pending pointer capture target override</a> for the pointer that caused the drag operation,
+and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code>
+before the <code>dragstart</code> event if necessary.</p>
```
mustaqahmed: Nit: can remove "... before the `dragstart` event..." because the whole para is about an action before `dragstart`.

index.html
```diff
@@ -861,6 +861,10 @@
...
             <p>If the user agent supports firing the <code>click</code> event (see <a title="compatibility mouse events" href="#dfn-compatibility-mouse-events">compatibility mouse events</a>) and if in an implicit release scenario both <code>click</code> and <code>lostpointercapture</code> events are fired, <code>click</code> should be fired before <code>lostpointercapture</code>.</p>
             <p>When the <a>pointer capture target override</a> is removed from its <code>ownerDocument</code>'s tree, clear the <a>pending pointer capture target override</a> and <a>pointer capture target override</a> nodes and fire a PointerEvent named <code>lostpointercapture</code> at the document.</p>
         </section>
```
mustaqahmed: Just realized that the `pointerId`of the `lostpointercapture` is not clear in this (existing) paragraph.  Could you please add line here?

mustaqahmed (approved):
LGTM with nits.

NavidZ:
This pull request may change quite a bit as a result of https://github.com/w3c/pointerevents/issues/205.
I'm waiting for @patrickkettner to reply on that to see whether they agree with that plan. But if we go we that plan we send a pointercancel before drag starts and that pointercancel would cause the loss of pointer capture.

index.html
NavidZ: I fixed this paragraph as well.

index.html
```diff
@@ -862,9 +862,18 @@
         <section>
             <h3>Implicit Release of Pointer Capture</h3>
             <p>Immediately after firing the <code>pointerup</code> or <code>pointercancel</code> events, a user agent MUST clear the <a>pending pointer capture target override</a> for the <code>pointerId</code> of the <code>pointerup</code> or <code>pointercancel</code> event that was just dispatched, and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code> if necessary.</p>
+            <p>Immediately before <a href="https://html.spec.whatwg.org/multipage/dnd.html#initiate-the-drag-and-drop-operation">drag operation starts</a> [[!HTML]],
+a user agent MUST send a <code>pointercancel</code> for the pointer that caused the drag operation.
+This <code>pointercancel</code> will follow the paragraph above regarding releasing the capture.</p>
+<div class="note">If the drag operation initiation is prevented through any means
```
mustaqahmed: To clarify with the first sentence here, you may use the same action word: "If the _start_ of the drag operation is prevented...".

Also "need" seems to make it sound like optional (or non-normative).  Let's omit it to say, "... there will be no `pointercancel` event".

index.html
```diff
@@ -862,9 +862,18 @@
         <section>
             <h3>Implicit Release of Pointer Capture</h3>
             <p>Immediately after firing the <code>pointerup</code> or <code>pointercancel</code> events, a user agent MUST clear the <a>pending pointer capture target override</a> for the <code>pointerId</code> of the <code>pointerup</code> or <code>pointercancel</code> event that was just dispatched, and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code> if necessary.</p>
+            <p>Immediately before <a href="https://html.spec.whatwg.org/multipage/dnd.html#initiate-the-drag-and-drop-operation">drag operation starts</a> [[!HTML]],
+a user agent MUST send a <code>pointercancel</code> for the pointer that caused the drag operation.
+This <code>pointercancel</code> will follow the paragraph above regarding releasing the capture.</p>
```
mustaqahmed: Sorry for not noticing this before: can't we move this paragraph to the Section 5.2.6 (`pointercancel`) above?  Then the second sentence here won't be needed because it's implied.  Moreover, the PR here is more about pointercancel than about implicit release.

appsforartists:
This addresses the original concerns in #205, but not the larger usability concern.  I've separated that into #213.

index.html
```diff
@@ -566,7 +566,12 @@
                 <ul>
                     <li>The user agent has determined that a pointer is unlikely to continue to produce events (for example, because of a hardware event).</li>
                     <li>After having fired the <code>pointerdown</code> event, if the pointer is subsequently used to manipulate the page viewport (e.g. panning or zooming).</li>
+                    <li>Immediately before <a href="https://html.spec.whatwg.org/multipage/dnd.html#initiate-the-drag-and-drop-operation">drag operation starts</a> [[!HTML]],
+                     for the pointer that caused the drag operation.</li>
                 </ul>
+<div class="note">If the start of the drag operation is prevented through any means
```
mustaqahmed: Nits:
- Can we move the Note to inside the last list element (`<li>...</li>`)?
- Raw text indentation.

mustaqahmed (approved):

PR #208, opened 2017-05-12
Fixed coordinate space for touch-action

mustaqahmed:
Related: #115


mustaqahmed:
@RByers: ptal. I don't think we need anything else for div vs frame rotation, right?

mustaqahmed:
This is simple enough to get merged now.  We can always correct later on.

index.html
```diff
@@ -768,7 +768,7 @@
...
+              <p>In the case of <code>pan-left</code>, <code>pan-right</code>, <code>pan-up</code> and <code>pan-down</code>, the direction is interpreted as the opposite of the physical movement in the local coordinate space. For example, <code>pan-up</code> corresponds to input event sequences where typically <code>screenY</code> is increasing (i.e. an interaction where the user moves a touch point down the screen) when the element has no effective CSS transforms. If the element is effectively rotated by 90 degrees counter-clockwise, <code>pan-up</code> would then correspond to input event sequences where <code>screenX</code> is increasing.</p></dd>
```
RByers, line 771: I guess saying `clientY` and `clientX` would now be more correct than referring to `screenY` and `screenX`, right?

RByers (commented):
Sorry for the delay.  Yeah that was definitely an improvement, but I think we can do a little better.

Also, can you add a test that covers the distinction here?  At least for the 90 degree rotated iframe case?  Given the current touch-action test architecture and cost/benefits trade offs I don't think it's worth trying to be exhaustive.

PR #210, opened 2017-05-16
Set target for coalesced events

### LINKED ISSUE #209: offsetX/offsetY is confusing for GetCoalescedEvents
See [this chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=721485) and the [definition](https://drafts.csswg.org/cssom-view/#dom-mouseevent-offsetx) of `offsetX`/`offsetY`.

Since an important use case of `GetCoalescedEvents` is to re-use (and share) existing code, perhaps we really should try harder to make these events look just like normal events?  Eg. copy the `target`, and set the "dispatch flag" so that `offsetX`/`offsetY` get computed relative to the target?

NavidZ:
Set target for coalesced events to
make sure attributes like offsetX/Y
are calculated correctly.

Closes #209

### DIFF
--- a/extension.html
+++ b/extension.html
@@ -81,7 +81,7 @@ In addition to that, the other attibutes related to the <a href="https://www.w3.
 (e.g. <code>currentTarget</code>, <code>eventPhase</code>) will have their default value
 and the related functions (e.g. <code>stopPropagation</code>, <code>stopImmediatePropagation</code>) will do nothing.
 Note that this rule doesn't apply to <code>target</code> and it should still be the same as the dispatched event's <code>target</code>.
-This guarantees the attributes like <code>offsetX</code>, <code>offsetY</code> (which are commonly used by drawing applications) to be calculated correctly for the coalesced events.
+This guarantees the attributes like <code>offsetX</code>, <code>offsetY</code> ([[!CSSOM-VIEW]]) which are commonly used by drawing applications to be calculated correctly for the coalesced events.
 The rest of the attributes might be different from the dispatched event.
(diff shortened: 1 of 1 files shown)


NavidZ:
@RByers 
I also broke the long paragraph into multiple lines in the html to make it easier to review the changes in the future. I also relaxed the conditions regarding the timestamp and attributes like movementX/Y of the dispatched event with respect to the coalesced events

extension.html
```diff
@@ -73,7 +73,22 @@
...
+When the event is created by the user agent the following attributes of the coalesced events will always have the same value as the dispatched event:
+<code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>, <code>isTrusted</code>, <code>target</code>.
+Also since these coalesced events are not going to be dispatched by themselves their <code>cancelable</code> and <code>bubbles</code> attributes are false.
+In addition to that, the other attibutes related to the <a href="https://www.w3.org/TR/uievents/#event-flow">event dispatch algorithm</a>
+(e.g. <code>currentTarget</code>, <code>eventPhase</code>) will have their default value
+and the related functions (e.g. <code>stopPropagation</code>, <code>stopImmediatePropagation</code>) will do nothing.
+Note that this rule doesn't apply to <code>target</code> and it should still be the same as the dispatched event's <code>target</code>.
+This guarantees the attributes like <code>offsetX</code>, <code>offsetY</code> (which are commonly used by drawing applications) to be calculated correctly for the coalesced events.
```
dtapuska: Can you reference the CSS OM spec here for these attributes?

dtapuska (commented):
lgtm

PR #231, opened 2018-02-06
Tweaks in respec config and fixing data-lt attributes

plehegar:
This fixes some respect bugs and take into account latest respect goodies.


patrickhlauke:
Thank you

PR #233, opened 2018-03-21
Add possibility of having multiple pointers canceled by user agents

### LINKED ISSUE #226: Can one active pointer cancel other pointers of different types?
The spec should be clear if a touch-action initiated by one pointer cancels all other pointers or not.

The model Chrome (and perhaps Edge also) supports today is that any scroll/zoom caused by _some_ active pointers triggers firing of `pointercancel` for _all_ active pointers.  This works great for multiple touch pointers because all our fingers on the touchscreen effectively define a single gesture.  But if we combine multiple pointer types, the "single gesture" idea is neither obvious nor intuitive.

Consider the case of dragging a finger on one `<div>` and while doing it, tapping on another `<div>` with a pen.  It would be wrong to say it's a pinch gesture, right?  Assume these two `<div>`s are siblings in the DOM to make this assumption more realistic.

Now consider another scenario: dragging a finger on a `touch-action: none` region and at the same time dragging a pen on a `touch-action: auto` region.  Again assume the regions are siblings in DOM, to keep effective touch-action computation off from our discussion.  Should one pointer's browser action cancel the other pointer?

I still think canceling _all_ pointers when a browser action starts is the simplest thing to do regardless of our expectations in the above scenario.  Anyone has a different opinion?

NavidZ:
Closes #226.

### DIFF
--- a/index.html
+++ b/index.html
@@ -561,7 +561,7 @@ interface PointerEvent : MouseEvent {
                 <ul>
                     <li>The user agent has determined that a pointer is unlikely to continue to produce events (for example, because of a hardware event).</li>
                     <li>After having fired the <code>pointerdown</code> event, if the pointer is subsequently used to manipulate the page viewport (e.g. panning or zooming).
-                     <div class="note">User agents can trigger panning or zooming through multiple pointer types (say touch and pen),
+                     <div class="note">User agents can trigger panning or zooming through multiple pointer types (such as touch and pen),
                      and therefore the start of a pan or zoom action may result in the cancellation of various pointers, including pointers with different pointer types.</div></li>
                     <li>Immediately before <a href="https://html.spec.whatwg.org/multipage/dnd.html#initiate-the-drag-and-drop-operation">drag operation starts</a> [[!HTML]],
                      for the pointer that caused the drag operation.
--- a/index.html
+++ b/index.html
@@ -561,8 +561,8 @@ interface PointerEvent : MouseEvent {
                 <ul>
                     <li>The user agent has determined that a pointer is unlikely to continue to produce events (for example, because of a hardware event).</li>
                     <li>After having fired the <code>pointerdown</code> event, if the pointer is subsequently used to manipulate the page viewport (e.g. panning or zooming).
-                     <div class="note">User agents can trigger panning or zooming through multiple pointer types (say "touch" and "pen"),
-                     and therefore can possibly cancel all pointers of those types for a single viewport manipulation.</div</li>
+                     <div class="note">User agents can trigger panning or zooming through multiple pointer types (say touch and pen),
+                     and therefore the start of a pan or zoom action may result in the cancellation of various pointers, including pointers with different pointer types.</div></li>
                     <li>Immediately before <a href="https://html.spec.whatwg.org/multipage/dnd.html#initiate-the-drag-and-drop-operation">drag operation starts</a> [[!HTML]],
                      for the pointer that caused the drag operation.
                      <div class="note">If the start of the drag operation is prevented through any means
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -560,7 +560,9 @@ <h3>The <dfn><code>pointercancel</code> event</dfn></h3>
                 <p>A user agent MUST <a href="#firing-events-using-the-pointerevent-interface">fire a pointer event</a> named <code>pointercancel</code> in the following circumstances:</p>
                 <ul>
                     <li>The user agent has determined that a pointer is unlikely to continue to produce events (for example, because of a hardware event).</li>
-                    <li>After having fired the <code>pointerdown</code> event, if the pointer is subsequently used to manipulate the page viewport (e.g. panning or zooming).</li>
+                    <li>After having fired the <code>pointerdown</code> event, if the pointer is subsequently used to manipulate the page viewport (e.g. panning or zooming).
+                     <div class="note">User agents can use multiple pointers for a particular action like scrolling or zooming.
```
mustaqahmed: Please emphasize pointer _types_.  Something like:
> User agents can trigger panning or zooming through multiple pointer types (say "touch" and "pen"), and therefore can possibly cancel all pointers of those types for a single viewport manipulation.

index.html
```diff
@@ -560,7 +560,9 @@ <h3>The <dfn><code>pointercancel</code> event</dfn></h3>
...
+                     In that case they can cancel all the pointers they deem necessary.</div></li>
```
mustaqahmed: Nit: the indentation looks odd around here.  Please unindent by 1 char here and Lines 567-570.

mustaqahmed (changes requested):

index.html
mustaqahmed: Just realized that the diff viewer in github doesn't isolate whitespace changes.  So let's not change whitespaces here.

index.html
```diff
@@ -561,8 +561,8 @@ <h3>The <dfn><code>pointercancel</code> event</dfn></h3>
                 <ul>
                     <li>The user agent has determined that a pointer is unlikely to continue to produce events (for example, because of a hardware event).</li>
                     <li>After having fired the <code>pointerdown</code> event, if the pointer is subsequently used to manipulate the page viewport (e.g. panning or zooming).
-                     <div class="note">User agents can use multiple pointers for a particular action like scrolling or zooming.
-                     In that case they can cancel all the pointers they deem necessary.</div></li>
+                     <div class="note">User agents can trigger panning or zooming through multiple pointer types (say "touch" and "pen"),
```
patrickhlauke: "(say "touch" or "pen")" may sound better as "(such as touch and pen)"

index.html
```diff
@@ -561,8 +561,8 @@ <h3>The <dfn><code>pointercancel</code> event</dfn></h3>
...
+                     and therefore can possibly cancel all pointers of those types for a single viewport manipulation.</div</li>
```
patrickhlauke: i'd reword the whole "and therefore...manipulation" to "and therefore the start of a pan or zoom action may result in the cancellation of various pointers, including pointers with different pointer types." or similar. Seems a bit clearer (assuming I understood this right)

patrickhlauke (changes requested):

index.html
```diff
@@ -561,8 +561,8 @@ <h3>The <dfn><code>pointercancel</code> event</dfn></h3>
                 <ul>
                     <li>The user agent has determined that a pointer is unlikely to continue to produce events (for example, because of a hardware event).</li>
                     <li>After having fired the <code>pointerdown</code> event, if the pointer is subsequently used to manipulate the page viewport (e.g. panning or zooming).
-                     <div class="note">User agents can trigger panning or zooming through multiple pointer types (say "touch" and "pen"),
-                     and therefore can possibly cancel all pointers of those types for a single viewport manipulation.</div</li>
+                     <div class="note">User agents can trigger panning or zooming through multiple pointer types (say touch and pen),
```
patrickhlauke: nitpick: i'd still change "say" to "such as" (the former sounds a tad too informal/colloquial)

patrickhlauke (approved):

mustaqahmed (approved):

PR #234, opened 2018-03-22
Added a note on null fromElement and toElement.

### LINKED ISSUE #167: Behavior of PointerEvent.fromElement/toElement
fromElement/toElement [are being spec'd](https://github.com/w3c/pointerevents/issues/605) and it [looks like](https://github.com/w3c/web-platform-tests/pull/4408#issuecomment-271921770) there are some oddities in the implementation of these properties for pointer events in both Chrome and Edge.  Once the UI Events spec text has landed, does PointerEvents need to override it at all?  Even if not, we probably need a test.

@mustaqahmed says he'll look into it

/cc @cvrebert @dtapuska

mustaqahmed:
Closes #167.

### DIFF
--- a/index.html
+++ b/index.html
@@ -436,79 +436,82 @@ interface PointerEvent : MouseEvent {
 
                 <section>
                   <h3>Attributes and Default Actions</h3>
-                        <p>The <code>bubbles</code> and <code>cancelable</code> properties and the default actions for the event types defined in this specification appear in the following table. Details of each of these event types are provided in <a href="#h-pointer-event-types">Pointer Event types</a>.</p>
-                        <table id="pointer-event-type-table" class="parameters">
-                            <thead><tr>
-                                <th>Event Type</th><th>Bubbles</th><th>Cancelable</th><th>Default Action</th></tr>
-                            </thead>
-                            <tbody>
-                                <tr>
-                                    <td><code>pointerover</code></td>
-                                    <td>Yes</td>
-                                    <td>Yes</td>
-                                    <td>None</td>
-                                </tr>
-                                <tr>
-                                    <td><code>pointerenter</code></td>
-                                    <td>No</td>
-                                    <td>No</td>
-                                    <td>None</td>
-                                </tr>
-                                <tr>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -436,79 +436,82 @@ <h2>Firing events using the <code>PointerEvent</code> interface</h2>
...
+                  <p>Similar to the <a href="https://www.w3.org/TR/uievents/#interface-mouseevent">MouseEvents</a> [[!DOM-LEVEL-3-EVENTS]] the <code>relatedTarget</code> should be initialized to the element whose bounds the pointer just left (in the case of a <code>pointerover</code> or <code>pointerenter</code> event) or the element whose bounds the pointer is entering (in the case of a <code>pointerout</code> or <code>pointerleave</code>). For other pointer events, this value will default to null. Note that when an element receives the pointer capture all the following events for that pointer are considered to be inside the boundary of the capturing element.</p>
+
+                  <div class="note">Many user agents expose non-standard attributes <code>fromElement</code> and <code>toElement</code> in MouseEvents to support legacy content.  In those user agents, the values of those (inherited) attributes in PointerEvents must be <code>null</code> to encourage the use of the standardized alternates (i.e. <code>target</code> and <code>relatedTarget</code>).</div>
```
NavidZ: Do you want to fold this along with the detail attribute in the earlier paragraph? I guess they all fall into the same bucket.
Also beside this paragraph there is no other change in other lines except indentation and extra spaces. Right? if you remove those that might be easier to catch the actual changes and review them.

mustaqahmed: - The problem is that (unlike `detail` and `composed`) these two attributes are not spec-ed or even described clearly anywhere.  I think a normative text won't make sense.  I moved the Note up, right after the para on other attributes. 
 Does it look reasonable now?

- Yikes, I only looked at a local diff tool so didn't realize that whitespace changes are not visible here.  Sorry.

NavidZ (approved):

NavidZ:
Thanks. looks good to me now.

PR #237, opened 2018-03-29
Remove local biblio entry

EiraGe:
No need to have local biblio entry to DOM4
and it's not Last Call Working Draft anymore :)


NavidZ:
Thanks Ella for fixing this. LGTM.

PR #238, opened 2018-03-29
Move single directional touch-action values to the extension spec

NavidZ:


patrickhlauke:
can you just remind me of the rationale for this? i lost track of where we discussed it / the reasons for moving

NavidZ:
This was actually implemeted in Chrome two years ago. @smaug---- told that Firefox is not going to be able to have this any time soon. That's why we said okay let's move it to the extension. What do you think we should do?

patrickhlauke:
Ah, gotcha. That's fine, it just slipped my mind why we did this, but makes sense to move it to extension. I assume the situation is similar for Edge then.

patrickhlauke (approved):

lpd-au:
Given the extra time, will this be included again for level 3?

NavidZ:
@lpd-au that's the plan.

Issue #160, opened 2016-12-19
Spec references both W3C DOM4 and WHATWG-DOM


RByers:
The definition of `composed` needed to reference WHATWG-DOM (since it's a new feature not yet reflected in the W3C copy).  IMHO we should just be consistent and reference WHATWG-DOM instead of DOM4 everywhere to avoid confusion about which algorithms exactly we were referring to (eg. see [this thread](https://groups.google.com/a/chromium.org/d/msg/blink-dev/T_amUSoAzHA/J2QsfpGKDwAJ))

patrickhlauke:
Potentially political minefield? Personally, I would say only do it for those cases where currently WHATWG-DOM is more accurate than DOM4?

chaals:
Actually, we are working on updating the W3C DOM spec. So please [file issues](https://github.com/w3c/dom/issues) for anything that you're missing and need so we can prioritise it.

In the next few days we'll give some more concrete information about the plans. I know the timing for explaining stuff is horrible for some places, but I have been working in countries where Christmas is either in January or not really a thing...

RByers:
Ok, filed [this issue](https://github.com/w3c/dom/issues/13).

patrickhlauke:
check to see if we can reference https://www.w3.org/TR/dom41/ / https://www.w3.org/TR/dom41/#composed-flag or whatever

NavidZ:
I'm going to talk to the person who added composed here in the spec and see whether it is okay whether to reference [this spec](https://github.com/w3c/dom) instead of whatwg.

NavidZ:
We have been referencing DOM4 in our document which apparently has the same entry as WHATWG-DOM in the reference. @patrickhlauke  do you think we should just switch them all to one? Say WHATWG-DOM? We'll see whether any objection rises from that. Note that TouchEvent spec has the same problem of referencing both and having the same entry twice in the reference section.

patrickhlauke:
to avoid politics - assuming DOM4 says the same now as WHATWG-DOM - i'd prefer just changing that tiny reference to DOM4. functionally the same, but avoids potential raised eyebrows and pearl-clutching at review time.

patrickhlauke:
Closed via #240 (and https://github.com/w3c/pointerevents/commit/e9bad530669af40ddf3ae5241c43c766da9d6dca)

PR #240, opened 2018-03-29
Update WHATWG-DOM to DOM4

NavidZ:
This is the only reference to WHATWG-DOM that we had.


NavidZ:
Closes #160

patrickhlauke (approved):

domenic:
This is very unfortunate, as it makes the spec unimplementable as-is. (Event dispatch is quite broken in the W3C DOM fork.)

patrickhlauke:
@domenic looking at the end result, i think the web isn't going to break just yet and the existing implementations are saved...the actual change now just points to [DOM4] https://w3c.github.io/pointerevents/#bib-DOM4 which in the normative references, it seems, still links to the living standard...

Issue #118, opened 2016-07-26
Clarify duration of pointer event suppression during a touch action


RByers:
[The spec](https://w3c.github.io/pointerevents/#the-touch-action-css-property) says:

> During the execution of a user agent touch behavior, the user agent must not fire subsequent pointer events for the pointer. 

Edge implements this by suppressing all pointer events as long as the touch-action is in progress, including momentum that follows (so repeated flicks on the same element don't send additional pointerdown).  Chrome implements this only until the next time a finger goes down.  Consider something like "subsequent touches that may contribute to that touch behavior will not generate pointer events").

RByers:
Andrew points out that the Edge behavior also generally results in a better user experience.  When you're doing a sequence of many flicks, if you happen to accidentally hit a touch-action:none region you really don't expect the scroll to grind to a stop.

RByers:
[Blink bug to match Edge behavior](https://bugs.chromium.org/p/chromium/issues/detail?id=631899).

NavidZ:
We talked about this and @patrickhlauke is going to add a informative text about this since it is hard to define normatively as underlying gestures are not well defined at this point.

PR #243, opened 2018-03-30
 Add informative note about UA handling of "compound" gestures

patrickhlauke:
Closes https://github.com/w3c/pointerevents/issues/118


patrickhlauke:
It would be good to make sure that this doesn't have language that makes legal people worried about specific names for gestures or similar.

NavidZ:
I'm a little late to the party :)
The changes looks good. For the future, just a little editorial thing from the html spec. They usually add a long paragraph in a few lines (just a few enters here and there which doesn't change the html output) so if next time someone changes part of it, it is easier to see what part is changed for the reviewers.

patrickhlauke:
ah yes, good point. may do a code reformat once we're close to stable.

Issue #221, opened 2017-10-02
Should setPointerCapture work in Shadow DOM


smaug----:
Per spec setPointerCapture shouldn't work in Shadow DOM, but f I read for example Chromium's source code right, it doesn't use "does not participate in its ownerDocument's tree", but Shadow including connected() check.

patrickhlauke:
@mustaqahmed @NavidZ any thoughts on this? is it a chrome bug, or something wrong that needs specifying in the spec?

mustaqahmed:
This looks like a crack in Chrome, we haven't been careful about Shadow DOM when we added the [check](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/dom/Element.cpp?rcl=195db4ec8d8bc4fc650015fdd86de98ce81748d1&l=3372).

Thanks for noticing this.  I will file a chromium bug.

mustaqahmed:
[Here](https://crbug.com/810882) is the Chromium bug.  I don't think we need to change the spec, so closing the issue.

smaug----:
I'm pretty sure we want setPointerCapture to work in Shadow DOM too.

patrickhlauke:
@smaug @mustaqahmed do you think this needs to be explicitly spec'd in the PE spec? if so, any pointers where/how?

smaug----:
Yes, it needs to be spec'ed explicitly in PointerEvents. 
Right now the spec says 
https://w3c.github.io/pointerevents/#dfn-pointer-capture
"If the Element on which this method is invoked does not participate in its ownerDocument's tree, throw an exception with the name InvalidStateError."
That needs to say something else than 'does not participate in its ownerDocument's tree

Maybe something like 'if element's owner document is not shadow-including ancestor of the element'
https://dom.spec.whatwg.org/#concept-shadow-including-ancestor
(I wonder if some spec has better language for this. Maybe HTML spec)

NavidZ:
In the call we agreed to follow the wording of other specs for this.

PR #246, opened 2018-04-02
Replace prose about `ownerDocument` tree with DOM4 concept of `connected`

patrickhlauke:
Closes https://github.com/w3c/pointerevents/issues/221


patrickhlauke:
As this goes outside of my comfort zone, I'd appreciate a good hard look from @smaug---- and @NavidZ to make sure this makes sense?

smaug----:
That change seems to capture the requirement.
The same thing could be possibly expressed with 'connected'.
https://dom.spec.whatwg.org/#connected

patrickhlauke:
what would be clearer do you think? something like

```
If the <code>Element</code> on which this method is invoked is not <code>connected</code> [[!DOM4]] - i.e. if the <code>ownerDocument</code> of the <code>Element</code> is not a shadow-including ancestor [[!DOM4]] of the element - throw an exception with the name <code>InvalidStateError</code>
```

smaug----:
Using 'connected' might be easier to read, though if one is not familiar with the term, one needs to look at the DOM spec. Anyhow, 'connected' should be fine.

patrickhlauke:
thanks @smaug----

PR #251, opened 2018-04-19
Add boundary events in implicit capture scenarios

### LINKED ISSUE #173: Should spec defines what triggers pointerover and pointerenter when implicitly releasing pointer capture?
I knew this had been discussed in #145, but I'm still not sure about what triggers pointerover and pointerenter when implicit releasing pointer capture.

When the pointer is captured, hit test target is the pointer capture target override object. That means UA won't run the normal hit test algorithm. When implicitly releasing pointer capture, the hit test target is changed, and UA must fire boundary events. Should spec define what triggers UA run the normal hit test algorithm to determine the hit test target and fire pointerover / pointerenter to it?

e.g. When the pointer is captured by element A while pointer is hovering on element B, pointer is implicitly released when pointerup. What trigger UA to run the hit test algorithm and fire pointerover / pointerenter to B (in the case of the pointer isn't moved) ?

Tested on Chrome, boundary events are fired with a pointermove event. Edge fires boundary events with the subsequent pointermove event (happened when the pointer is moved).

NavidZ:
closes #173

### DIFF
--- a/index.html
+++ b/index.html
@@ -837,7 +837,14 @@ partial interface Navigator {
         </section>
         <section>
             <h3>Implicit Release of Pointer Capture</h3>
-            <p>Immediately after firing the <code>pointerup</code> or <code>pointercancel</code> events, a user agent MUST clear the <a>pending pointer capture target override</a> for the <code>pointerId</code> of the <code>pointerup</code> or <code>pointercancel</code> event that was just dispatched, and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code> if necessary.</p>
+            <p>
+               Immediately after firing the <code>pointerup</code> or <code>pointercancel</code> events,
+               a user agent MUST clear the <a>pending pointer capture target override</a>
+               for the <code>pointerId</code> of the <code>pointerup</code> or <code>pointercancel</code> event that was just dispatched,
+               and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code> if necessary.
+               After running <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps,
+               if the pointer supports hover user agent MUST also send corresponding boundary events if necessary to reflect the current position of the pointer with no capture.
+            </p>
             <p>If the user agent supports firing the <code>click</code> event (see <a title="compatibility mouse events" href="#dfn-compatibility-mouse-events">compatibility mouse events</a>) and if in an implicit release scenario both <code>click</code> and <code>lostpointercapture</code> events are fired, <code>click</code> should be fired before <code>lostpointercapture</code>.</p>
             <p>When the <a>pointer capture target override</a> is removed from its <code>ownerDocument</code>'s tree,
 clear the <a>pending pointer capture target override</a> and <a>pointer capture target override</a> nodes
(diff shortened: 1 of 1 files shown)


NavidZ:
@mingchou

index.html
```diff
@@ -837,7 +837,14 @@ <h2><dfn>Implicit Pointer Capture</dfn></h2>
...
+            <p>
+               Immediately after firing the <code>pointerup</code> or <code>pointercancel</code> events,
+               a user agent MUST clear the <a>pending pointer capture target override</a>
+               for the <code>pointerId</code> of the <code>pointerup</code> or <code>pointercancel</code> event that was just dispatched,
+               and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code> if necessary.
+               After running <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps,
+               if the pointer supports hover user agent MUST also send corresponding boundary events if necessary to reflect the current position of the pointer with no capture.
```
patrickhlauke: "pointer supports hover, the user agent MUST"

index.html
```diff
@@ -837,7 +837,14 @@ <h2><dfn>Implicit Pointer Capture</dfn></h2>
         </section>
         <section>
             <h3>Implicit Release of Pointer Capture</h3>
-            <p>Immediately after firing the <code>pointerup</code> or <code>pointercancel</code> events, a user agent MUST clear the <a>pending pointer capture target override</a> for the <code>pointerId</code> of the <code>pointerup</code> or <code>pointercancel</code> event that was just dispatched, and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code> if necessary.</p>
+            <p>
```
patrickhlauke: a minor thing, but maybe worth not leaving `<p>` and `</p>` hanging on their own lines

index.html
```diff
@@ -837,7 +837,13 @@ <h2><dfn>Implicit Pointer Capture</dfn></h2>
...
+               and then run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps to fire <code>lostpointercapture</code> if necessary.
+               After running <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps,
+               if the pointer supports hover, user agent MUST also send corresponding boundary events if necessary
+               to reflect the current position of the pointer with no capture. </p>
             <p>If the user agent supports firing the <code>click</code> event (see <a title="compatibility mouse events" href="#dfn-compatibility-mouse-events">compatibility mouse events</a>) and if in an implicit release scenario both <code>click</code> and <code>lostpointercapture</code> events are fired, <code>click</code> should be fired before <code>lostpointercapture</code>.</p>
             <p>When the <a>pointer capture target override</a> is removed from its <code>ownerDocument</code>'s tree,
 clear the <a>pending pointer capture target override</a> and <a>pointer capture target override</a> nodes
```
mustaqahmed: Bike-shedding on an old text, sorry: the imperative verb only in this para looks odd.  Could you please change it too?  Something like:
> the pending pointer capture target override and pointer capture target override nodes *should be cleared* and also a lostpointercapture event corresponding to the captured pointer *should be fired*.

mustaqahmed (approved):
Looks good.  Requesting a minor tweak to make the section consistent.

PR #252, opened 2018-04-24
Additional security/privacy clarifications following horizontal review

patrickhlauke:
See https://lists.w3.org/Archives/Public/public-pointer-events/2018AprJun/0046.html


NavidZ:
Looks good to me.
But I suggest breaking the paragraphs into multiple lines in the file (rather than having the whole paragraph in one line). Having multiple lines makes reviewing the changes in the future much easier.

patrickhlauke:
Ah yes sorry. I think the breaking up of lines is something that we'd need to do document-wide, once we're happy it's stable (as the single long paragraph structure is pretty much pervasive throughout).

patrickhlauke:
@smaug---- any thoughts on this? (think it's fairly non-controversial)

smaug----:
I guess that is good enough. Perhaps Tor will provide some default values for sensor related data.

PR #253, opened 2018-05-01
GlobalEventHandlers is a mixin

saschanaz:
[GlobalEventHandlers](https://html.spec.whatwg.org/multipage/webappapis.html#globaleventhandlers) is now an `interface mixin`, so extensions should use [`partial interface mixin`](https://heycam.github.io/webidl/#partial-interface-mixin).


saschanaz:
Oh, the spec refers W3C HTML for this part which still has older Web IDL syntax with legacy `NoInterfaceObject`. Hmm.

patrickhlauke:
Discussed in the call 2 May. Closing this for now, but will revisit in future (assuming W3C spec may update to match)

PR #257, opened 2018-06-06
Propose a getPredictedEvents API with example

### LINKED ISSUE #38: Consider adding API to get predicted points
Some platforms have built-in prediction algorithms which apps would ideally handle differently from confirmed points (eg. a painting app would use predicted points only temporarily).  We could have an API directly similar to #22 but in the opposite time direction (instead of the past points, it's the future ones).

NavidZ:
Closes #38


patrickhlauke:
merging this at this stage, as further discussion can happen directly on the extension document in future.

PR #260, opened 2018-07-04
Introduce pointerrawupdate to address the high frequency usecases

### LINKED ISSUE #214: Define a low latency event that isn't occur in the document lifecycle
Chrome has experimented in shipping events relative to rAF. pointermove is only sent once for every rAF callback and occurs just before the rAF callback. Specifically this was done for touch in Chrome 59 and mouse in Chrome 60. See some discussion here #9.

With pointerevent dispatched at rAF time it is useful for some apps that don't produce any in the normal document lifecycle (see issue whatwg/html#2659) to have input right away.

I was wondering if it we should add a pointermove-raw event to enable the cases when pointer moves with low latency are needed. We want to ensure that this isn't the typical usage as the raw events can occur at a much higher frequency (eg. 1000Hz) so I don't think we want to use the same event type as that would conflate the problem. 

Talking with @RByers we discussed that this event shouldn't bubble and is non-cancelable.

NavidZ:
Fixes #214

### DIFF
--- a/extension.html
+++ b/extension.html
@@ -59,14 +59,26 @@ The features in this specification extend or modify those found in <a href="http
       <h2>Extensions to the Pointer Event types</h2>
       <p>Below is the new event type that is added to the <a href="https://w3c.github.io/pointerevents/index.html#pointer-event-types">existing pointer event types</a>.</p>
       <section>
+        <h3>The <dfn><code>pointermove</code> event</dfn></h3>
+        <p>A user agent MUST <a href="#firing-events-using-the-pointerevent-interface">fire a pointer event</a> named <code>pointermove</code> when a pointer changes button state.
+        Additionally one <code>pointermove</code> MUST be fired when pointer changes coordinates, pressure, tangential pressure, tilt, twist, or
+        contact geometry (e.g. <code>width</code> and <code>height</code>) and the circumstances produce no other pointer events defined in this specification.
+        This <code>pointermove</code> that is not caused by button state change MUST be fired only once right before firing an event caused by button state change or
+        <a href="https://html.spec.whatwg.org/#event-loop-processing-model">animation frame callbacks</a>. If there are mutliple changes to these attributes
+        the <code>pointermove</code> events before reaching animation frame callback or a button state change event,
+        these events will get coalesced and will be exposed via <a data-lt="getCoalescedEvents">getCoalescedEvents</a> API for the single dispatched <code>pointermove</code> event.
+        The final coordinates of such events should be used for finding the target of the event.
+      </section>
+      <section>
         <h3>The <dfn><code>pointerrawmove</code> event</dfn></h3>
         <p>A user agent MUST <a href="https://w3c.github.io/pointerevents/index.html#firing-events-using-the-pointerevent-interface">fire a pointer event</a>
-           named <code>pointerrawmove</code> when a pointing device attribute is changed.
-           As oppose to <code>pointermove</code> which could be fired once every frame and will have coalesced event data,
+           named <code>pointerrawmove</code> when a pointing device attribute
+           (i.e. coordinates, pressure, tangential pressure, tilt, twist, or contact geometry) is changed.
+           As oppose to <code>pointermove</code> which could be delayed until an animation frame for changes in these attributes,
            <code>pointerrawmove</code> will get fired as soon as an attribute is changed
            and as frequent as the javascript can handle the events.
            Because of that the target of <code>pointerrawmove</code> events might be different from the <code>pointermove</code> events
-           due to the fact that <code>pointermove</code> events are being sent once every frame and the final position of the even
+           due to the fact that <code>pointermove</code> events are being coalesced and the final position of the event
            which is used for finding the target could be different from its coalesced events.
            Note that if the javascript handlers are slow, <code>pointerrawmove</code> events can also get coalesced.
            The attributes of <code>pointerrawmove</code> are mostly the same as <code>pointermove</code> with the exception of
@@ -133,7 +145,7 @@ So the first event will have the smallest <code>timeStamp</code> which should st
 
     <section>
       <h2>Timing of coalesced event firing</h2>
-      <p>Coalescing events only happens for <code>pointermove</code> events. All the coalesced events should be fired before the next frame or before any <code>pointerup</code>, <code>pointerdown</code>, <code>pointercancel</code> events. Note that user agents don't need to hit-test all the coalesced events as none of them will have null targets.</p>
+      <p>Coalescing events only happens for <code>pointermove</code> events. All the coalesced events should be fired before the next frame or before any <code>pointerup</code>, <code>pointerdown</code>, <code>pointercancel</code> events. Note that user agents don't need to hit-test all the coalesced events as all of them will have the same target as the dispatched event.</p>
       <p>When the user agent dispatches any <code>PointerEvent</code> it MUST dispatch all the events held back so far for coalescing, regardless of their <code>pointerIds</code>. The ordering of all these dispatched events should be in a way that resemebles the most with the actual events' ordering. For example if a <code>pointerdown</code> event causes the dispatch for the coalesced <code>pointermove</code> events the user agent SHOULD first dispatch one <code>pointermove</code> event with all those coalesced events of a <code>pointerId</code> followed by the <code>pointerdown</code> event.</p>
       <p>Here is an example of the actual events happening in one frame with increasing <code>timestamps</code> and the events dispatched by the user agent:
         <table class="parameters">
--- a/extension.html
+++ b/extension.html
@@ -73,7 +73,7 @@ The features in this specification extend or modify those found in <a href="http
         <h3>The <dfn><code>pointerrawmove</code> event</dfn></h3>
         <p>A user agent MUST <a href="https://w3c.github.io/pointerevents/index.html#firing-events-using-the-pointerevent-interface">fire a pointer event</a>
            named <code>pointerrawmove</code> when a pointing device attribute
-           (i.e. coordinates, pressure, tangential pressure, tilt, twist, or contact geometry) is changed.
+           (i.e. button state, coordinates, pressure, tangential pressure, tilt, twist, or contact geometry) is changed.
            As oppose to <code>pointermove</code> which could be delayed until an animation frame for changes in these attributes,
            <code>pointerrawmove</code> will get fired as soon as an attribute is changed
            and as frequent as the javascript can handle the events.
--- a/extension.html
+++ b/extension.html
@@ -56,6 +56,25 @@ The features in this specification extend or modify those found in <a href="http
       </p>
     </section>
     <section>
+      <h2>Extensions to the Pointer Event types</h2>
+      <p>Below is the new event type that is added to the <a href="https://w3c.github.io/pointerevents/index.html#pointer-event-types">existing pointer event types</a>.</p>
+      <section>
+        <h3>The <dfn><code>pointerrawmove</code> event</dfn></h3>
+        <p>A user agent MUST <a href="https://w3c.github.io/pointerevents/index.html#firing-events-using-the-pointerevent-interface">fire a pointer event</a>
+           named <code>pointerrawmove</code> when a pointing device attribute is changed during the capture.
+           Note that the target of such event will always be the captured node.
+           The captured node should expect such an event only after
+           it received <code>gotpointercapture</code> and until it receives <code>lostpointercapture</code>.
+           As oppose to <code>pointermove</code> which could be fired once every frame and will have coalesced event data,
+           <code>pointerrawmove</code> will get fired as soon as an attribute is changed
+           and as frequent as the javascript can handle the events.
+           Note that if javascript is slow, <code>pointerrawmove</code> events can also get coalesced.
+           The attributes of <code>pointerrawmove</code> are mostly the same as <code>pointermove</code> with the exception of
+           <code>bubbles</code> and <code>cancelable</code> which MUST be false for <code>pointerrawmove</code>.
+           User agent SHOULD not fire <a href="https://w3c.github.io/pointerevents/index.html#dfn-compatibility-mouse-events">compatibility mouse events</a> for <code>pointerrawmove</code>.</p>
+      </section>
+    </section>
+    <section>
       <h2>Extensions to the <code>PointerEvent</code> interface</h2>
       <p>
         The following section describes extensions to the existing <a href="https://w3c.github.io/pointerevents/index.html#pointerevent-interface"><code>PointerEvent</code> interface</a>.
@@ -75,7 +94,7 @@ partial interface PointerEvent {
           <dl dfn-for="PointerEvent" link-for="PointerEvent">
               <dt><dfn>getCoalescedEvents</dfn></dt>
               <dd>
-                  <p>Returns a sequence of all <code>PointerEvents</code> that were coalesced into the dispatched <code>pointermove</code> event.
+                  <p>Returns a sequence of all <code>PointerEvents</code> that were coalesced into the dispatched <code>pointermove</code> or <code>pointerrawmove</code> event.
 When the event is created by the user agent the following attributes of the coalesced events will always have the same value as the dispatched event:
 <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>, <code>isTrusted</code>, <code>target</code>.
 Also since these coalesced events are not going to be dispatched by themselves their <code>cancelable</code> and <code>bubbles</code> attributes are false.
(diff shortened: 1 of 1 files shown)


NavidZ:
@smaug---- what do you think of pointerrawmove to address the low latency usecases? I tried to remove the overhead of hittesting on every one of those events by restricting it to only capturing mode.
I wanted to start the implementation in Chrome and wanted to know what you think of it first.

smaug----:
(discussion about features shouldn't happen in PRs, but in issues)

extension.html
```diff
@@ -55,6 +55,25 @@ <h2>Introduction</h2>
...
+      <p>Below is the new event type that is added to the <a href="https://w3c.github.io/pointerevents/index.html#pointer-event-types">existing pointer event types</a>.</p>
+      <section>
+        <h3>The <dfn><code>pointerrawmove</code> event</dfn></h3>
+        <p>A user agent MUST <a href="https://w3c.github.io/pointerevents/index.html#firing-events-using-the-pointerevent-interface">fire a pointer event</a>
+           named <code>pointerrawmove</code> when a pointing device attribute is changed.
+           As oppose to <code>pointermove</code> which could be fired once every frame and will have coalesced event data,
+           <code>pointerrawmove</code> will get fired as soon as an attribute is changed
+           and as frequent as the javascript can handle the events.
+           Because of that the target of <code>pointerrawmove</code> events might be different from the <code>pointermove</code> events
+           due to the fact that <code>pointermove</code> events are being sent once every frame and the final position of the even
```
smaug----: pointermove isn't defined to be sent once every frame.

extension.html
```diff
@@ -55,6 +55,25 @@ <h2>Introduction</h2>
...
+           which is used for finding the target could be different from its coalesced events.
+           Note that if the javascript handlers are slow, <code>pointerrawmove</code> events can also get coalesced.
```
smaug----: ok, this makes the behavior from specification rather odd. pointermove isn't defined to be only once per frame and then pointerrawmove may get coalesced too, so in practice the events could be exactly the same.

smaug---- (changes requested):

extension.html
NavidZ: You are right. This part is not yet specced. Not sure how to get around it. More on it in the other comment.

extension.html
NavidZ: The idea was that pointermove will get delayed at the very least until right before rAF. But pointerrawmove will not have that delay at all. However both of these events might get unlimited delay and get coalesced during that time based on the how much the javascript code keeps the code busy. Does this at least make the intention for this event clear and how it is different from pointermove?

Now regarding how to spec it, the problem is that rAF related stuff is not fully specced yet and I believe I need to change HTML spec for this and reference https://html.spec.whatwg.org/#event-loop-processing-model here.

smaug----: I understand what you try to spec, it is just a tad hard to spec when pointermove isn't spec'ed to work exactly how it works in Blink (nor in Gecko). Pointerevents spec says "A user agent MUST fire a pointer event named pointermove when a pointer changes coordinates." That actually kind of prevents firing only at rAF time. So I guess we should fix that too.

NavidZ: I tried to explain it better and also added a new pointermove description. Let me know what you think.

patrickhlauke:
@smaug---- there's no rush on this as it targets the extension/possible v3 stuff, but...is this PR now looking better? can it be merged?

NavidZ:
@smaug---- let me know if the latest of this change captures what you had in mind.

extension.html
```diff
@@ -55,6 +55,37 @@ <h2>Introduction</h2>
...
       </p>
     </section>
+    <section>
+      <h2>Extensions to the Pointer Event types</h2>
+      <p>Below is the new event type that is added to the <a href="https://w3c.github.io/pointerevents/index.html#pointer-event-types">existing pointer event types</a>.</p>
+      <section>
+        <h3>The <dfn><code>pointermove</code> event</dfn></h3>
```
smaug----, line 62: I don't understand the text. "Below is the new event type..." "the pointermove". pointermove sure isn't a new type.

extension.html
```diff
@@ -55,6 +55,37 @@ <h2>Introduction</h2>
...
+        <p>A user agent MUST <a href="#firing-events-using-the-pointerevent-interface">fire a pointer event</a> named <code>pointermove</code> when a pointer changes button state.
+        Additionally one <code>pointermove</code> MUST be fired when pointer changes coordinates, pressure, tangential pressure, tilt, twist, or
+        contact geometry (e.g. <code>width</code> and <code>height</code>) and the circumstances produce no other pointer events defined in this specification.
+        This <code>pointermove</code> that is not caused by button state change MUST be fired only once right before firing an event caused by button state change or
+        <a href="https://html.spec.whatwg.org/#event-loop-processing-model">animation frame callbacks</a>. If there are mutliple changes to these attributes
```
smaug----: mutliple

extension.html
```diff
@@ -55,6 +55,37 @@ <h2>Introduction</h2>
...
+        the <code>pointermove</code> events before reaching animation frame callback or a button state change event,
+        these events will get coalesced and will be exposed via <a data-lt="getCoalescedEvents">getCoalescedEvents</a> API for the single dispatched <code>pointermove</code> event.
+        The final coordinates of such events should be used for finding the target of the event.
+      </section>
+      <section>
+        <h3>The <dfn><code>pointerrawmove</code> event</dfn></h3>
+        <p>A user agent MUST <a href="https://w3c.github.io/pointerevents/index.html#firing-events-using-the-pointerevent-interface">fire a pointer event</a>
+           named <code>pointerrawmove</code> when a pointing device attribute
+           (i.e. button state, coordinates, pressure, tangential pressure, tilt, twist, or contact geometry) is changed.
+           As oppose to <code>pointermove</code> which could be delayed until an animation frame for changes in these attributes,
```
smaug----: this talks about "could", yet the text above uses "MUST be fired only once"

extension.html
```diff
@@ -55,6 +55,37 @@ <h2>Introduction</h2>
...
+           <code>pointerrawmove</code> will get fired as soon as an attribute is changed
+           and as frequent as the javascript can handle the events.
+           Because of that the target of <code>pointerrawmove</code> events might be different from the <code>pointermove</code> events
+           due to the fact that <code>pointermove</code> events are being coalesced and the final position of the event
+           which is used for finding the target could be different from its coalesced events.
```
smaug----: It would be good to explain how coalesced events are pointerrawmove differ. For example, other than .target are the properties the same?

extension.html
```diff
@@ -90,7 +121,7 @@ <h2>Extensions to the <code>PointerEvent</code> interface</h2>
 The dispatched event's attributes will be initalized in a way that is best representative of all the coalesced events.
 For example its <a href="https://www.w3.org/TR/pointerlock/#widl-MouseEvent-movementX">movementX</a> and <a href="https://www.w3.org/TR/pointerlock/#widl-MouseEvent-movementY">movementY</a> ([[!POINTERLOCK]]) COULD be the sum of those of all the coalesced events.
 None of the events in the sequence will have (nested) coalesced events, so <code>getCoalescedEvents</code> returns an empty sequence for them.
-This API always returns at least one coalesced event for <code>pointermove</code> events and an empty list for other types of <code>PointerEvents</code>.</p>
+This API always returns at least one coalesced event for <code>pointermove</code> and <code>pointerrawmove</code> events and an empty list for other types of <code>PointerEvents</code>.</p>
```
smaug----: Why we need this for pointerrawmove?

extension.html
```diff
@@ -114,7 +145,7 @@ <h2>Extensions to the <code>PointerEvent</code> interface</h2>
 
     <section>
       <h2>Timing of coalesced event firing</h2>
-      <p>Coalescing events only happens for <code>pointermove</code> events. All the coalesced events should be fired before the next frame or before any <code>pointerup</code>, <code>pointerdown</code>, <code>pointercancel</code> events. Note that user agents don't need to hit-test all the coalesced events as none of them will have null targets.</p>
+      <p>Coalescing events only happens for <code>pointermove</code> events. All the coalesced events should be fired before the next frame or before any <code>pointerup</code>, <code>pointerdown</code>, <code>pointercancel</code> events. Note that user agents don't need to hit-test all the coalesced events as all of them will have the same target as the dispatched event.</p>
```
smaug----: ... especially when this talks about coalescing happening for pointermove events only.

NavidZ: Fair. I forgot to update this. Particularly I already talked about the dispatch time of these coalesced events before. I removed that sentence from this paragraph.

extension.html
NavidZ: I guess the same question would also hold for a pointermove that only has one coalesced event. Because in that case pointermove itself is that coalesced events. I believe the answer is that both pointerrawmove and pointermove may get coalesced. So in javascript code it is easier to always handle coalesced events of these events (if they care about that precision) instead of a check for the size of the coalesced events and if it is 0 then use the actual event itself. That is  the only reason I can think of. What do you think? Is this way more ergonomic and easy to use (like the sample drawing code later in this doc)?

extension.html
NavidZ: done

NavidZ:
@smaug---- could you take a look at this one and see whether I addressed your concerns?

extension.html
```diff
@@ -55,6 +55,39 @@ <h2>Introduction</h2>
...
+        <p>A user agent MUST <a href="https://w3c.github.io/pointerevents/index.html#firing-events-using-the-pointerevent-interface">fire a pointer event</a>
+           named <code>pointerrawmove</code> when a pointing device attribute
+           (i.e. button state, coordinates, pressure, tangential pressure, tilt, twist, or contact geometry) is changed.
+           As oppose to <code>pointermove</code> which is delayed until an animation frame reached or button change events happen,
+           <code>pointerrawmove</code> will get fired as soon as an attribute is changed
+           and as frequent as the javascript can handle the events.
+           Because of that the <code>target</code> of <code>pointerrawmove</code> events might be different from the <code>pointermove</code> events
+           due to the fact that <code>pointermove</code> events are being coalesced and the final position of the event
+           which is used for finding the <code>target</code> could be different from its coalesced events.
+           Note that if the javascript handlers are slow, <code>pointerrawmove</code> events can also get coalesced.
```
smaug----: "Note that if the javascript handlers are slow, pointerrawmove events can also get coalesced."  what does this actually mean. What is slow JS? handler taking more than 16ms? handler forcing rAF to happen later? or what? And what is the reason for this coalescing? If the page wants raw events, it should get raw events. That would be simpler and easier to understand API.

NavidZ: Let's discussion a little more about this. Just to make sure whether we should change the design and logic around this event or just the text here.

I agree that when the page does express the desire to get the pointerrawmove events then presumably they have a fast handler that doesn't take long. But imagine a case that the app is not well written and takes say a few milliseconds to handle even a pointerrawmove and I have a 1000Hz mouse and I'm interacting with the page. So in this case user agent is sending 1000 move events to the page (assuming mouse is moving around) or 1 event every millisecond. So user agent either should drop these raw move events or queue them forever but the queue is only growing in that case (as the incoming events are faster than the handled rate). So user agent should either drop some of the rawmove events or coalesce them and expose those coalesced events with getCoalescedEvents APIs. right?

smaug----: But if events are coalesced, when does that happen? When is it ok for UA to coalesce?
Is it ok to fire just one pointerrawmove per frame and coalesce the rest. Should the spec be a bit more precise. Something like... if UA detects that rendering will be updated and there are pending pointerrawmoves, fire just one event and coalesce the rest in the queue?

NavidZ: I see. So in general you are okay with coalescing the rawmove events. The only problem is a reasonable definition around it and when to do the coalescing. Let me take another stab at the text.

NavidZ: Sorry for the delay. I pushed a new commit. I used this sentence here instead: "Note that if UA detects the thread running javascript handlers for this event is not idle, pointerrawmove events will get coalesced and they will all be delivered as coalesced events of one pointerrawmove event as soon as the thread goes idle". Is that more precise?

smaug----: Rather vague still, since "thread" isn't defined. If we want to define this properly, it should probably talk about tasks and task queues.
Also, if there are pending idleCallbacks, will those cause coalescing?

NavidZ: @smaug---- How about now?

extension.html
```diff
@@ -55,6 +55,44 @@ <h2>Introduction</h2>
...
+    <section>
+      <h2>Extensions to the Pointer Event types</h2>
+      <p>Below is a new pointer event type as well as a modification to an existing event in the <a href="https://w3c.github.io/pointerevents/index.html#pointer-event-types">existing pointer event types</a>.</p>
+      <section>
+        <h3>The <dfn><code>pointermove</code> event</dfn></h3>
+        <p>A user agent MUST <a href="#firing-events-using-the-pointerevent-interface">fire a pointer event</a> named <code>pointermove</code> when a pointer changes button state.
+        Additionally one <code>pointermove</code> MUST be fired when pointer changes coordinates, pressure, tangential pressure, tilt, twist, or
+        contact geometry (e.g. <code>width</code> and <code>height</code>) and the circumstances produce no other pointer events defined in this specification.
+        This <code>pointermove</code> that is not caused by button state change MUST be fired only once right before firing an event caused by button state change or
+        <a href="https://html.spec.whatwg.org/#event-loop-processing-model">animation frame callbacks</a>. If there are multiple changes to these attributes
```
smaug----: This doesn't seem to follow what implementations do. If there is nothing animating, *move events still fire at some point.
And I think events should fire sooner even in case there is something animating. More like before resize steps.

extension.html
```diff
@@ -55,6 +55,44 @@ <h2>Introduction</h2>
...
+        the <code>pointermove</code> events before reaching animation frame callback or a button state change event,
+        these events will get coalesced and will be exposed via <a data-lt="getCoalescedEvents">getCoalescedEvents</a> API for the single dispatched <code>pointermove</code> event.
+        The final coordinates of such events should be used for finding the target of the event.
+      </section>
+      <section>
+        <h3>The <dfn><code>pointerrawmove</code> event</dfn></h3>
+        <p>A user agent MUST <a href="https://w3c.github.io/pointerevents/index.html#firing-events-using-the-pointerevent-interface">fire a pointer event</a>
+           named <code>pointerrawmove</code> when a pointing device attribute
+           (i.e. button state, coordinates, pressure, tangential pressure, tilt, twist, or contact geometry) is changed.
+           As oppose to <code>pointermove</code> which is delayed until an animation frame reached or button change events happen,
```
smaug----: yeah, this doesn't seem to be what implementations do. (tested also Chrome)

extension.html
NavidZ: What does this have to do with whether something being animated or not? Event if nothing is animating rAF is still there even if it doesn't run any callbacks in js. Can you explain what you tested? Chrome always does delay pointermoves with the current implementation to right before rAF.

extension.html
NavidZ: Similar to the comment above. What is the scenario that you are testing that is not doing this?

extension.html
```diff
@@ -63,34 +63,36 @@ <h3>The <dfn><code>pointermove</code> event</dfn></h3>
...
         Additionally one <code>pointermove</code> MUST be fired when pointer changes coordinates, pressure, tangential pressure, tilt, twist, or
         contact geometry (e.g. <code>width</code> and <code>height</code>) and the circumstances produce no other pointer events defined in this specification.
-        This <code>pointermove</code> that is not caused by button state change MUST be fired only once right before firing an event caused by button state change or
-        <a href="https://html.spec.whatwg.org/#event-loop-processing-model">animation frame callbacks</a>. If there are multiple changes to these attributes
-        the <code>pointermove</code> events before reaching animation frame callback or a button state change event,
-        these events will get coalesced and will be exposed via <a data-lt="getCoalescedEvents">getCoalescedEvents</a> API for the single dispatched <code>pointermove</code> event.
+        These events may be coalesced or aligned to <a href="https://html.spec.whatwg.org/#event-loop-processing-model">animation frame callbacks</a> basesd on UA decision.
```
smaug----: basesd

extension.html
```diff
@@ -63,34 +63,36 @@ <h3>The <dfn><code>pointermove</code> event</dfn></h3>
...
+        The coalesdced events information will be exposed via <a data-lt="getCoalescedEvents">getCoalescedEvents</a> API for the single dispatched <code>pointermove</code> event.
```
smaug----: coalesdced

NavidZ: done

extension.html
NavidZ: done

smaug---- (approved):

PR #261, opened 2018-08-22
pointerenter and pointerleave should not be composed

smaug----:
Not quite sure about reasonable text here. The spec should probably have similar table as UIEvents.
https://github.com/w3c/uievents/pull/210


NavidZ:
Yeah. We used to have the composed in the table above. But since it was true for all we put it in the text. But now that they are different it might be nice to put them back into the table. I'm also fine with ui events style of having one table for each event.

NavidZ:
Looks good to me.

PR #262, opened 2018-08-23
Add informative note about tangential pressure and Windows

patrickhlauke:
Based on email discussions with Scott Low/Matt Rakow (Microsoft) and @NavidZ , tangential pressure is not exposed on Windows, but works correctly on MacOS (when using a Wacom external digitizer). This leaves tangential pressure in the spec, but notes that on Windows it's not working. @plehegar I assume this is acceptable for a spec?

### DIFF
--- a/index.html
+++ b/index.html
@@ -301,6 +301,7 @@ interface PointerEvent : MouseEvent {
                     <dt><dfn>tangentialPressure</dfn></dt>
                         <dd>
                           <p>The normalized tangential pressure (also known as barrel pressure), typically set by an additional control (e.g. a finger wheel on an airbrush stylus), of the pointer input in the range of [-1,1], where 0 is the neutral position of the control. Note that some hardware may only support positive values in the range of [0,1]. For hardware that does not support tangential pressure, the value MUST be 0.</p>
+                          <div class="note">Information about tangential pressure is currently not exposed by the operating system's input APIs on Windows.</div>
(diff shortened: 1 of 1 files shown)


NavidZ:
LGTM

patrickhlauke:
Comment from Matt Rakow (Microsoft) on mailing list http://lists.w3.org/Archives/Public/public-pointer-events/2018JulSep/0029.html

> Since the spec already defines behavior in the case that tangential pressure is not supported, and (I think) no one is signing up to do comprehensive testing across various OS's and versions to drop those details in the note, I'm not sure the note is helpful.  At minimum you'll need to indicate the version of Windows you're referring to and keep it updated over time if support is added later.
>
> Perhaps instead of the note, consider updating the normative statement to be more explicit about this case, e.g.
>
> "For hardware that does not support tangential pressure" ->
> "For hardware and platforms that do not support tangential pressure"

Actually, agree with Matt's alternative take here. I'll see if this can be worded a bit more "evergreen" without also requiring testing/calling out platforms.

plehegar:
LGTM

plehegar (approved):

patrickhlauke:
Abandoned this PR in favour of the tiny editorial clarification here https://github.com/patrickhlauke/pointerevents/commit/14353f37388e2577e69e3a58e853a262d016f96a

PR #264, opened 2018-09-07
Fix respec errors, minor editorial changes "for devices"/"for hardware", add Navid Zolghadr as co-editor

patrickhlauke:


patrickhlauke:
apologies, i completely missed that i only pushed the editorial and respec changes from a while ago to my OWN fork, rather than the official repo. this is a catch-up PR now to address this, as well as adding Navid as co-editor.

PR #267, opened 2018-10-02
Use connected keyword instead of removing a node

### LINKED ISSUE #259: "When the pointer capture target override is removed from its ownerDocument's tree" should talk about target not being connected anymore
I think we do want this work in Shadow DOM

NavidZ:
fixes #259


NavidZ:
@smaug----  does this address your concern?

smaug----:
I guess we can say it that way too. There is of course the issue that when lostpointercapture is supposed to be dispatched.

NavidZ:
Immediately? :)
I see what you are seeing. But so far in V2 we never had a pattern of referencing and hooking into event loop steps. I feel like we may need to update the whole spec to hook to those steps everywhere and not only here. Maybe in V3?

smaug----:
What does immediately mean? At which point of which algorithm? Before or after CEReactions?
I wonder what implementations do.

smaug----:
But ok, I guess this is fine. We have similar issues in the platform when dealing with focus/blur events etc.

patrickhlauke:
Thank you

PR #270, opened 2018-11-27
Pointer Capture is normative. Fix #269

### LINKED ISSUE #269: Pointer Capture section marked as non-normative
https://w3c.github.io/pointerevents/#pointer-capture (must be an error from `<section>` rejiggling)

plehegar:
This section was incorrectly marked informative. Fixes #269

### DIFF
--- a/index.html
+++ b/index.html
@@ -807,12 +807,14 @@ partial interface Navigator {
     </section>
     <section>
         <h1><dfn>Pointer Capture</dfn></h1>
+        <section class='informative'>
+          <h2>Introduction</h2>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
digging way back into https://github.com/w3c/pointerevents/commit/a9bbc3cc5ff4d28443573aed9b265dfd73a50441 it seems originally the introduction to that section and the figure were informative.

```
<h1><dfn>Pointer Capture</dfn></h1>	
    <div class="informative">	
        <p>Pointer capture allows ...</p>	
        <figure id="figure_slider">	
            ...
        </figure>	
    </div>	
    	
    <section>	
        <h2>Setting Pointer Capture</h2>
        ....
```

Wonder if we should go for something similar again, or if we're ok just removing it and going with what you propose here.

patrickhlauke:
Thinking about this some more (and looking at similar sections), it's probably fine to go with your proposed change here. There doesn't seem to be any harm in having the intro and examples there not explicitly marked non-normative... @NavidZ @mbrubeck et al, shout if you disagree.

plehegar:
I added a commit to make the intro to that section informative. It requires a dedicated section but makes it clear.

NavidZ:
Looks good to me.

patrickhlauke (approved):

PR #280, opened 2019-04-25
Clarify coalesce events and predicted event's target 

EiraGe:
As the suggestion from tag review, changing the text for `target`

### DIFF
--- a/extension.html
+++ b/extension.html
@@ -154,8 +154,7 @@ So the first event will have the smallest <code>timeStamp</code> which should st
               </dd>
 
               <p>
-              A <a href="pointerevents2#pointerevent-interface">PointerEvent</a>'s <a href="https://dom.spec.whatwg.org/#get-target"><code> get target </code></a> algorithm, returns the associate dispatched event's target if it is a non-dispatch event (not being dispatched to a target, i.e. coalesced events
-                and predicted events); otherwise returns the object that the event is targeted using <a href="https://www.w3.org/TR/uievents/#event-flow">Event dispatch and DOM event flow</a>.
+              Coalesced events and predicted events' <a href="https://dom.spec.whatwg.org/#get-target"><code> get target </code></a> algorithm,  if associated dispatched event is not null, return the associated event's <a href="http://dom.spec.whatwg.org/#event-target"><code>target</code></a>; otherwise, returns this event's target.
               </p>
           </dl>
       </div>
--- a/extension.html
+++ b/extension.html
@@ -183,8 +183,8 @@ partial interface PointerEvent {
             boolean. When an event is created the attribute must be initialized to false.</p>
             <p> A <a> PointerEvent</a> has an associated <dfn> predicted events targets dirty</dfn>
             boolean. When an event is created the attribute must be initialized to false.</p>
-            <p> When <a>PointerEvent</a>'s target change, set both <code> coalesced event targets
-              dirty</code> and <code> predicted events targets dirty</code> to true.</p>
+            <p> When <a>PointerEvent</a>'s target is changed, set both <code> coalesced event
+              targets dirty</code> and <code> predicted events targets dirty</code> to true.</p>
 
 
               <dt><dfn>getCoalescedEvents</dfn></dt>
--- a/extension.html
+++ b/extension.html
@@ -152,6 +152,11 @@ The rest of the attributes might be different from the dispatched event based on
 The events in the sequence will have increasing <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a> ([[!WHATWG-DOM]]).
 So the first event will have the smallest <code>timeStamp</code> which should still be greater than the last event in the coalesced events.</p>
               </dd>
+
+              <p>
+                A <code>PointerEvent</code>'s <code> get target </code> algorithm, returns the associate dispatched event's target if it is a non-dispatch event (not being dispatched to a target, i.e. coalesced events
+                and predicted events); otherwise returns the object that the event is targeted using <a href="https://www.w3.org/TR/uievents/#event-flow">Event dispatch and DOM event flow</a>.
+              </p>
           </dl>
       </div>
     </section>
--- a/extension.html
+++ b/extension.html
@@ -154,7 +154,7 @@ So the first event will have the smallest <code>timeStamp</code> which should st
               </dd>
 
               <p>
-                A <code>PointerEvent</code>'s <code> get target </code> algorithm, returns the associate dispatched event's target if it is a non-dispatch event (not being dispatched to a target, i.e. coalesced events
+              A <a href="pointerevents2#pointerevent-interface">PointerEvent</a>'s <a href="https://dom.spec.whatwg.org/#get-target"><code> get target </code></a> algorithm, returns the associate dispatched event's target if it is a non-dispatch event (not being dispatched to a target, i.e. coalesced events
                 and predicted events); otherwise returns the object that the event is targeted using <a href="https://www.w3.org/TR/uievents/#event-flow">Event dispatch and DOM event flow</a>.
               </p>
           </dl>
--- a/extension.html
+++ b/extension.html
@@ -117,45 +117,74 @@ partial interface PointerEvent {
 };
           </pre>
           <dl data-dfn-for="PointerEvent" data-link-for="PointerEvent">
+            <p>A <a>PointerEvent</a> has an associated <dfn> coalesced event list</dfn> (a list of zero or more <code>PointerEvents</code>).
+            If this event is a <code>pointermove</code> or <code>pointerrawupdate</code> event, it is a sequence of all <code> PointerEvents</code> that were coasesced into this event;
+            otherwise it is the empty list. </p>
+            <p> When a <var>PointerEvent</var> is created, run the following steps for each <var>event</var> in the <a> coalesced event list</a>:
+              <ol>
+                <li><p>Set <var>event</var>'s <code>pointerId</code>, <code> pointerType</code>, <code>isPrimary</code> and <code>isTrusted</code> to
+                  the <var>PointerEvent</var>'s <code>pointerId</code>, <code> pointerType</code>, <code>isPrimary</code> and <code>isTrusted</code>.</li>
+                <li><p>Set <var>event</var>'s <code>cancelable</code> and <code>bubbles</code> attributes to false.</li>
+                <li><p>Set <var>event</var>'s <a>coalesced event list</a> and <a> predicted event list</a> to empty list.</li>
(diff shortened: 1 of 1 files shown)


extension.html
```diff
@@ -141,17 +136,21 @@ <h2>Extensions to the <code>PointerEvent</code> interface</h2>
...
+Similar to the coalesced events, the following attributes of the predicted events will always have the same value as the dispatched event:
+<code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>, <code>isTrusted</code>.
+<code>cancelable</code> and <code>bubbles</code> attributes will be false.
 The rest of the attributes might be different from the dispatched event based on the prediction algorithm in user agent.
 The events in the sequence will have increasing <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a> ([[!WHATWG-DOM]]).
 So the first event will have the smallest <code>timeStamp</code> which should still be greater than the last event in the coalesced events.</p>
               </dd>
+
+              <p>
+                The <code> target </code> attribute's getter of both coalesced events and predicted events, when invoked, must return the associate dispatched event's target.
```
NavidZ: I'm not sure if this addresses @annevk concerns. Although you did change the text to talk about the target's getter and hence that might be good enough.

But let's ask @annevk. Does this look okay or do you also want a note in https://dom.spec.whatwg.org/#dom-event-target mentioning this spec might override the getter?

annevk: I think it needs to be more than a note, it needs to be part of the model, right? You want to define some kind of extensibility hook that you use here.

NavidZ: When you say an "extensibility hook", do you not consider the Event.target as a hook that can be extended? Like what this [PR](https://github.com/EiraGe/dom/pull/1) refers to. Do you want like an explicit function like getTarget that is returning the value of target attribute and that can be overridden here instead or something like that?

annevk: Yeah, `Event.prototype.target` is a public API you cannot really touch. You need to add some kind of internal-facing operation you can override in other specifications, which also makes it clear from the DOM Standard that this is a thing some specifications do (it should probably have a note pointing them out, only this one so far).

EiraGe: I updated this and also the other [PR](https://github.com/EiraGe/dom/pull/1) for DOM. Is this clearer now?

extension.html
```diff
@@ -152,6 +152,11 @@ <h2>Extensions to the <code>PointerEvent</code> interface</h2>
 The events in the sequence will have increasing <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a> ([[!WHATWG-DOM]]).
 So the first event will have the smallest <code>timeStamp</code> which should still be greater than the last event in the coalesced events.</p>
               </dd>
+
+              <p>
+                A <code>PointerEvent</code>'s <code> get target </code> algorithm, returns the associate dispatched event's target if it is a non-dispatch event (not being dispatched to a target, i.e. coalesced events
```
annevk: This should link to some kind of associated event slot right? And presumably you'd then be able to branch on that for the return value, without having to be hand-wavy about it.

EiraGe: I added the link to PointerEvent interface (@NavidZ could you please help me take a look if I did it correctly)
I also added the link to "dom#get-target" (which wasn't exist yet). I'll create the PR to dom once this approach get approved and try to merge these two at the same time.

annevk:
I think I'd like to take a step back and ask for some more context, since I'm still not quite sure I understand coalesced events enough. Are they truly standalone objects or do they also have some memory of the event path?

What here for instance protects shadow root nodes from being exposed?

NavidZ:
@annevk what do you mean the memory of the event path?
This all is more or less explained in this draft as well. But just to reiterate. They will have the coordinates of the event path and they go back to the last event that was dispatched to js. So for example if js is keeping the main thread busy they wouldn't receive any event until they are done running the js code. Then they will get one move event and all the other move events that have happened  during this busy time as coalesced events of that dispatched event. Note that the coalesced events aren't hit-tested by themselves (maybe this is the confusion). So it is not like that they will have their own targets. They just return the same target as the dispatched event. So when the dispatched target is protected against shadow DOM through normal dispatching algorithms steps these coalesced events will get that protection automatically. Is that more clear?

annevk:
Thanks! So given an event X, X's coalesced events will override their own target with X's target. So the get target implementation for a coalesced event is something like:

1. If **associated event** is non-null, then return **associated event**'s target.
2. Return this's target.

(With the various terms cross-referenced, e.g., [target](https://dom.spec.whatwg.org/#event-target).)

If so, that would indeed preserve shadow tree encapsulation invariants I think as far as targets go, though it's unclear to me how any relative coordinates would work in this setup.

NavidZ:
They will go with their own definition with no further modification. For example offsetX/Y says they will be calculated based on the location of the target of the event. The target comes form that definition in this spec and the offsetX/Ys will be calculated based on that target (i.e. associated event target).

annevk:
Ah yeah, they're computed from absolute values, despite that not really working with how events work specification-wise. Meh, so much to fix.

NavidZ:
@annevk did you want further changes on this change?

annevk:
Yeah, I think this should be written more like https://github.com/w3c/pointerevents/pull/280#issuecomment-493519005 and shouldn't require any references to the UI Events specification as these concepts are defined in the DOM Standard.

EiraGe:
I've updated this as suggest in [#280 (comment)](https://github.com/w3c/pointerevents/pull/280#issuecomment-493519005), please take another look :)

annevk (commented):
This is better, but there are still some aspects lacking. What are coalesced and predicted events? Instead of saying that, you want to define how this works for their event interface implementations.

In addition, for the affected event interfaces, you need to define the "associated dispatched event" field (so you can link it from here). And presumably you need to set that field when dispatching an event that has events associated with it.

NavidZ:
> In addition, for the affected event interfaces, you need to define the "associated dispatched event" field (so you can link it from here). And presumably you need to set that field when dispatching an event that has events associated with it.

@annevk we wanted to avoid having another idl definition for these events. That was intentional from the beginning so that developers could just pass these events to their handlers and it would still work despite checking for types and whatnot. Additionally there was no plan to expose that "associated dispatched event" in the idl to the javascript at this point in time. My understanding is that if we define that idl technically js will have access to that field and also can type check the coalesced/predicted events. Right?

annevk:
No, you don't change the IDL, but you change the definition of their implementation, including adding the appropriate internal slots. I'm talking about the "private" spec-level API, not the JavaScript API. I think @domenic might have given a Google-internal talk on this.

extension.html
```diff
@@ -117,40 +117,101 @@ <h2>Extensions to the <code>PointerEvent</code> interface</h2>
...
+            <p>A <a>PointerEvent</a> has an associated <dfn> coalesced event list</dfn> (a list of
+            zero or more <code>PointerEvents</code>). If this event is a <code>pointermove</code>
+            or <code>pointerrawupdate</code> event, it is a sequence of all <code> PointerEvent
+            </code> that were coasesced into this event; otherwise it is the empty list.</p>
+            <p>The events in the <code >coalesced event list </code> will have increasing
+            <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a>
+            ([[!WHATWG-DOM]]), so the first event will have the smallest <code>timeStamp</code>.
+            </p>
+
+            <p> When a <var>PointerEvent</var> is created, run the following steps for each
```
NavidZ: I wonder if you want to move this right after the definition of predicted event list as well.

extension.html
```diff
@@ -117,40 +117,101 @@ <h2>Extensions to the <code>PointerEvent</code> interface</h2>
...
+                <li><p>Set <var>event</var>'s <a>coalesced event list</a> and <a> predicted event
+                    list</a> to empty list.</li>
+                <li><p> Initialize the rest of the attributes the same way as <a>PointerEvent</a>.
+              </ol>
+            </p>
+            </p>
+
+            <p> A <a> PointerEvent</a> has an associated <dfn> coalesced events targets dirty</dfn>
+            boolean. When an event is created the attribute must be initialized to false.</p>
+            <p> A <a> PointerEvent</a> has an associated <dfn> predicted events targets dirty</dfn>
```
NavidZ: You should be able to merge these paragraphs into one and just say ..." has associated coalesced events targets dirty and predicted events targets dirty bits." Also I believe you need to remove the space inside <dfn> as that will break the linking in respec.

NavidZ (commented):
I love this way much better. It has very well defined what Blink does and it leaves no vague definition around. I believe we can improve the initialization part of it slightly better when we merge this into the pointerevent spec itself as all the initialization could be in one place.

Thanks @anne and @domenic for the suggestions.

extension.html
EiraGe: Sounds right, I merge this with the predicted event list one. They are the same.

extension.html
EiraGe: Done merge into 1 paragraph and space removed.

NavidZ (approved):
It looks good to me.
@annevk can you also take a look and see whether it addresses your concern. This is a complete rewrite to make sure we captured everything and all the details.

annevk (commented):
Thanks, this is an improvement and makes the data structures clearer, but there's not much normative requirements that go towards filling them.

I'd expect the coalescing to be defined and the creation of, e.g., pointerrawupdate, to make it clear how that ends up filling the two private slots.

NavidZ:
@annevk That part was intentional on the spec. Right now FF and Chromium do different algorithms and heuristics for coalescing. We wanted to first try and expose those events using this API. Then we can iterate later on different spec to land on a standard behavior on what and how to coalesce events. In that sense we wanted to leave this to the UA behavior for the current version. There is a normative text that says the logic for coalescing and what not is "based on UA decision".

annevk:
While it can sometimes be reasonable to mark certain bits "implementation-defined", I still think it would be useful to have a more coherent description of how the various pieces fit together. The current description might lead someone to think that if they were to create a synthetic event these fields would be non-empty, for instance.

NavidZ:
@annevk Note that we did provide an example of about aligning to animation frame callback. But again didn't want to explain too much on that as it is still up for debate. Regarding the constructor we can add a sentence about the fact that only the events that are dispatched by the UA has that property and otherwise the value of these lists (i.e. coalescedevents and predictedevents) come from the init dictionary that is already defined. But I don't see this pattern in the existing PE v2 or ui events either for this matter. As it seems clear in the those spec that what you pass in the init dictionary will be set to the generated event in all cases. Still let me know if you want us to add that here.

NavidZ:
@annevk is the current state alright?

NavidZ:
This pull request is a strict improvement to what we have and I believe it address all the major concerns in the TAG review. @annevk feel free to file any issue against this repo if you believe something is missing.

PR #281, opened 2019-05-10
Add better intro to the extension doc

NavidZ:
Adding introduction about getPredictedEvents and pointerrawupdate to the extension doc


extension.html
```diff
@@ -52,7 +52,29 @@ <h2>Introduction</h2>
...
+	The <code>getCoalescedEvents</code> API (similar to Android <a href="https://developer.android.com/reference/android/view/MotionEvent.html#getHistoricalX(int, int)">getHistorical</a> APIs
+        and iOS <a href="https://developer.xamarin.com/api/member/MonoTouch.UIKit.UIEvent.GetCoalescedTouches/p/MonoTouch.UIKit.UITouch/">GetCoalescedTouches</a> API) exposes all the events
+        that were coalesced into a single event.
+        Coalescing events into a single event enables user agents to prevent processing of the events that don’t get a chance to update the frame content visually.
```
dtapuska: Probably should be worded as:
User agents may coalesce a stream of events into a single event sent to javascript. This situation can occur when many events are generated by the hardware in between the requestAnimationFrame boundaries or for long running javascript code that misses the frame boundary.

extension.html
```diff
@@ -52,7 +52,29 @@ <h2>Introduction</h2>
...
+        Basically if the javascript handlers process the events and update the screen contents faster than the screen refresh rate only the last update will be rendered and shown in the next frame.
+        So for example instead of sending a few <code>pointermove</code> events in the same frame UA can choose to coalesce all the move events happened in a frame into one event and send only
+        that event in a frame representing the last state of the pointer and javascript will react to that.
+        This would work fine with most of the applications and would save a lot of redundant processing on the web.
```
dtapuska: Most applications care only about the last cooridinates and not the intermediate coordinates that occured. However, a class of applications such as drawing applications may wish precise (subframe timing) coordinates in order to draw better detailed lines.

extension.html
```diff
@@ -55,13 +55,13 @@ <h2>Introduction</h2>
...
         So for example instead of sending a few <code>pointermove</code> events in the same frame UA can choose to coalesce all the move events happened in a frame into one event and send only
-        that event in a frame representing the last state of the pointer and javascript will react to that.
-        This would work fine with most of the applications and would save a lot of redundant processing on the web.
-        However, some applications like drawing apps may want to take advantage of the precise history of a pointer within a frame time as user moves the pointer on the screen to be able to draw smoother and better looking curves.
-        Note that they still draw at the end of each frame but they care to know the history of the pointer and not only the latest state.
+        that event before requestAnimationFrame representing the last state of the pointer and javascript will react to that.
+        Most applications care only about the last coordinates and not the intermediate coordinates that occured.
```
dtapuska: occurred

extension.html
```diff
@@ -55,13 +55,13 @@ <h2>Introduction</h2>
...
         and iOS <a href="https://developer.xamarin.com/api/member/MonoTouch.UIKit.UIEvent.GetCoalescedTouches/p/MonoTouch.UIKit.UITouch/">GetCoalescedTouches</a> API) exposes all the events
         that were coalesced into a single event.
-        Coalescing events into a single event enables user agents to prevent processing of the events that don’t get a chance to update the frame content visually.
+        User agents may coalesce a stream of events into a single event sent to javascript.
+        This situation can occur when many events are generated by the hardware in between the requestAnimationFrame boundaries or for long running javascript code that misses the frame boundary.
         Basically if the javascript handlers process the events and update the screen contents faster than the screen refresh rate only the last update will be rendered and shown in the next frame.
         So for example instead of sending a few <code>pointermove</code> events in the same frame UA can choose to coalesce all the move events happened in a frame into one event and send only
```
dtapuska: So for example,

extension.html
```diff
@@ -55,13 +55,13 @@ <h2>Introduction</h2>
 	The <code>getCoalescedEvents</code> API (similar to Android <a href="https://developer.android.com/reference/android/view/MotionEvent.html#getHistoricalX(int, int)">getHistorical</a> APIs
         and iOS <a href="https://developer.xamarin.com/api/member/MonoTouch.UIKit.UIEvent.GetCoalescedTouches/p/MonoTouch.UIKit.UITouch/">GetCoalescedTouches</a> API) exposes all the events
         that were coalesced into a single event.
-        Coalescing events into a single event enables user agents to prevent processing of the events that don’t get a chance to update the frame content visually.
+        User agents may coalesce a stream of events into a single event sent to javascript.
+        This situation can occur when many events are generated by the hardware in between the requestAnimationFrame boundaries or for long running javascript code that misses the frame boundary.
         Basically if the javascript handlers process the events and update the screen contents faster than the screen refresh rate only the last update will be rendered and shown in the next frame.
```
dtapuska: Basically, if the javascript handlers process the events and update the screen contents faster than the screen refresh rate only the last update will render and show in the next frame.

extension.html
dtapuska: "same frame UA can choose" sounds odd

NavidZ: How about now?

dtapuska (approved):

PR #300, opened 2019-08-23
Define active document for the active pointers

### LINKED ISSUE #291: Setting pointer capture on an element in a parent frame when the pointer was made active in a child frame
The test [pointerevent_mouse_pointercapture_inactivate_pointer.html](https://github.com/web-platform-tests/wpt/blob/master/pointerevents/pointerevent_mouse_pointercapture_inactivate_pointer.html) specifically checks that handling a `pointerdown` event dispatched on a target element in a child frame and calling `setPointerCapture()` in the parent frame on an element contained in the parent frame should *not* trigger pointer capture. 

However, I'm not finding explicit text in the spec that would specify this behavior. The only step in [Setting Pointer Capture](https://w3c.github.io/pointerevents/#setting-pointer-capture) that should fails silently is:

> 4. If the pointer is not in the active buttons state, then terminate these steps.

But I can't see anything that specifies that, in this scenario, the pointer would not be in the active buttons state.

As it stands on [wpt.fyi](https://wpt.fyi/results/pointerevents/pointerevent_mouse_pointercapture_inactivate_pointer.html?label=master&label=experimental), I see Chrome and Edge pass this test, but Firefox and Safari do not.

NavidZ:
Limit the pointer capture API scope to only work
when it is called on the document who got the
last event in the given pointer stream.

Fixes #291

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Nov 4, 2019, 9:35 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fpointerevents%2F506fdd165b523f5b2cf02402ffa63812cce3e10d%2Findex.html%3FisPreview%3Dtrue)

```
[33m📡 HTTP Error 520:[39m [36mhttps://rawcdn.githack.com/w3c/pointerevents/506fdd165b523f5b2cf02402ffa63812cce3e10d/index.html[39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/pointerevents%23300.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -201,6 +201,8 @@ eventTarget.dispatchEvent(event);</code>
         <dl>
             <dt><dfn>active buttons state</dfn></dt>
                 <dd>The condition when a pointer has a non-zero value for the <code>buttons</code> property. For mouse, this is when the device has at least one button depressed. For touch, this is when there is physical contact with the digitizer. For pen, this is when either the pen has physical contact with the digitizer, or at least one button is depressed while hovering.</dd>
+            <dt><dfn>active document</dfn></dt>
+                <dd>For every <a>active pointer</a>, the document that received the last event from that pointer.</dd>
             <dt><dfn>active pointer</dfn></dt>
                 <dd>Any touch contact, pen stylus, mouse cursor, or other pointer that can produce events.  If it is possible for a given pointer (identified by a unique <code>pointerId</code>) to produce additional events within the document, then that pointer is still considered active. Examples:
                     <ul>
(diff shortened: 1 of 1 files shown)


NavidZ:
@smaug---- @graouts
I feel like it might be too wordy. But I guess PointerEvent spec is not that algorithmic. I tried to spec this one more algorithmic.

Just something to remember regarding an embedded iframe. When the pointer moves on top of an embedded iframe the embedder will receive one 'pointerover' event on the iframe element and never receives a 'pointerleave' while the pointer is on top of the embedded iframe. That's why I didn't rely on the boundary events in this pull request.

index.html
```diff
@@ -391,17 +393,24 @@ <h2>The <dfn>Primary Pointer</dfn></h2>
...
-                    <li>Otherwise, set the target to the object returned by normal hit test mechanisms (out of scope for this specification).</li>
+                    <li>If the <a>pointer capture target override</a> has been set for the pointer, set the <var>eventTarget</var> to <a>pointer capture target override</a> object.</li>
+                    <li>Otherwise, set the <var>eventTarget</var> to the object returned by normal hit test mechanisms (out of scope for this specification).</li>
                 </ul>
 
-                <p>If this is a <code>pointerdown</code> event, the associated device is a direct manipulation device and the target is an <code>Element</code>, then <a href="#setting-pointer-capture">set pointer capture</a> for this <code>pointerId</code> to the target element as described in <a>implicit pointer capture</a>.
+                <p>If the <var>eventTarget</var> is an <code>Element</code>, then let <var>targetDocument</var> be <var>eventTarget</var>'s <a href="https://dom.spec.whatwg.org/#concept-node-document">node document</a> [[!DOM4]].
+                   Otherwise, <var>eventTarget</var> is a <code>Window</code> object,
```
smaug----: When can the eventTarget be Window?

index.html
```diff
@@ -391,17 +393,24 @@ <h2>The <dfn>Primary Pointer</dfn></h2>
             <section>
                 <h2>Firing events using the <code>PointerEvent</code> interface</h2>
                 <p>To <dfn>fire a pointer event named e</dfn> means to <dfn>fire an event named e</dfn> as defined in [[!DOM4]] with an event using the <a>PointerEvent</a> interface whose attributes are set as defined in <a href="#pointerevent-interface"><code>PointerEvent</code> Interface</a> and <a href="#attributes-and-default-actions">Attributes and Default Actions</a>.</p>
-                <p>If the event is not <code>gotpointercapture</code> or <code>lostpointercapture</code>, run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps for this <code>PointerEvent</code>.
+                <p>If the <var>event</var>'s <code>type</code> is not <code>gotpointercapture</code> or <code>lostpointercapture</code>, run <a href="#process-pending-pointer-capture">Process Pending Pointer Capture</a> steps for this <code>PointerEvent</code>.
```
graouts: Sorry to be a stickler, but this change, while welcome, is unrelated to issue #291.

index.html
```diff
@@ -391,17 +393,24 @@ <h2>The <dfn>Primary Pointer</dfn></h2>
...
 
-                <p>The target object at which the event is fired is determined as follows:
+                <p>The <var>eventTarget</var> object at which the <var>event</var> is fired is determined as follows:
```
graouts: Same for this change, this should be in a separate PR.

index.html
```diff
@@ -391,17 +393,24 @@ <h2>The <dfn>Primary Pointer</dfn></h2>
...
                 <ul>
-                    <li>If the <a>pointer capture target override</a> has been set for the pointer, set the target to <a>pointer capture target override</a> object.</li>
-                    <li>Otherwise, set the target to the object returned by normal hit test mechanisms (out of scope for this specification).</li>
+                    <li>If the <a>pointer capture target override</a> has been set for the pointer, set the <var>eventTarget</var> to <a>pointer capture target override</a> object.</li>
+                    <li>Otherwise, set the <var>eventTarget</var> to the object returned by normal hit test mechanisms (out of scope for this specification).</li>
```
graouts: Same for this change.

index.html
graouts: That's a good question. I don't know that this could ever happen due to pointer capture.

graouts (commented):
The question about whether an event may be targeted at a `Window` object is a good one. I think once it is answered and the drive-by changes are moved to a dedicated PR, this will be good. I like the new text.

index.html
NavidZ: I have been thinking about this and couldn't come up with any case. As a matter of fact Chrome code certainly never targets the window directly. Note that Window will be in the propagation path but never is the target if the event.
I have added the window case for the case of completion as it was different to get the document from a window object. But I'll remove it  to be more precise as we know it doesn't happen at least with the current state of the things.

NavidZ: Also pointer capture cannot cause that. As pointer capture API is from the Element idl and window doesn't inherit from that in anyway.

graouts (approved):

smaug---- (approved):

mustaqahmed:
We need a WPT that requests pointer-capture on an element from a different frame.

smaug----:
https://searchfox.org/mozilla-central/source/testing/web-platform/tests/pointerevents/pointerevent_pointercapture_in_frame.html is possibly related test

mustaqahmed:
We now have WPT coverage for this: [pointerevent_pointercapture_in_frame.html](https://wpt.fyi/results/pointerevents?label=master&label=experimental&aligned&q=pointerevent_pointercapture_in_frame.html).

PR #316, opened 2020-02-17
add altitudeAngle/azimuthAngle

### LINKED ISSUE #274: tiltX and tiltY aren't as helpful as just having altitude angle
Pointer Events for a stylus currently use `tiltX`, `tiltY` and `twist` to give the orientation of the stylus in space. I believe that just `twist` and `altitude` would be sufficient for most cases, and easier to understand.

`altitude` is the angle between the screen surface and the stylus, ignoring the rotation or twist. Many styluses don't have an "up" direction around the barrel (e.g. Apple Pencil), so spinning the stylus in your fingers shouldn't change any values. Most drawing applications therefore just use the angle between the screen and the stylus to determine the angle of the painting nib, and the direction in which the stylus is "falling" for direction. Mapping those from `tiltX` and `tiltY` is confusing. It also makes those values dependent on the twist.

patrickhlauke:
Adapted the description slightly from https://w3c.github.io/touch-events/

Admittedly jotted down very handwavy advice note about implementations - this needs reviewing. Also, if this is taken up, we'll need relevant W3C tests for these new attributes.

Closes #274

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/316.html" title="Last updated on Apr 18, 2020, 4:01 PM UTC (53010a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/316/90ac201...53010a3.html" title="Last updated on Apr 18, 2020, 4:01 PM UTC (53010a3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1200,10 +1200,51 @@ partial interface Navigator {
 function spherical2tilt(altitudeAngle, azimuthAngle){
   let radToDeg = 180/Math.PI;
 
-  let tanAlt = Math.tan(altitudeAngle);
+  let tiltXrad = 0;
+  let tiltYrad = 0;
 
-  let tiltXrad = Math.atan(Math.cos(azimuthAngle) / tanAlt);
-  let tiltYrad = Math.atan(Math.sin(azimuthAngle) / tanAlt);
+  if(altitudeAngle == 0){
+    // the pen is in the X-Y plane
+    if(azimuthAngle == 0 || azimuthAngle == 2*Math.PI){
+      // pen is on positive X axis
+      tiltXrad = Math.PI/2;
+    }
+    if(azimuthAngle == Math.PI/2){
+      // pen is on positive Y axis
+      tiltYrad = Math.PI/2;
+    }
+    if(azimuthAngle == Math.PI){
+      // pen is on negative X axis
+      tiltXrad = -Math.PI/2;
+    }
+    if(azimuthAngle == 3*Math.PI/2){
+      // pen is on negative Y axis
+      tiltYrad = -Math.PI/2;
+    }
+    if(azimuthAngle&gt;0 && azimuthAngle&lt;Math.PI/2){
+      tiltXrad = Math.PI/2;
+      tiltYrad = Math.PI/2;
+    }
+    if(azimuthAngle&gt;Math.PI/2 && azimuthAngle&lt;Math.PI){
+      tiltXrad = -Math.PI/2;
+      tiltYrad = Math.PI/2;
+    }
+    if(azimuthAngle&gt;Math.PI && azimuthAngle&lt;3*Math.PI/2){
+      tiltXrad = -Math.PI/2;
+      tiltYrad = -Math.PI/2;
+    }
+    if(azimuthAngle&gt;3*Math.PI/2 && azimuthAngle&lt;2*Math.PI){
+      tiltXrad = Math.PI/2;
+      tiltYrad = -Math.PI/2;
+    }
+  }
+
+  if(altitudeAngle != 0){
+    let tanAlt = Math.tan(altitudeAngle);
+
+    tiltXrad = Math.atan(Math.cos(azimuthAngle) / tanAlt);
+    tiltYrad = Math.atan(Math.sin(azimuthAngle) / tanAlt);
+  }
 
   return {"tiltX":tiltXrad*radToDeg, "tiltY":tiltYrad*radToDeg};
 }
@@ -1215,16 +1256,16 @@ function tilt2spherical(tiltX, tiltY){
   let azimuthAngle = 0;
 
   if(tiltXrad == 0){
-    if(tiltYrad > 0){
+    if(tiltYrad &gt; 0){
       azimuthAngle = Math.PI/2;
     }
-    else if(tiltYrad < 0){
+    else if(tiltYrad &lt; 0){
       azimuthAngle = 3*Math.PI/2;
     }
   }
 
   if(tiltYrad == 0){
-    if(tiltXrad < 0){
+    if(tiltXrad &lt; 0){
       azimuthAngle = Math.PI;
     }
   }
@@ -1237,13 +1278,11 @@ function tilt2spherical(tiltX, tiltY){
 
   if(tiltXrad != 0 && tiltYrad != 0 &&
      Math.abs(tiltXrad) != Math.PI/2 && Math.abs(tiltYrad) != Math.PI/2){
-    //if pen is segment [(x0,y0,0),(x0+a, y0+b,c)] c>0, and the screen is facing up
-    //towards positive z axis
     let tanX = Math.tan(tiltXrad);
     let tanY = Math.tan(tiltYrad);
 
     azimuthAngle = Math.atan2(tanY, tanX);
-    if(azimuthAngle < 0){
+    if(azimuthAngle &lt; 0){
       azimuthAngle += 2*Math.PI;
     }
   }
--- a/index.html
+++ b/index.html
@@ -1197,63 +1197,74 @@ partial interface Navigator {
         <p>Depending on the specific hardware and platform, user agents will likely only receive one set of values for the transducer orientation relative to the screen plane - either <code>tiltX</code> / <code>tiltY</code> or <code>altitudeAngle</code> / <code>azimuthAngle</code>. The following basic code provides an initial suggested approach for converting these values.</p>
         <pre id="example_12" class="example" title="Converting between tiltX/tiltY and altitudeAngle/azimuthAngle"><code>/* Converting between tiltX/tiltY and altitudeAngle/azimuthAngle */
 
-function spherical2tilt(altitudeAngle, azimuthAngle) {
-  var tiltX = Math.atan(Math.cos(azimuthAngle) / Math.tan(altitudeAngle)) * 180 / Math.PI;
-  var tiltY = Math.atan(Math.sin(azimuthAngle) / Math.tan(altitudeAngle)) * 180 / Math.PI;
-  return [tiltX, tiltY]
+function spherical2tilt(altitudeAngle, azimuthAngle){
+  let radToDeg = 180/Math.PI;
+
+  let tanAlt = Math.tan(altitudeAngle);
+
+  let tiltXrad = Math.atan(Math.cos(azimuthAngle) / tanAlt);
+  let tiltYrad = Math.atan(Math.sin(azimuthAngle) / tanAlt);
+
+  return {"tiltX":tiltXrad*radToDeg, "tiltY":tiltYrad*radToDeg};
 }
 
-function tilt2spherical(tiltX, tiltY) {
-  var altitudeAngle = 0, azimuthAngle = 0;
-  var tiltXrad = tiltX * Math.PI / 180;
-  var tiltYrad = tiltY * Math.PI / 180;
+function tilt2spherical(tiltX, tiltY){
+  let tiltXrad = tiltX * Math.PI/180;
+  let tiltYrad = tiltY * Math.PI/180;
 
-  if (tiltXrad == 0) {
-    if (tiltYrad > 0) {
+  let azimuthAngle = 0;
+
+  if(tiltXrad == 0){
+    if(tiltYrad > 0){
       azimuthAngle = Math.PI/2;
-    } else if (tiltYrad < 0) {
+    }
+    else if(tiltYrad < 0){
       azimuthAngle = 3*Math.PI/2;
     }
   }
 
-  if (tiltYrad == 0) {
-    if (tiltXrad < 0) {
+  if(tiltYrad == 0){
+    if(tiltXrad < 0){
       azimuthAngle = Math.PI;
     }
   }
 
-  if (Math.abs(tiltXrad) == Math.PI/2) {
+  if(Math.abs(tiltXrad) == Math.PI/2){
+    // not enough information to calculate azimuth
     azimuthAngle = 0;
   }
 
-  if ((tiltXrad !=0) && (tiltYrad !=0) && (Math.abs(tiltXrad) != Math.PI/2) && (Math.abs(tiltYrad) != Math.PI/2)) {
-    tanX = Math.tan(tiltXrad);
-    tanY = Math.tan(tiltYrad);
-    if ((tanX > 0) && (tanY > 0)) {
-      azimuthAngle = Math.atan(tanY/tanX);
-    }
-    if ((tanX < 0) && (tanY > 0)) {
-      azimuthAngle = Math.PI/2 + Math.atan(-tanX/tanY);
-    }
-    if ((tanX < 0) && (tanY < 0)) {
-      azimuthAngle = Math.PI + Math.atan(tanY/tanX);
-    }
-    if ((tanX > 0) && (tanY < 0)) {
-      azimuthAngle = 3.0/2.0*Math.PI + Math.atan(-tanX/tanY);
+
+  if(tiltXrad != 0 && tiltYrad != 0 &&
+     Math.abs(tiltXrad) != Math.PI/2 && Math.abs(tiltYrad) != Math.PI/2){
+    //if pen is segment [(x0,y0,0),(x0+a, y0+b,c)] c>0, and the screen is facing up
+    //towards positive z axis
+    let tanX = Math.tan(tiltXrad);
+    let tanY = Math.tan(tiltYrad);
+
+    azimuthAngle = Math.atan2(tanY, tanX);
+    if(azimuthAngle < 0){
+      azimuthAngle += 2*Math.PI;
     }
   }
 
-  if (Math.abs(tiltXrad) == Math.PI/2) {
-    altitudeAngle = 0;
-  } else if (tiltXrad == 0) {
-    altitudeAngle = Math.PI/2 - Math.abs(tiltYrad);
-  } else if (tiltYrad == 0) {
-    altitudeAngle = Math.PI/2 - Math.abs(tiltXrad);
-  } else {
-    altitudeAngle = Math.atan(1.0/Math.sqrt(Math.pow(Math.tan(tiltXrad),2) + Math.pow(Math.tan(tiltYrad),2)));
+  let altitudeAngle = 0;
+
+  if (Math.abs(tiltXrad) == Math.PI/2){
+      altitudeAngle = 0
   }
-  
-  return [altitudeAngle, azimuthAngle];
+  else if (tiltXrad == 0){
+    altitudeAngle = Math.PI/2 - Math.abs(tiltYrad);
+  }
+  else if(tiltYrad == 0){
+    altitudeAngle = Math.PI/2 - Math.abs(tiltXrad);
+  }
+
+  if(tiltXrad != 0 && tiltYrad != 0 && Math.abs(tiltXrad) != Math.PI/2){
+    altitudeAngle =  Math.atan(1.0/Math.sqrt(Math.pow(Math.tan(tiltXrad),2) + Math.pow(Math.tan(tiltYrad),2)));
+  }
+
+  return {"altitudeAngle":altitudeAngle, "azimuthAngle":azimuthAngle};
 }</code>
 </pre>
     </section>
--- a/index.html
+++ b/index.html
@@ -287,6 +287,8 @@ dictionary PointerEventInit : MouseEventInit {
     long        tiltX = 0;
     long        tiltY = 0;
     long        twist = 0;
+    double      altitudeAngle = 0;
+    double      azimuthAngle = 0;
     DOMString   pointerType = "";
     boolean     isPrimary = false;
     sequence&lt;PointerEvent> coalescedEvents = [];
@@ -304,6 +306,8 @@ interface PointerEvent : MouseEvent {
     readonly        attribute long        tiltX;
     readonly        attribute long        tiltY;
     readonly        attribute long        twist;
+    readonly        attribute double      altitudeAngle;
+    readonly        attribute double      azimuthAngle;
     readonly        attribute DOMString   pointerType;
     readonly        attribute boolean     isPrimary;
     sequence&lt;PointerEvent> getCoalescedEvents();
@@ -355,6 +359,14 @@ interface PointerEvent : MouseEvent {
                         <dd>
                             <p>The clockwise rotation (in degrees, in the range of [0,359]) of a transducer (e.g. pen stylus) around its own major axis. For hardware and platforms that do not report twist, the value MUST be 0.</p>
                         </dd>
+                    <dt><dfn>altitudeAngle</dfn></dt>
+                        <dd>
+                            <p>The altitude (in radians) of the transducer (e.g. pen stylus), in the range [0,π/2] - where 0 is parallel to the X-Z plane, and π/2 is perpendicular to the X-Z plane. The value 0 should be used for devices which do not support this property.</p>
+                        </dd>
+                    <dt><dfn>azimuthAngle</dfn></dt>
+                        <dd>
+                            <p>The azimuth angle (in radians) of the transducer (e.g. pen stylus), in the range [0, 2π] - where 0 represents a transducer whose cap is pointing in the direction of increasing X values on the X-Z plane, and π/2 represents a transducer whose cap is pointing in the direction of increasing Z values on the X-Z plane. The value 0 should be used for devices which do not support this property.</p>
+                        </dd>
                     <dt><dfn>pointerType</dfn></dt>
                         <dd>
                             <p>Indicates the device type that caused the event (mouse, pen, touch, etc.). If a user agent is to <a>fire a pointer event</a> for a mouse, pen stylus, or touch input device, then the value of <code>pointerType</code> MUST be according to the following table:</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -405,6 +417,9 @@ <h2><code>PointerEvent</code> Interface</h2>
...
 
+                <div class-"note">Pointer Events include two complementary sets of attributes to express the orientation of a transducer relative the X-Z plane: <dfn><code>tiltX</code></dfn> / <dfn><code>tiltY</code></dfn> (introduced in the original Pointer Events specification), and <dfn><code>altitudeAngle</code></dfn> / <dfn><code>azimuthAngle</code></dfn> (adopted from the <a href="https://w3c.github.io/touch-events/">Touch Events - Level 2</a> specification).
+                    Implementations MUST support both sets of attributes. When a Pointer Event is created, and only one set of values is provided, the complementary set of attributes MUST be calculated and initialized.</div>
```
NavidZ: "only one set of values is provided" is a bit confusing. I assume you meant only one set is provided by the platform.

NavidZ (commented):
Do you think it makes sense to include the transformation between these two sets of value as part of the spec?

index.html
patrickhlauke: hope the change clarifies this further

patrickhlauke:
@NavidZ @smaug---- et al, please have a look at the new informative section. i cribbed the core of the code from @k3a 's https://gist.github.com/k3a/2903719bb42b48c9198d20c2d6f73ac1 ... but my programmer chops are not what they used to be, so check that i didn't mess it up (also, opted to go for a simplified pseudo-code/javascript approach rather than Perl)

patrickhlauke:
and in fact, I just realised that @k3a's code uses radians for both sets of values ... but `tiltX`/`tiltY` use degrees (between `-90` and `90`). so assume this needs extra conversion steps in both cases...?

patrickhlauke:
also, looking at https://github.com/jquery/PEP/pull/376/files#diff-3b5aeb16d85e265fdb40c6063ac62a96R160 it seems there the `cos` and `sin` are flipped compared to this (if you look at the `altitudeAngle`/`azimuthAngle` to `tiltX`/`tiltY` conversion). which one's right?

index.html
```diff
@@ -1177,6 +1192,28 @@ <h2>Mapping for devices that do not support hover</h2>
...
+azimuthAngle = 0;
+tiltX = tiltX * Math.PI / 180;
+tiltY = tiltY * Math.PI / 180;
+if (tiltX != 0) {
+  azimuthAngle = Math.PI/2 - Math.atan2(-Math.cos(tiltX) * Math.sin(tiltY), Math.cos(tiltY) * Math.sin(tiltX));
+  if (azimuthAngle &lt; 0) {
+    azimuthAngle += 2*Math.PI;
+  }
+}
+altitudeAngle = Math.PI/2 - Math.acos(Math.cos(tiltX) * Math.cos(tiltY));</code>
```
NavidZ: I did some back of the envelope calculations and my equations are different from these.
I tried some sanity checks on these but they don't add up. First the formulas are not reversible. I mean if I use a given tiltX,tiltY and change it to azimuth,altitude and again change it back I won't get the same tiltX,tiltY. 

Also the boundary cases that are easy to calculate with mind doesn't match either. So let's first agree on those as those also show the origin of these domains and whatnot.

So in my mind these are equivalent. What do you think  @patrickhlauke, @smaug----  @grorg @graouts
`tiltX=0, tiltY=0` == `altitude=pi/2, azimuth=0` (but azimuth technically could be anything as this is pen standing straight)
`tiltX=90, tiltY=0` == `altitude=0, azimuth=0` (this is pen laying on the X axis its side completely touching the surface)
`tiltX=0, tiltY=90 `== `altitude=0, azimuth=pi/2` (this is pen laying on the Y axes and its side completely touching the surface)
`tiltX=90, tiltY=90` == `altitude=0, azimuth=3*pi/2`


These values are based on the assumption that azimuth is going to be 0 on X axis (we can choose something like north and top of the screen but let's discuss) and going to increase on counterclock-wise rotation when looking at the plane from the top. Also the X axis is in the right direction of the screen and y axis is towards the top of the screen and Z axis is coming out of the plane (like right-handed coordinate system shown on the graphics in the spec).

However, the current Chrome implementation is a left handed one for tiltX and tiltY. i.e. right is positive X and down is positive Y. It does seem inline with @patrickhlauke [demo](https://patrickhlauke.github.io/touch/pen-tracker/) graphics. Unfortunately FF showed zero for tiltX/Y on my Surface book I have for testing. Not sure how the old Edge used to behave. So I believe It makes sense to change the spec in that case to match Chrome as the only implementation on that (i.e. X axis to the right of the screen, Y axis to the bottom and Z coming out of the plane). If we do so the expectation above regarding the transformation between the two coordinate system also change. So let's first decide on this.

patrickhlauke:
> the formulas are not reversible

that's probably because from what i can see, the altitude/azimuth approach has no concept of a pen's twist. if you tilt a pen from 0 to fully perpendicular and then carry on tilting further, in the tiltX/tiltY model you can tell the pen was moved over the perpendicular point, whereas in the other model it's the same as if you tilted the pen from the other side up. if that makes sense.

NavidZ:
> that's probably because from what i can see, the altitude/azimuth approach has no concept of a pen's twist. if you tilt a pen from 0 to fully perpendicular and then carry on tilting further, in the tiltX/tiltY model you can tell the pen was moved over the perpendicular point, whereas in the other model it's the same as if you tilted the pen from the other side up. if that makes sense.

Unfortunately I don't have a pen that reports a nonzero twist. But are you saying if I twist the pen the tilt axes will change and no longer parallel to the screen borders? I know we don't do any adjustment on Chrome and just report what we get from the OS.

patrickhlauke:
no sorry, i mashed two thoughts together there. let me see if i can clarify: on windows etc. where tiltX/tiltY is used, pens also have a particular orientation/twist (e.g. where the button is on the barrel). and that lets you naturally distinguish between a pen that's only raised less than 90 degress from the surface (with the button pointing to the ceiling, if you will) and one that's gone "over" the perpendicular position (with the button pointing to the floor). but in the altitude/azimuth model, the second case doesn't exist. if you go over the perpendicular position, the azimuth simply goes 180 degrees the other way and the more you continue, the lower you're making the altitude again. you don't have a concept of a pen being upside down / gone past the perpendicular position (basically because altitude is only from 0 to 90 degrees, whereas in the tiltx/tilty model you have an additional "attitude" of the pen which makes you make more specific statements between 0 and 180 degrees).

anyway, sounds like this could get very complicated and edge-casey. something i hadn't considered initially. 

also, worth going to the source of what @k3a based their code on, perhaps https://code.woboq.org/qt5/qtbase/src/plugins/platforms/windows/qwindowstabletsupport.cpp.html#527 as that seems to have some further nuances in the code?

patrickhlauke:
thinking about this some more (though admittedly i'm slowly out of my technical depth here), essentially the two models differ at a fundamental level, tied to the kinds of sensors (or the data provided in the low-level/OS-level APIs). the `tiltX`/`tiltY` model is richer since it captures states that are simply not available/detectable in the `altitudeAngle`/`azimuthAngle` model ... those where a stylus is still pointing one particular way but the user went "over" the perpendicular mark.

photos showing the pen going beyond perpendicular

wondering if it's sufficient to address this in prose? aknowledge that going from the `tiltX`/`tiltY` to the `altitudeAngle`/`azimuthAngle` model is "lossy"? and that on platforms that (natively/in their low-level APIs) only support the latter, certain nuances (like the pen being over the perpendicular) won't be captured/capturable?

also yes, @grorg @graouts i'd be interested if there's an official `azimuthAngle` for when a stylus is perfectly perpendicular (where it could be anything really)? is there a default/fallback of zero?

NavidZ:
> thinking about this some more (though admittedly i'm slowly out of my technical depth here), essentially the two models differ at a fundamental level, tied to the kinds of sensors (or the data provided in the low-level/OS-level APIs). the `tiltX`/`tiltY` model is richer since it captures states that are simply not available/detectable in the `altitudeAngle`/`azimuthAngle` model ... those where a stylus is still pointing one particular way but the user went "over" the perpendicular mark.

So looking at these pictures both of them show different sets of tiltX/tiltY and also altitude/azimuth values. Why do you say they have the same altitude/azimuth? They might have the same altitude as it seems they are in the same "height" (I'm just eyeballing it). But the azimuth is certainly different. For one on the left it is more close to 7xpi/4 and for the one on the right it seems to be 3xpi/4. I don't see the lossy model you are referring to here.

> also yes, @grorg @graouts i'd be interested if there's an official azimuthAngle for when a stylus is perfectly perpendicular (where it could be anything really)? is there a default/fallback of zero?

Regarding the straight pen and the fact that perpendicular state could have any azimuth, we do have a similar situation for tiltX/tiltY. But at another location which is less likely to happen. That is when the pen is lying on the x axels on the surface. That can have any tiltY. Having said that I'm not saying we even get any events from the device in that case as the tip might not be even touching the screen and hence the device might not generate any events. So that is just hypothetical.

patrickhlauke:
it might be that i'm getting myself confused here. need to dig out my pen and look at some values coming back. i somehow had it in my head that there's a distinction between these two states, but i might be completely wrong (and if i am, good)

pen being held "over"/"beyond" the perpendicular, and same pen position but held with the other hand from the other side

[edit: the difference is probably only in the `twist` of the pen rather than `tiltX`/`tiltY` ... i think i got myself all confused thinking in theory, i should just open up and use my tests and echo out the raw values i think to get my mind straightened out]

NavidZ:
> it might be that i'm getting myself confused here. need to dig out my pen and look at some values coming back. i somehow had it in my head that there's a distinction between these two states, but i might be completely wrong (and if i am, good)
> 
> pen being held "over"/"beyond" the perpendicular, and same pen position but held with the other hand from the other side
> 
> [edit: the difference is probably only in the `twist` of the pen rather than `tiltX`/`tiltY` ... i think i got myself all confused thinking in theory, i should just open up and use my tests and echo out the raw values i think to get my mind straightened out]

Yeah. I don't have a pen that supports twist in hand to see what its driver reports. But my understanding is that if you hold the pen at a particular position in space and just twist it only the twist value should change and not tiltX/tiltY.  @SummerLW do you happen to have the stylus that reports twist handy to test this?

liviutinta:
I did the trig math for altitude/azimuth so please have a look. Let me know if I made mistakes. 

If pen's tip is resting at `(x0, y0, 0)` and let's say the length of the pen is `L=sqrt(a^2+b^2+c^2)` then the pen is spatially fully described by its tip `(x0, y0, 0)` and cap `(a+x0, b+y0, c)`. 

Now we can compute `tiltX`, `tiltY`, `altitude` and `azimuth` in terms of `a`,`b`,`c`. The terms `x0`,`y0` do not matter as they are not participating in computing distances. `x0`, `y0` will not appear going forward.

`TiltX` angle is described by the angle between normals to the planes `(pen, Y)` and `(Z,Y)`: `N(pen,Y) = (-c, 0, a)` and `N(Z,Y)=(-1, 0, 0)`. Which gives us:
```
cos(tiltX)=c/sqrt(a^2+c^2) 
sin(tiltX)=a/sqrt(a^2+c^2)
tan(tiltX)=a/c
```
`TiltY` angle is described by the angle between normals to the planes `(pen, X)` and `(Z,X)`: `N(pen,X)=(0,c,-b)`, `N(Z,X)=(0,1,0)`. Which gives us:
```
cos(tiltY)=c/sqrt(b^2+c^2)
sin(tiltY)=b/sqrt(b^2+c^2)
tan(tiltY)=b/c
```
`Altitude` is the angle between the vector `(a,b,c)` and plane `(X,Y)` described by angle between vectors `(a,b,c)` and `N(X,Y) = (0,0,1)`. Which gives us:
```
cos(altitude)=sqrt(a^2+b^2)/sqrt(a^2+b^2+c^2)
sin(altitude)=c/sqrt(a^2+b^2+c^2)
tan(altitude)=c/sqrt(a^2+b^2)
``` 
`Azimuth` is the angle between the projection of the pen on `(X,Y)` plane and the `X` axis. The projection of the pen has vector `(a,b,0)`, the `X` axis has vector `(1,0,0)`. Which gives us:
```
sin(azimuth)=b/sqrt(a^2+b^2)
cos(azimuth)=a/sqrt(a^2+b^2)
tan(azimuth)=b/a
``` 
Having the above formulas it is easy to verify the formulas at https://github.com/jquery/PEP/pull/376/files#diff-3b5aeb16d85e265fdb40c6063ac62a96R160 to be correct:
```
tiltX=Math.atan(Math.cos(azimuth) / Math.tan(altitude)) * radToDeg
tiltY=Math.atan(Math.sin(azimuth) / Math.tan(altitude)) * radToDeg
```
For computing `altitude` from `tiltX` and `tiltY` we can easily compute `a` and `b` in terms of `c` from `tan(tiltY)=b/c` and `tan(tiltX)=a/c` and then substitute in formula for `altitude`.
```
altitude=Math.atan(1/sqrt(Math.pow(Math.tan(tiltX),2)+Math.pow(Math.tan(tiltY),2))) 
```
For `azimuth` we can use function `Math.atan2` where `Math.atan2(y,x)` represents the angle between `X` axis and vector `(x,y,0)`.
Using `tan(tiltY)=b/c` and `tan(tiltX)=a/c` we have the following formula (making the assumption `c > 0`):
```
tanX = Math.tan(tiltX)
tanY = Math.tan(tiltY)
azimuth = Math.atan2(tanY,tanX)
if(azimuth < 0)
   azimuth += 2*Math.PI
```
Added code for the above here (code includes edge cases):
https://gist.github.com/liviutinta/0d040270ecf3390972421cb2164fbf37

[1] https://www.w3.org/TR/pointerevents3/
[2] https://w3c.github.io/touch-events/
[3] https://www.analyzemath.com/stepbystep_mathworksheets/vectors/vector3D_angle.html
[4] https://onlinemschool.com/math/library/analytic_geometry/plane_angl/

index.html
```diff
@@ -355,6 +359,14 @@ <h2><code>PointerEvent</code> Interface</h2>
                         <dd>
                             <p>The clockwise rotation (in degrees, in the range of [0,359]) of a transducer (e.g. pen stylus) around its own major axis. For hardware and platforms that do not report twist, the value MUST be 0.</p>
                         </dd>
+                    <dt><dfn>altitudeAngle</dfn></dt>
+                        <dd>
+                            <p>The altitude (in radians) of the transducer (e.g. pen stylus), in the range [0,π/2] - where 0 is parallel to the X-Z plane, and π/2 is perpendicular to the X-Z plane. The value 0 should be used for devices which do not support this property.</p>
```
liviutinta: Based on https://w3c.github.io/touch-events/#dom-touch-altitudeangle I believe the plane should be X-Y not X-Z. When reading the definitions I found the touch-events one more intuitive: pen parallel to the surface / perpendicular to the surface.

index.html
```diff
@@ -355,6 +359,14 @@ <h2><code>PointerEvent</code> Interface</h2>
...
+                        </dd>
+                    <dt><dfn>azimuthAngle</dfn></dt>
+                        <dd>
+                            <p>The azimuth angle (in radians) of the transducer (e.g. pen stylus), in the range [0, 2π] - where 0 represents a transducer whose cap is pointing in the direction of increasing X values on the X-Z plane, and π/2 represents a transducer whose cap is pointing in the direction of increasing Z values on the X-Z plane. The value 0 should be used for devices which do not support this property.</p>
```
liviutinta: Based on https://w3c.github.io/touch-events/#dom-touch-azimuthangle, same comment as for altitudeAngle, I believe it should be plane X-Y not X-Z. As well, I feel for azimuth angle π/2 the axis should be Y, not Z. The definition of azimuth angle for touch events seems to me more intuitive.

patrickhlauke: Agree, the X-Z thing was my mistake. I've expanded this (also adding a clock analogy, which may be easier to convey/understand) and the fact that the value increases when going clockwise

index.html
patrickhlauke: Agree, the X-Z thing was my mistake. I've corrected this and expanded this a bit to talk about the surface.

patrickhlauke:
@liviutinta made a stab at changing the code to JavaScript (as I'm not sure if we want python code here - while in theory I guess any code would do, it feels a bit more natural providing it in JS here, as in the end it's just pseudo-code for browsers to then implement)). I'm not the cleanest or most elegant of coders, so happy for this to be made nicer/more modern/better documented.

liviutinta:
After discussion with @NavidZ and while re-writing the Python code in Javascript, I was able to simplify the `azimuth` formula by using `Math.atan2`. I've updated my post [above](https://github.com/w3c/pointerevents/pull/316#issuecomment-606765923).
Javascript code is here: https://gist.github.com/liviutinta/0d040270ecf3390972421cb2164fbf37

patrickhlauke:
thanks @liviutinta - i've rejigged it slightly (prefer to keep just two functions, rather than splitting out into helper functions for special cases). think it still all works as it should.

did the changes to the explanations/prose make (more) sense?

index.html
```diff
@@ -1197,63 +1197,74 @@ <h2>Converting between <code>tiltX</code> / <code>tiltY</code> and <code>altitud
...
-    }
-    if ((tanX < 0) && (tanY < 0)) {
-      azimuthAngle = Math.PI + Math.atan(tanY/tanX);
-    }
-    if ((tanX > 0) && (tanY < 0)) {
-      azimuthAngle = 3.0/2.0*Math.PI + Math.atan(-tanX/tanY);
+
+  if(tiltXrad != 0 && tiltYrad != 0 &&
+     Math.abs(tiltXrad) != Math.PI/2 && Math.abs(tiltYrad) != Math.PI/2){
+    //if pen is segment [(x0,y0,0),(x0+a, y0+b,c)] c>0, and the screen is facing up
```
liviutinta: The comments are mainly my notes (as I did the math in a,b,c first). I think they are not required in the final document.

liviutinta (commented):
I have no problem understanding the definitions now. Really like the clock analogy as it makes it very clear what is going on. 
For azimuth I believe it would be useful to specify that Y values increase in the direction of "6 o'clock" to mitigate any misunderstanding of the coordinates system.

index.html
```diff
@@ -1197,63 +1197,74 @@ <h2>Converting between <code>tiltX</code> / <code>tiltY</code> and <code>altitud
         <p>Depending on the specific hardware and platform, user agents will likely only receive one set of values for the transducer orientation relative to the screen plane - either <code>tiltX</code> / <code>tiltY</code> or <code>altitudeAngle</code> / <code>azimuthAngle</code>. The following basic code provides an initial suggested approach for converting these values.</p>
         <pre id="example_12" class="example" title="Converting between tiltX/tiltY and altitudeAngle/azimuthAngle"><code>/* Converting between tiltX/tiltY and altitudeAngle/azimuthAngle */
 
-function spherical2tilt(altitudeAngle, azimuthAngle) {
-  var tiltX = Math.atan(Math.cos(azimuthAngle) / Math.tan(altitudeAngle)) * 180 / Math.PI;
-  var tiltY = Math.atan(Math.sin(azimuthAngle) / Math.tan(altitudeAngle)) * 180 / Math.PI;
-  return [tiltX, tiltY]
+function spherical2tilt(altitudeAngle, azimuthAngle){
```
liviutinta, line 1200: Reviewing I realized I did not originally consider edge cases for `spherical2tilt`. The concern is when `altitudeAngle` is zero because `Math.tan(altitudeAngle)` will also be zero and the formula won't work (as we divide by `Math.tan(altitudeAngle)`). This case describes the pen flat on the surface of `X-Y` plane. When pen is on the `X` axis, `tiltY` doesn't really exist so I set it to zero. When pen is on `Y` axis, `tiltX` doesn't really exist and I set it to zero. The reason the tilt angles do not exist is because the pen and the axis form one line, and we need at least two lines to have a plane in order to calculate an angle.  

Javascript code here: https://gist.github.com/liviutinta/0d040270ecf3390972421cb2164fbf37

index.html
```diff
@@ -355,6 +359,14 @@ <h2><code>PointerEvent</code> Interface</h2>
                         <dd>
                             <p>The clockwise rotation (in degrees, in the range of [0,359]) of a transducer (e.g. pen stylus) around its own major axis. For hardware and platforms that do not report twist, the value MUST be 0.</p>
                         </dd>
+                    <dt><dfn>altitudeAngle</dfn></dt>
+                        <dd>
+                            <p>The altitude (in radians) of the transducer (e.g. pen stylus), in the range [0,π/2] - where 0 is parallel to the surface (X-Y plane), and π/2 is perpendicular to the surface. For hardware and platforms that do not report tilt or angle, the value MUST be 0.</p>
```
NavidZ, line 364: I believe the default value of altitude should match the default value of tiltX/TiltY. So if they are 0 (for example when things aren't supported) then altitude should be pi/2.

One general suggestion is that if you add a new line after every comma or sentence it avoids the long lines here and easier to review. This is what I have seen in other specs as well and people seem to be generally happy with it.

NavidZ (approved):
Beside the comment regarding the value of altitude needing to be pi/2 when it is not supported everything else looks good to me.

index.html
patrickhlauke: regarding the default value: this matches what touch events level 2 says https://w3c.github.io/touch-events/#dom-touch-altitudeangle

i've also tested this on an ipad (with no stylus, just using finger) and a modified version of my HUD tracker (that explicitly still shows touch events, rather than pointer events) https://patrickhlauke.github.io/touch/tracker/multi-touch-tracker-pointer-hud-toucheventsonly.html and indeed `azimuthAngle` and `altitudeAngle` are both defined (so the props exist), and return `0`. while i agree that it would be logical for the default to be pi/2, i think it's better to match touch events and what current implementations there do for defaults.

patrickhlauke: as for line length, agree...but as most other text in our spec already breaks this idea, i just went with this approach at this point. may need a good separate isolated reformatting pull request

NavidZ: I'm just trying to follow the same logic we did for pressure. In TouchEvent spec force is also zero if the device doesn't support it. But in PointerEvents it is 0.5 as we thought it makes more sense for the apps to still work with those devices.
With the same argument I believe we should diverge from TouchEvent spec in this case as well and set azimuth to pi/2 when it is not supported by the hardware. WDYT?

patrickhlauke:
Not forgetting that we'll then also need matching w3c tests

index.html
patrickhlauke: sorry, only saw this comment after i went ahead with the merge...

while yes, we diverged for the pressure, i think here the danger / potential for confusion is greater because we went for exactly the same property names as touch events. happy to revisit though if you/others feel strongly about it.

NavidZ: I feel like the ultimate question is how usually apps use this attribute. Since we do not expose a not-supported value for these attributes, apps will have no way of finding out whether 0 is real 0 or it is not supported. 

For pressure that was troublesome (and hence I'm not sure how that would even worked in the touch event spec). If the app draws a line with a thickness relative to the pressure and with 0 pressure draws a line with 0 thickness (i.e. nothing) then if we had set the not supported pressure value to 0 then those app would just fail with those devices. So we chose 0.5 as the pressure for those devices (include mouse) in PointerEvents. Again they will have no way of finding out whether it is a real pressure or it is the not supported value. Hence the reason I'm confused how that choice even worked in TouchEvent spec. I can only assume all touch surfaces have some sort of pressure detection or apps just don't use the force attribute in TouchEvents the way I described above otherwise it has to fail.

And now the real question is how apps use the altitude value. Between 0 and pi/2 I felt pi/2 seems more natural to what apps might want to do with them. But I'm no expert. @smaug---- @grorg for more visibility.

patrickhlauke: conceptually, i totally agree. just going by the reality of what webkit/safari seems to do with force and altitudeAngle in situations where those aren't detected (`0` in both cases). but yes, maybe @grorg has some insight into what, if anything, web apps on iOS (are supposed to) do here to distinguish between a "real" `0` and a "isn't supported" `0`)

patrickhlauke: though i'll note that perhaps for `force` it's slightly moot because even the lightest of touches on a touchscreen is detected as having force `0`. it's only any further/harder force that is measured. and i believe that, since not all iOS devices have force sensing, the general approach taken by app developers has been to work regardless, of force / even when force is `0`, and then provide *additional* functionality based on any force greater than `0`.

i'm not completely against saying that in PE, we would instead default to an `altitudeAngle` of pi/2 for logical/conceptual reasons...but then we likely want to have a very clear warning.

further thought, and more crucially: IF in touch events the default is zero, and that's somehow due to what the sensor/OS sends along...then our conversion algorithm we provide also won't know if the zero it gets (when it then needs to convert it to `tiltX`/`tiltY`) is "not supported" or a "real" `0` ...

NavidZ: I'm also okay either way as long as we clearly consider the use cases that these new attributes are used for and justify our decision.

Regarding the conversion within the UA in Chrome we actually do have a n/a for those values like pressure when things aren't supported and then we report 0 at the very end. So I would imagine we get the "not supported" from the platforms and we can do the right thing in the conversion phase.

patrickhlauke: maybe `altitudeAngle` of `0` is simply assumed as being the default / not supported value because it's impossible (i assume) to really have that situation, since when the stylus is truly resting fully flat on the surface, the stylus tip isn't touching the screen itself so wouldn't be detectable anyway (unless you have a stylus that is "hover-capable" like, say, the pen on the galaxy note). i assume the apple pencil doesn't do hovering, in which case this assumption of "zero can't be the 'real' value" may make sense

patrickhlauke: confirmed (as i don't have a device myself) here https://twitter.com/yatil/status/1253658517241765895

NavidZ: I remember another reason for not exposing a "not supported" value was that developers didn't need to check for it every time. They just use the attribute and it works reasonably.
However, if we do go down the route of exposing a value that developers need to check for not supported case and do something other than what they would do with the actual value we are exposing I personally prefer to be much more explicit on that. Like expose an N/A or something that is clearly outside the range of possible values. I agree that Apple pen doesn't send 0 altitude at this point but I think we should be future proofing the API as well.

patrickhlauke: question is if even for non-apple-pencil stylus scenarios if there's a situation where a completely flat stylus would be detected or not.

i'm just thinking that IF there are web apps already dealing with altitudeAngle for touch events/pencil, they'd be familiar with the "if it's zero, just ignore this altogether" scenario. and because we use the exact same property name in PE, it may cause more confusion / need for forking code if in PE we then have different default.

either way yes, we need to document this more clearly either way, and also look at the pseudocode to check for that edge case/scenario.

patrickhlauke: split this out into https://github.com/w3c/pointerevents/issues/320

smaug----:
https://searchfox.org/mozilla-central/rev/05f486c76f84830041e51b7c68e3e697dc768454/testing/web-platform/tests/pointerevents/pointerevent_tiltX_tiltY_to_azimuth_altitude.html and others are testing this

PR #317, opened 2020-03-09
Change the type of click, auxclick, and contextmenu to PointerEvent

### LINKED ISSUE #100: Should "click", "dblclick" and "contextmenu" events be PointerEvents?
@dtapuska just noticed that Edge sends `click` as an instance of `PointerEvent`, not just `MouseEvent`.  I think we've agreed before that this was a good idea, but I don't think we've talked about speccing it.

I think we can just add a short section to the PE spec "upgrading" `click` to `PointerEvent`.  Also `dblclick` and `contextmenu`.  @teddink, that's it right?  I didn't check drag events or anything.

NavidZ:
This is the followup change to WHATWG html spec and ui events spec pull requests to change the type of click, auxclick, and contextmenu events to pointerevents:

- https://github.com/w3c/uievents/pull/259
- https://github.com/whatwg/html/pull/5324

Closes https://github.com/w3c/pointerevents/issues/100

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/317.html" title="Last updated on Dec 14, 2020, 12:43 AM UTC (ce9a14b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/317/90ac201...ce9a14b.html" title="Last updated on Dec 14, 2020, 12:43 AM UTC (ce9a14b)">Diff</a>


patrickhlauke:
what's the status of this? just waiting for a review, or any more things happening upstream/in UIEvents?

NavidZ:
The ui events part landed. The html spec part is sent. But Domenic suggested to wait until Chrome at least enables this and makes sure there is no regression. If so we will land the one in html spec. We can do either. Either merge now and if it turns out to cause some unrecoverable regressions then revert it. Or just wait until Chrome ships and confirms there is no bad regressions and then merge. I'll leave it up to you.

index.html
```diff
@@ -781,6 +781,17 @@ <h3>The <dfn><code>gotpointercapture</code> event</dfn></h3>
...
             </section>
+            <section>
+                <h3>The <dfn><code>click</code>, <code>auxclick</code>, and <code>contextmenu</code> events</dfn></h3>
+                <p>This section is an addition to <a href="https://w3c.github.io/uievents/#event-type-click">click</a> and
+                   <a href="https://w3c.github.io/uievents/#event-type-auxclick">auxclick</a> events defined in [[UIEVENTS]] and
+                   <a href="https://html.spec.whatwg.org/multipage/indices.html#event-contextmenu">contextmenu</a> defined in [[HTML]].
+                   The type of these events MUST be <code>PointerEvent</code> but the dispatch process is going be what is defined in the original specifications.
```
patrickhlauke: "but the dispatch process is going be what is defined in the original specification" > "but the dispatch process is going to match that of the original specification"

index.html
```diff
@@ -781,6 +781,17 @@ <h3>The <dfn><code>gotpointercapture</code> event</dfn></h3>
...
+                   The additional attributes defined in this spec for <code>PointerEvent</code> for these events are going to be their default values except <code>pointerId</code> and <code>pointertype</code>.
+                   <code>pointerId</code> and <code>pointertType</code> of these events MUST be the same as the PointerEvents that caused these events if they are generated by a PointerEvent stream.
+                   Otherwise, if they were generated by something other than a PointerEvent stream such as
```
patrickhlauke: "stream such as voice recognition or keyboard the" > "stream - such as voice recognition software or a keyboard interaction - the"

patrickhlauke (changes requested):

patrickhlauke (approved):

patrickhlauke:
Just getting this ready but waiting for any tests/regression checks before merging.

smaug----:
@NavidZ Is this now enabled by default in Chrome?

index.html
```diff
@@ -313,7 +313,7 @@ <h2><code>PointerEvent</code> Interface</h2>
                 <dl data-dfn-for="PointerEvent" data-link-for="PointerEvent">
                     <dt><dfn>pointerId</dfn></dt>
                         <dd>
-                            <p>A unique identifier for the pointer causing the event. This identifier MUST be unique from all other <a data-lt="active pointer">active pointers</a> in the <a>top-level browsing context</a> (as defined by [[HTML]]) at the time. A user agent MAY recycle previously retired values for <code>pointerId</code> from previous active pointers, if necessary.</p>
+                            <p>A unique identifier for the pointer causing the event. This identifier MUST be a positive number and unique from all other <a data-lt="active pointer">active pointers</a> in the <a>top-level browsing context</a> (as defined by [[HTML]]) at the time. A user agent MAY recycle previously retired values for <code>pointerId</code> from previous active pointers, if necessary.</p>
```
smaug----: This is not backwards compatible.

smaug---- (dismissed):

NavidZ:
Looking at the code it seems it is still under experimental flag.

@mustaqahmed @liviutinta were running some experiments to make sure it is backward compatible and it doesn't cause any issues. I haven't heard any bad news out of that so far. Could you provide any update you might have?

liviutinta:
We are running an experiment, and gradually expanding its scope. There were no bugs logged yet. If there are no bugs logged by November we plan to start the shipping process. We believe the major risk will be around clientX/clientY being fractional. If there is any breakage, we plan to have a flag in the code to return clientX/clientY as long but keep click/auxclick/contextmenu events as PointerEvent.

index.html
mustaqahmed: I suggest ruling out only "zero" as a special case (and keep allowing negative ids implicitly):

> This identifier MUST be unique from all other active pointers in the top-level browsing context at the time, **except for the value zero which indicates the event is from a non-pointing device**.

patrickhlauke: took a slightly different approach to wording ... see if that makes sense though

patrickhlauke (approved):

index.html
```diff
@@ -781,6 +781,17 @@ <h3>The <dfn><code>gotpointercapture</code> event</dfn></h3>
...
             </section>
+            <section>
+                <h3>The <dfn><code>click</code>, <code>auxclick</code>, and <code>contextmenu</code> events</dfn></h3>
+                <p>This section is an addition to <a href="https://w3c.github.io/uievents/#event-type-click">click</a> and
+                   <a href="https://w3c.github.io/uievents/#event-type-auxclick">auxclick</a> events defined in [[UIEVENTS]] and
+                   <a href="https://html.spec.whatwg.org/multipage/indices.html#event-contextmenu">contextmenu</a> defined in [[HTML]].
+                   The type of these events MUST be <code>PointerEvent</code> but the dispatch process is going to match that of the original specification.
+                   The additional attributes defined in this spec for <code>PointerEvent</code> for these events are going to be their default values except <code>pointerId</code> and <code>pointertype</code>.
```
mustaqahmed: I found "additional" a bit unclear.  What about this?

> For these events, all <code>PointerEvent</code> specific attributes (defined in this spec) other than <code>pointerId</code> and <code>pointertype</code> will have their default values.

index.html
```diff
@@ -781,6 +781,17 @@ <h3>The <dfn><code>gotpointercapture</code> event</dfn></h3>
...
+                   <code>pointerId</code> and <code>pointertType</code> of these events MUST be the same as the PointerEvents that caused these events if they are generated by a PointerEvent stream.
+                   Otherwise, if they were generated by something other than a PointerEvent stream - such as voice recognition
```
mustaqahmed: Let's avoid the word "stream" which currently indicated a "batch" from a given pointerId (used twice).  The current spec uses two other terms to mean the concept we want here: "pointing device" or "pointer device".  So I suggest this:
> Otherwise, if they were generated by something other than a pointing device...

index.html
```diff
@@ -313,7 +313,7 @@ <h2><code>PointerEvent</code> Interface</h2>
                 <dl data-dfn-for="PointerEvent" data-link-for="PointerEvent">
                     <dt><dfn>pointerId</dfn></dt>
                         <dd>
-                            <p>A unique identifier for the pointer causing the event. This identifier MUST be unique from all other <a data-lt="active pointer">active pointers</a> in the <a>top-level browsing context</a> (as defined by [[HTML]]) at the time. A user agent MAY recycle previously retired values for <code>pointerId</code> from previous active pointers, if necessary.</p>
+                            <p>A unique identifier for the pointer causing the event. This identifier MUST be unique from all other <a data-lt="active pointer">active pointers</a> in the <a>top-level browsing context</a> (as defined by [[HTML]]) at the time. The <code>pointerId</code> value of zero is reserved to indicate events that were generated by something other than a PointerEvent stream, such as a non-pointing device. A user agent MAY recycle previously retired values for <code>pointerId</code> from previous active pointers, if necessary.</p>
```
mustaqahmed: Please check my other suggestion to avoid the word "stream".

mustaqahmed (commented):
I have a few more suggestions, sorry didn't spot them last time.

index.html
patrickhlauke: good stuff, yes. changed

index.html
patrickhlauke: done

index.html
patrickhlauke: done

patrickhlauke:
@smaug---- @mustaqahmed @NavidZ @liviutinta one final review ... we happy with the latest wording/changes on this?

index.html
```diff
@@ -781,6 +781,17 @@ <h3>The <dfn><code>gotpointercapture</code> event</dfn></h3>
...
             </section>
+            <section>
+                <h3>The <dfn><code>click</code>, <code>auxclick</code>, and <code>contextmenu</code> events</dfn></h3>
+                <p>This section is an addition to <a href="https://w3c.github.io/uievents/#event-type-click">click</a> and
+                   <a href="https://w3c.github.io/uievents/#event-type-auxclick">auxclick</a> events defined in [[UIEVENTS]] and
+                   <a href="https://html.spec.whatwg.org/multipage/indices.html#event-contextmenu">contextmenu</a> defined in [[HTML]].
+                   The type of these events MUST be <code>PointerEvent</code> but the dispatch process is going to match that of the original specification.
+                   For these events, all <code>PointerEvent</code> specific attributes (defined in this spec) other than <code>pointerId</code> and <code>pointertype</code> will have their default values.
+                   <code>pointerId</code> and <code>pointertType</code> of these events MUST be the same as the PointerEvents that caused these events if they are generated by a PointerEvent stream.
```
mustaqahmed: One more "stream" to get rid of.  Here is a suggestion:

> The <code>pointerId</code> and <code>pointertType</code> attributes of these events MUST be the same as the PointerEvents that caused these events if they are generated by a pointing device. 
 If the events are generated by a non-pointing device (such as voice recognition software or a keyboard interaction), <code>pointerId</code> MUST be 0 and <code>pointerType</code> MUST be an empty string.

mustaqahmed (approved):
Thanks Patrick for addressing my comments, LGTM.  I have one last suggestion; if you see a better way, please go ahead without waiting for my review again.

index.html
patrickhlauke: Done

index.html
```diff
@@ -313,7 +313,7 @@ <h2><code>PointerEvent</code> Interface</h2>
                 <dl data-dfn-for="PointerEvent" data-link-for="PointerEvent">
                     <dt><dfn>pointerId</dfn></dt>
                         <dd>
-                            <p>A unique identifier for the pointer causing the event. This identifier MUST be unique from all other <a data-lt="active pointer">active pointers</a> in the <a>top-level browsing context</a> (as defined by [[HTML]]) at the time. A user agent MAY recycle previously retired values for <code>pointerId</code> from previous active pointers, if necessary.</p>
+                            <p>A unique identifier for the pointer causing the event. This identifier MUST be unique from all other <a data-lt="active pointer">active pointers</a> in the <a>top-level browsing context</a> (as defined by [[HTML]]) at the time. The <code>pointerId</code> value of zero is reserved to indicate events that were generated by something other than a pointing device. A user agent MAY recycle previously retired values for <code>pointerId</code> from previous active pointers, if necessary.</p>
```
smaug----, line 316: "value of zero is reserved to indicate events that were generated by something other than a pointing device." seems like backwards incompatible change.

NavidZ: You are right. But the reason we chose that was no other implementation uses 0 that at this point. So we believe it should be safe to do so now. Otherwise, technically all the numbers are valid in pointerid space. Actually now that we are at it for breaking compat we can force all valid pointerid ones to be positive numbers (again something that the current implementations already do AFAIK) so maybe we can reserver those negative numbers for other future features or something.
Having said that, unless we add yet another field to the prototype to indicate non-pointing device I cannot imagine a way to be fully backward compatible.

mustaqahmed: Good point...looks like we assumed `pointerId` must start with 1.  After your comment, I realized this behavior is mentioned only in the Note below.  I will file an issue now.

smaug----:
There seem to be tests for this. <event>_is_a_pointerevent...

PR #322, opened 2020-05-25
Translate between tiltX/tiltY and azimuth/altitude. Fixes #321

### LINKED ISSUE #321: It is not possible to generate one set of angle values from other set
@patrickhlauke @smaug---- @liviutinta

Based on the latest spec we have the following line:
"When an untrusted (synthetic) Pointer Event is created programmatically using the constructor, and only one set of values is provided, the complementary set of attributes MUST be calculated and initialized"

But apparently based on the current bindings and how the default dictionary values are set we cannot do that. Mainly all the values are present in the PointerEventInit (either when user explicitly set them and if not with their default values as per current spec).
After talking to @domenic, it seems that they expect all the constructors to be dummy setters from dictionary values and if we want to achieve the behavior we described we need to also patch the binding generation from idl and also [dom spec](https://dom.spec.whatwg.org/#concept-event-constructor).

I wonder if we can come up with an easier solution here.

liviutinta:
Add to Pointer Events interface 4 static methods to translate between tiltX/tiltY and azimuth/altitude angles.
+@NavidZ 
+@domenic
+@patrickhlauke

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 27, 2020, 1:31 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fliviutinta%2Fpointerevents%2Fec324ae79c708178c8c361dea73f0244197e2a82%2Findex.html%3FisPreview%3Dtrue)

```
[33m📡 HTTP Error 522:[39m [36mhttps://rawcdn.githack.com/liviutinta/pointerevents/ec324ae79c708178c8c361dea73f0244197e2a82/index.html[39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/pointerevents%23322.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -287,8 +287,6 @@ dictionary PointerEventInit : MouseEventInit {
     long        tiltX = 0;
     long        tiltY = 0;
     long        twist = 0;
-    double      altitudeAngle = 0;
-    double      azimuthAngle = 0;
     DOMString   pointerType = "";
     boolean     isPrimary = false;
     sequence&lt;PointerEvent> coalescedEvents = [];
@@ -306,12 +304,14 @@ interface PointerEvent : MouseEvent {
     readonly        attribute long        tiltX;
     readonly        attribute long        tiltY;
     readonly        attribute long        twist;
-    readonly        attribute double      altitudeAngle;
-    readonly        attribute double      azimuthAngle;
     readonly        attribute DOMString   pointerType;
     readonly        attribute boolean     isPrimary;
     [SecureContext] sequence&lt;PointerEvent> getCoalescedEvents();
     sequence&lt;PointerEvent> getPredictedEvents();
+    static                    double      getAzimuthAngle(long tiltX, long tiltY);
+    static                    double      getAltitudeAngle(long tiltX, long tiltY);
+    static                    long        getTiltX(double azimuthAngle, double altitudeAngle);
+    static                    long        getTiltY(double azimuthAngle, double altitudeAngle);
 };
               </pre>
                 <dl data-dfn-for="PointerEvent" data-link-for="PointerEvent">
@@ -359,14 +359,6 @@ interface PointerEvent : MouseEvent {
                         <dd>
                             <p>The clockwise rotation (in degrees, in the range of [0,359]) of a transducer (e.g. pen stylus) around its own major axis. For hardware and platforms that do not report twist, the value MUST be 0.</p>
                         </dd>
-                    <dt><dfn>altitudeAngle</dfn></dt>
-                        <dd>
-                            <p>The altitude (in radians) of the transducer (e.g. pen stylus), in the range [0,π/2] - where 0 is parallel to the surface (X-Y plane), and π/2 is perpendicular to the surface. For hardware and platforms that do not report tilt or angle, the value MUST be 0.</p>
-                        </dd>
-                    <dt><dfn>azimuthAngle</dfn></dt>
-                        <dd>
-                            <p>The azimuth angle (in radians) of the transducer (e.g. pen stylus), in the range [0, 2π] - where 0 represents a transducer whose cap is pointing in the direction of increasing X values (point to "3 o'clock" if looking straight down) on the X-Y plane, and the values progressively increase when going clockwise (π/2 at "6 o'clock", π at "9 o'clock", 3π/2 at "12 o'clock"). When the transducer is perfectly perpendicular to the surface (<code>altitudeAngle</code> of π/2), the value MUST be 0. For hardware and platforms that do not report tilt or angle, the value MUST be 0.</p>
-                        </dd>
                     <dt><dfn>pointerType</dfn></dt>
                         <dd>
                             <p>Indicates the device type that caused the event (mouse, pen, touch, etc.). If a user agent is to <a>fire a pointer event</a> for a mouse, pen stylus, or touch input device, then the value of <code>pointerType</code> MUST be according to the following table:</p>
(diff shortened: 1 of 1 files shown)


plehegar:
note that the IP checker is rightfully blocking here. If this is planned to be merged, we'll need to resolve the IP question.

index.html
```diff
@@ -413,12 +405,42 @@ <h2><code>PointerEvent</code> Interface</h2>
...
                         </dd>
+                      <dt><dfn>getAzimuthAngle</dfn> <code>(long tiltX, long tiltY)</code></dt>
+                        <dd>
+                            Return the <code>azimuthAngle</code> angle (in radians) of the transducer (e.g. pen stylus) calculated from
+                            <code>tiltX</code> and <code>tiltY</code> angles, in the range [0, 2π] - where 0 represents a transducer whose cap is pointing in the
+                            direction of increasing X values (point to "3 o'clock" if looking straight down) on the X-Y plane, and the
+                            values progressively increase when going clockwise (π/2 at "6 o'clock", π at "9 o'clock", 3π/2 at "12 o'clock").
+                            When the transducer is perfectly perpendicular to the surface (<code>tiltX</code> and <code>tiltY</code> of 0),
+                            the returned value MUST be 0.
+                        </dd>
```
domenic, line 416: It's not clear to me as an implementer how I would implement this method. In particular this section describes the return value, but does not give an algorithm for computing it. The only normative requirements is that `getAzimuthAngle(0, 0)` must return `0`, but it says nothing about the other 2^32 values for each of tiltX and tiltY.

Same for the other three methods.

index.html
```diff
@@ -306,12 +304,14 @@ <h2><code>PointerEvent</code> Interface</h2>
...
-    readonly        attribute double      altitudeAngle;
-    readonly        attribute double      azimuthAngle;
     readonly        attribute DOMString   pointerType;
     readonly        attribute boolean     isPrimary;
     [SecureContext] sequence&lt;PointerEvent> getCoalescedEvents();
     sequence&lt;PointerEvent> getPredictedEvents();
+    static                    double      getAzimuthAngle(long tiltX, long tiltY);
+    static                    double      getAltitudeAngle(long tiltX, long tiltY);
+    static                    long        getTiltX(double azimuthAngle, double altitudeAngle);
+    static                    long        getTiltY(double azimuthAngle, double altitudeAngle);
```
domenic, line 314: Bikeshed: I'm not sure the "get" prefix makes much sense here, since this isn't getting some pre-stored data. I'd just name them `azimuthAngle(...)` etc.

index.html
```diff
@@ -306,12 +304,14 @@ <h2><code>PointerEvent</code> Interface</h2>
     readonly        attribute long        tiltX;
     readonly        attribute long        tiltY;
     readonly        attribute long        twist;
-    readonly        attribute double      altitudeAngle;
-    readonly        attribute double      azimuthAngle;
     readonly        attribute DOMString   pointerType;
     readonly        attribute boolean     isPrimary;
     [SecureContext] sequence&lt;PointerEvent> getCoalescedEvents();
     sequence&lt;PointerEvent> getPredictedEvents();
+    static                    double      getAzimuthAngle(long tiltX, long tiltY);
```
NavidZ, line 311: Alternatively, we can just add getAzimuth and getAltitude functions with no parameter. It gives away the transformation from the current data of the pointerevents.
It does have one caveat. If a test program wants to generate a synthetic pointerevent with a given azimuth/altitude it has to do it by hand and pass it to the constructor. WDYT?

patrickhlauke:
indeed, @liviutinta please see https://labs.w3.org/repo-manager/pr/id/w3c/pointerevents/322

liviutinta:
> indeed, @liviutinta please see https://labs.w3.org/repo-manager/pr/id/w3c/pointerevents/322

@patrickhlauke I've linked my W3C account and my Github account. @plehegar do I need to be a member of the Pointer Events working group as well?

patrickhlauke:
> @patrickhlauke I've linked my W3C account and my Github account. @plehegar do I need to be a member of the Pointer Events working group as well?

i just ran the revalidation, and it now passes the ipr check, so all good (as long as you're member of an organisation that has a rep on the group you're fine)

index.html
liviutinta: @domenic does web idl support overloading for attributes? If I use `static long tiltX(double azimuthAngle, double altitudeAngle)` will it work given there is a `readonly attribute long tiltX`?

index.html
liviutinta: There is a "Converting between `tiltX`/`tiltY` and `altitudeAngle`/`azimuthAngle`" section which provides Javascript code for conversion back and forth. Is that enough?
For all methods I need to provide direction on handling inputs outside of the supported intervals: `[-90,90]` for `tiltX`, `tiltY` and respectively: `[0,π/2]` for `altitudeAngle` and `[0,2π]` for `azimuthAngle`. Would it be ok to recommend the methods throw RangeError exceptions for invalid input?

index.html
liviutinta: > Alternatively, we can just add getAzimuth and getAltitude functions with no parameter. It gives away the transformation from the current data of the pointerevents.

Having `getAzimuthAngle` and `getAltitudeAngle` methods on the PointerEvent makes sense.

> It does have one caveat. If a test program wants to generate a synthetic pointerevent with a given azimuth/altitude it has to do it by hand and pass it to the constructor.

This is the use-case for which I added `static long getTiltX(double azimuth, double altitude)` and `static long getTiltY(double azimuth, double altitude)` methods: to allow easy `tiltX`/`tiltY` calculation given `azimuth`/`altitude` pair.

patrickhlauke:
Per the discussion at today's meeting https://lists.w3.org/Archives/Public/public-pointer-events/2020AprJun/0050.html we've decided to go a different route. Closing this for now.

PR #323, opened 2020-05-28
Update azimuth/altitude/tiltX/tiltY to not require default values in pointer events web idl

liviutinta:
Remove default values for `tiltX`,`tiltY`,`azimuthAngle`,`altitudeAngle`.
Change default value for `altitudeAngle` to `π/2`.
Describe rounding method for calculated `tiltX`,`tiltY`.
If only the value for one attribute is specified, the other attribute must be populated with the default value.
+@patrickhlauke 
+@smaug----  
+@NavidZ
+@domenic

***
<a href="https://pr-preview.s3.amazonaws.com/liviutinta/pointerevents/pull/323.html" title="Last updated on Jun 26, 2020, 8:58 PM UTC (8e22d72)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/323/0dd4377...liviutinta:8e22d72.html" title="Last updated on Jun 26, 2020, 8:58 PM UTC (8e22d72)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -424,14 +424,21 @@ interface PointerEvent : MouseEvent {
                 <p>The <dfn><code>PointerEventInit</code></dfn> dictionary is used by the <dfn><code>PointerEvent</code></dfn> interface's constructor to provide a mechanism by which to construct untrusted (synthetic) pointer events. It inherits from the {{MouseEventInit}} dictionary defined in [[UI-EVENTS]]. The steps for constructing an event are defined in [[DOM]]. See the <a href="#examples" title="examples">examples</a> for sample code demonstrating how to fire an untrusted pointer event.</p>
 
                 <div class="note">
-                    Pointer Events include two complementary sets of attributes to express the orientation of a transducer relative to the X-Y plane: <dfn><code>tiltX</code></dfn> / <dfn><code>tiltY</code></dfn> (introduced in the original Pointer Events specification), and <dfn><code>altitudeAngle</code></dfn> / <dfn><code>azimuthAngle</code></dfn> (adopted from the <a href="https://w3c.github.io/touch-events/">Touch Events - Level 2</a> specification).
-                    Implementations MUST support both sets of attributes. When an untrusted (synthetic) Pointer Event is created programmatically using the constructor, and only one set of attributes is provided, the complementary set of attributes MUST be calculated and initialized by the user agent. If both sets of attributes are provided, no calculation should be performed.
-
-                    <p>
-                    If <code>azimuth</code> and <code>altitude</code> are provided, the user agent MUST initialize <code>tiltX</code>, <code>tiltY</code>
-                    with the calculated values rounded to the nearest integer value. For values falling exactly midway between integers, the user agend MUST always round up.
-                    </p>
-                    <p>If only the value of one of the two attributes is provided, the other attribute MUST be initialized to the default value.</p>
+                    <p>Pointer Events include two complementary sets of attributes to express the orientation of a
+                    transducer relative to the X-Y plane: <code>tiltX</code> / <code>tiltY</code>
+                    (introduced in the original Pointer Events specification), and
+                    <code>azimuthAngle</code> / <code>altitudeAngle</code>
+                    (adopted from the <a href="https://w3c.github.io/touch-events/">Touch Events - Level 2</a> specification).
+                    Implementations SHOULD provide both sets of attributes for trusted events.</p>
+                    <p>When an untrusted (synthetic) Pointer Event is created programmatically using the constructor,
+                    and only one set of attributes is provided, the complementary set of attributes SHOULD be calculated and
+                    initialized by the user agent. If both sets of attributes are provided, no calculation should be performed.
+                    If only the value of one of the two attributes is provided, the other attribute SHOULD be initialized to
+                    the default value.</p>
+                    <p>When the user agent calculates <code>tiltX</code> / <code>tiltY</code> from
+                    <code>azimuthAngle</code> / <code>altitudeAngle</code> it SHOULD round the final integer values using
+                    <a href="https://tc39.es/ecma262/#sec-math.round"><code>Math.round</code></a> ([[ECMA-262]]) rules.</p>
+                    <p></p>
                 </div>
 
                 <p>A <a>PointerEvent</a> has an associated <dfn>coalesced event list</dfn> (a list of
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -361,7 +361,13 @@ <h2><code>PointerEvent</code> Interface</h2>
                         </dd>
                     <dt><dfn>altitudeAngle</dfn></dt>
                         <dd>
-                            <p>The altitude (in radians) of the transducer (e.g. pen stylus), in the range [0,π/2] - where 0 is parallel to the surface (X-Y plane), and π/2 is perpendicular to the surface. For hardware and platforms that do not report tilt or angle, the value MUST be 0.</p>
+                            <p>The altitude (in radians) of the transducer (e.g. pen stylus), in the range [0,π/2] - where 0 is parallel to the surface (X-Y plane), and π/2 is perpendicular to the surface. For hardware and platforms that do not report tilt or angle, the value MUST be π/2.</p>
+                            <div class="note">
+                                When the hardware or platform does not report the tilt or angle, as opposed to the default value of 0 defined for <code>altitudeAngle</code>
+                                in <a href="https://w3c.github.io/touch-events/">Touch Events - Level 2</a> specification,
```
NavidZ, line 367: Just an editorial comment and not important really. You can also reference the touch events with brackets similar to other places in this spec so that it appear in the normative references.

index.html
```diff
@@ -417,8 +423,16 @@ <h2><code>PointerEvent</code> Interface</h2>
...
+                    Implementations MUST support both sets of attributes. When an untrusted (synthetic) Pointer Event is created programmatically using the constructor, and only one set of attributes is provided, the complementary set of attributes MUST be calculated and initialized by the user agent. If both sets of attributes are provided, no calculation should be performed.
+
+                    <p>
+                    If <code>azimuth</code> and <code>altitude</code> are provided, the user agent MUST initialize <code>tiltX</code>, <code>tiltY</code>
+                    with the calculated values rounded to the nearest integer value. For values falling exactly midway between integers, the user agend MUST always round up.
+                    </p>
+                    <p>If only the value of one of the two attributes is provided, the other attribute MUST be initialized to the default value.</p>
```
NavidZ: So does this line refer to both when out of 4 only 1 is present and when out of 4 only 3 is present?

NavidZ:
@liviutinta btw, did this solve our implementation problem we see in Chromium code for "has*" functions?

liviutinta:
> @liviutinta btw, did this solve our implementation problem we see in Chromium code for "has*" functions?

Yes, removing default values for `tiltX`, `tiltY`, `altitudeAngle`, `azimuthAngle` in the `pointer_event.idl` and `pointer_event_init.idl` files solves the problem we've had with "has*" functions. Now, the functions work as I originally expected: for all properties in the set (`tiltX`, `tiltY`, `altitudeAngle`, `azimuthAngle`) with no value in the init dictionary the corresponding "has*" function returns false.

index.html
liviutinta: > So does this line refer to both when out of 4 only 1 is present and when out of 4 only 3 is present?

My intention was to cover both cases. Basically I would like to apply formulas only in these 2 cases:
- `(hasTiltX || hasTiltY) && !hasAzimuthAngle && !hasAltitudeAngle`
- `(hasAzimuthAngle || hasAltitudeAngle) && !hasTiltX && !hasTiltY`

If `tiltX` and `tiltY` and one of `azimuthAngle`/`altitudeAngle` values are provided, I cannot guarantee that `azimuthAngle`/`altitudeAngle` value as provided by the user matches the formula with regards to `tiltX` and `tiltY`. Same goes if `azimuthAngle` and `altitudeAngle` and one of `tiltX`/`tiltY` values are provided.

My assumption is that the user agent is not going to overwrite a value provided by the user. If the user provides values that are not correct it is an authoring problem.

index.html
liviutinta: I checked the [reference database](https://www.specref.org/?q=touch%20events) and there is no entry for Touch Events - Level 2. I believe this is why it is referenced like this here.

NavidZ (approved):
Sounds good to me.

index.html
```diff
@@ -417,8 +423,16 @@ <h2><code>PointerEvent</code> Interface</h2>
...
-                    Implementations MUST support both sets of attributes. When an untrusted (synthetic) Pointer Event is created programmatically using the constructor, and only one set of values is provided, the complementary set of attributes MUST be calculated and initialized.</div>
+                <div class="note">
+                    Pointer Events include two complementary sets of attributes to express the orientation of a transducer relative to the X-Y plane: <dfn><code>tiltX</code></dfn> / <dfn><code>tiltY</code></dfn> (introduced in the original Pointer Events specification), and <dfn><code>altitudeAngle</code></dfn> / <dfn><code>azimuthAngle</code></dfn> (adopted from the <a href="https://w3c.github.io/touch-events/">Touch Events - Level 2</a> specification).
```
mustaqahmed: Please use `<p>...</p>` because we have multiple paragraphs in this `<div>`.

index.html
```diff
@@ -417,8 +423,16 @@ <h2><code>PointerEvent</code> Interface</h2>
...
+                    Implementations MUST support both sets of attributes. When an untrusted (synthetic) Pointer Event is created programmatically using the constructor, and only one set of attributes is provided, the complementary set of attributes MUST be calculated and initialized by the user agent. If both sets of attributes are provided, no calculation should be performed.
```
mustaqahmed: > Implementations MUST support both sets of attributes.

I find "support" a bit too vague.  I suggest this to make it clearer:

> Implementations MUST provide both sets of attributes for trusted events.

index.html
```diff
@@ -417,8 +423,16 @@ <h2><code>PointerEvent</code> Interface</h2>
...
+
+                    <p>
```
mustaqahmed: I have two editorial suggestions about this para about rounding:

1. This para also applies also to trusted events (when UA fills in missing values from device driver).  But the next para as well as the last two sentences in the above para is about untrusted events only.  I think it would be easier to read if we move this para to the end, and have a separate para in the middle for untrusted events.  Wdyt?

2. Let's omit the standard rounding rule here, instead refer to `Math.round()`: https://tc39.es/ecma262/#sec-math.round. I suggest the following to cover both untrusted and trusted cases:
> When the user agent calculates tiltX/tiltY from azimuth/altitude, it SHOULD round the final integer values using the Math.round rules [ref].

index.html
```diff
@@ -417,8 +423,16 @@ <h2><code>PointerEvent</code> Interface</h2>
...
 
-                <div class-"note">Pointer Events include two complementary sets of attributes to express the orientation of a transducer relative the X-Z plane: <dfn><code>tiltX</code></dfn> / <dfn><code>tiltY</code></dfn> (introduced in the original Pointer Events specification), and <dfn><code>altitudeAngle</code></dfn> / <dfn><code>azimuthAngle</code></dfn> (adopted from the <a href="https://w3c.github.io/touch-events/">Touch Events - Level 2</a> specification).
-                    Implementations MUST support both sets of attributes. When an untrusted (synthetic) Pointer Event is created programmatically using the constructor, and only one set of values is provided, the complementary set of attributes MUST be calculated and initialized.</div>
+                <div class="note">
```
mustaqahmed, line 426: I think "MUST" doesn't fit well in non-normative text like "note".  "SHOULD" looks like a better fit. https://tools.ietf.org/html/bcp14

mustaqahmed (commented):
This looks good but I would ask for a few minor edits.

index.html
liviutinta: Done.

index.html
liviutinta: Done

index.html
liviutinta: Done

index.html
liviutinta: Done

mustaqahmed (approved):
Thanks Liviu, this looks good to me now.

mustaqahmed:
We will need approvals from @smaug---- and @patrickhlauke.

NavidZ:
@smaug---- and @patrickhlauke does this look okay?

patrickhlauke (approved):

mustaqahmed:
We have this WPT for this change: [pointerevent_tiltX_tiltY_to_azimuth_altitude.html](https://wpt.fyi/results/pointerevents/pointerevent_tiltX_tiltY_to_azimuth_altitude.html?label=master&label=experimental&aligned&view=subtest).

PR #331, opened 2020-09-16
Config for ECHIDNA and travis

plehegar:
This will allow the spec to be automatically published.

Publication reports are sent to
  https://lists.w3.org/Archives/Public/public-tr-notifications/

If you wish to be cc'ed on the notification, add "cc=foo@bar.com,foo1@bar1.com)" to the curl command in .travis.yml.
See https://github.com/w3c/echidna/wiki/How-to-use-Echidna

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+# ECHIDNA automatic publishing for respec
+# see https://github.com/w3c/echidna/wiki/Setting-up-Echidna-as-a-GitHub-hook
+language: node_js
+sudo: false
+branches:
+  only:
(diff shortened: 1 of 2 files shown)


patrickhlauke (approved):

sidvishnoi:
Hi! A bit late, but I've created a GitHub Action for this and a bit more: [Spec Prod](https://github.com/sidvishnoi/spec-prod/).
It's being used at [[Gamepad]](https://github.com/w3c/gamepad/blob/gh-pages/.github/workflows/pr-push.yml) and a few other places.

PR #349, opened 2021-02-16
Reword/expand touch-action definition

### LINKED ISSUE #276: `touch-action: none` definition is too vague
The spec for the [`touch-action`](https://w3c.github.io/pointerevents/#the-touch-action-css-property) property currently has this to say about its [`none`](
https://w3c.github.io/pointerevents/#details-of-touch-action-values) value:

> Touches that begin on the element MUST NOT trigger default touch behaviors.

The other values for the `touch-action` property concern themselves with panning and zooming, so it makes sense to conclude that those "default touch behaviors" may be related to panning or zooming, but there is no explicit mention of this specifically, nor is there a mention of user interactions around navigation and text. I think this spec should be clearer, even if only as a note, about what is the expected behaviour of `touch-action: none` on text for instance. Chrome on Android allows text selection and traversing a link by tapping on it, but it seems equally conformant to prevent text selection and link traversal.

patrickhlauke:
- trying to make it clear that `touch-action` is ONLY concerned with panning/scrolling/zooming, not other things like text selection/highlighting, or whether or not a UA triggers a link or form control.
- trying to make the slightly handwavy way in which `manipulation` has been defined clearer, without using the specific wording that we determined can't be used due to IPR concencern (relating to a particular way in which zooming can happen, and which has historically led to that 300ms-ish delay in touchscreen interactions)

Closes https://github.com/w3c/pointerevents/issues/276

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/349.html" title="Last updated on Apr 1, 2021, 6:28 PM UTC (91043be)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/349/9b051c9...91043be.html" title="Last updated on Apr 1, 2021, 6:28 PM UTC (91043be)">Diff</a>


index.html
```diff
@@ -989,15 +989,15 @@ <h2>Determining supported touch behavior</h2>
...
             <dt>pan-x<br>pan-left<br>pan-right<br>pan-y<br>pan-up<br>pan-down</dt>
-            <dd>
-              <p>The user agent MAY consider touches that begin on the element only for the purposes of scrolling that starts in any of the directions specified by all of the listed values.  Once scrolling is started, the direction may be reversed by the user even if scrolls that start in the reversed direction are disallowed. In contrast, when scrolling is restricted to starting along a single axis (eg. <code>pan-y</code>), the axis cannot be changed during the scroll.<p>
+            <dd>The user agent MAY consider touches that begin on the element only for the purposes of scrolling that starts in any of the directions specified by all of the listed values.  Once scrolling is started, the direction may be reversed by the user even if scrolls that start in the reversed direction are disallowed. In contrast, when scrolling is restricted to starting along a single axis (eg. <code>pan-y</code>), the axis cannot be changed during the scroll.</dd>
```
mustaqahmed: Perhaps replace "purposes of scrolling" with "purposes of viewport panning/scrolling" to be consistent with the paras above and below?

index.html
```diff
@@ -989,15 +989,15 @@ <h2>Determining supported touch behavior</h2>
...
             <dt>manipulation</dt>
-            <dd>The user agent MAY consider touches that begin on the element only for the purposes of scrolling and continuous zooming.  Any additional behaviors supported by <code>auto</code> are out of scope for this specification.</dd>
+            <dd>The user agent MAY consider touches that begin on the element for the purposes of panning/scrolling and <strong>continuous</strong> zooming, but MUST NOT trigger other related touch behaviors such as <strong>non-continuous</strong> zooming in specific steps / to a specific element.</dd>
```
mustaqahmed: Not sure what I am missing here but I can't see what the terms "(non-)continuous zooming" mean.  Do they mean the same idea as in the second Note in Sec9.1?  Or we just need to isolate double-tap zoom as a "non-continuous zoom"?

Would this be easier to follow if we say:
> The user agent MAY allow scrolling/panning in any direction (pan-x or pan-y)...

At least that's [what Chrome does](https://source.chromium.org/chromium/chromium/src/+/master:cc/input/touch_action.h;drc=9287e2cd3a6b75b0f640a1c0b4c198aa5d7c9c1f;l=34).

patrickhlauke: In short yes, manipulation means you're allowed to do everything except double-tap-to-zoom, but we're not allowed to say double-tap-to-zoom as that defines a specific gesture implementation. so this was a way to somehow explain the effect without saying those words.

manipulation also allows single finger double-tap-then-hold to zoom in/out with one finger (like chrome does for maps for instance) and pinch-to-zoom. it really only isolates double-tap-to-zoom, at least that's the idea/my understanding. and here, i'm kind of attempting to define what this double-tap-to-zoom is...a non-continuous step-wise gesture

index.html
patrickhlauke: good point. done

index.html
patrickhlauke: does this make sense, or does it need further rewording?

index.html
```diff
@@ -989,15 +989,15 @@ <h2>Determining supported touch behavior</h2>
...
           </dl>
+          <p><code>touch-action</code> only covers touch behaviors related to viewport panning/scrolling and zooming. Any additional default user agent behaviors, such as text selection/highlighting, or form controls and link activation, MUST NOT be affected by this CSS property.</p>
```
mustaqahmed: Does a single "or" read better?  
> such as text selection/highlighting, form controls or link activation

index.html
mustaqahmed: We talked about possibly adding a glossary item for "continuous" zooming because so far this term was never used in normative text.  This is tricky to define using only the allowed terms, but let me try anyway:

> **Continuous zoom**: a zoom effect that takes place while one touch pointer remains active.

If this is okay (_please check carefully!_), perhaps we don't need to define "non-continuous", so can drop "...in specific steps / to a specific element".  Thoughts?

index.html
```diff
@@ -989,15 +989,15 @@ <h2>Determining supported touch behavior</h2>
           <h2>Details of <code>touch-action</code> values</h2>
           <dl>
             <dt>auto</dt>
-            <dd>The user agent MAY determine any permitted touch behaviors, such as panning and zooming of the viewport, for touches that begin on the element.</dd>
+            <dd>The user agent MAY determine any permitted touch behaviors related to viewport panning/scrolling and zooming for touches that begin on the element.</dd>
```
mustaqahmed: This was there before, let's fix this while we are here:

> MAY determine

is confusing, we can switch to:

> MAY consider

like in the following paras.

index.html
```diff
@@ -989,15 +989,15 @@ <h2>Determining supported touch behavior</h2>
...
             <dt>none</dt>
-            <dd>Touches that begin on the element MUST NOT trigger default touch behaviors.</dd>
+            <dd>Touches that begin on the element MUST NOT trigger default touch behaviors related to viewport panning/scrolling and zooming for touches that begin on the element.</dd>
```
mustaqahmed: With the existing start of the sentence, we can drop the last phrase "for touches that begin on the element"?

mustaqahmed (commented):
Sorry for taking so long...I will be quicker in the next round.

index.html
patrickhlauke: Turned this around slightly, as it's about the activation (of form controls or links)

index.html
patrickhlauke: Good spot, done

index.html
patrickhlauke: Good spot, done.

mustaqahmed:
Thanks for fixing the definitions and clarifying old text around it, the whole section looks great now.  LGTM.

patrickhlauke:
@smaug---- any thoughts on this? i think it makes most sense for me to merge this (if we're ok with it) and then either rebase the other PR https://github.com/w3c/pointerevents/pull/350 or even supersede it with a fresh one that takes this as the basis.

index.html
```diff
@@ -982,24 +982,24 @@ <h2>Determining supported touch behavior</h2>
...
 
         <section>
           <h2>Details of <code>touch-action</code> values</h2>
           <dl>
             <dt>auto</dt>
-            <dd>The user agent MAY determine any permitted touch behaviors, such as panning and zooming of the viewport, for touches that begin on the element.</dd>
+            <dd>The user agent MAY consider any permitted touch behaviors related to viewport panning/scrolling and zooming for touches that begin on the element.</dd>
             <dt>none</dt>
-            <dd>Touches that begin on the element MUST NOT trigger default touch behaviors.</dd>
+            <dd>Touches that begin on the element MUST NOT trigger default touch behaviors related to viewport panning/scrolling and zooming.</dd>
```
smaug----: This is a bit unclear now. Which default behavior may then be triggered? panning/scrolling and zooming aren't allowed but if some others are allowed, which ones?

smaug---- (dismissed):
The meaning of 'none' is now a bit unclear

index.html
patrickhlauke: would it make sense to say something along the lines of "default touch behaviors related to the movement of active pointers (such as panning/scrolling and zooming); any other behaviors (such as activation of a link/form control, or a long-press and drag for text selection) are allowed."

or does that make it too wordy?

patrickhlauke:
A few more thoughts to discuss:

at the start of [9. Declaring candidate regions for default touch behaviors](https://w3c.github.io/pointerevents/#declaring-candidate-regions-for-default-touch-behaviors), it reads:

```
<p>For touch input, the default action of any and all pointer events MUST NOT be a manipulation of the viewport (e.g. panning or zooming).</p>
```

followed by the note:

```
Touch manipulations are intentionally not a default action of pointer events. Removing this dependency on the cancellation of events facilitates performance optimizations by the user agent.
```

It's unclear to me, as a reader, what this actually means/refers to. What concept is this trying to convey? I'm guessing it's essentially saying here that panning/zooming is special and not counted as a "default action" in the sense of [5.1.3.1 Attributes and Default Actions](https://w3c.github.io/pointerevents/#attributes-and-default-actions), and that for this reason panning/zooming also can't be cancelled by cancelling/preventing the event. This may need a bit more clarification/expansion. Perhaps even foreshadowing what we say later about `pointercancel` being fired. And then leading into the actual reason for this whole section: "instead, authors can control how user agents should handle panning/zooming using the `touch-action` CSS property instead".

That first sentence (about "default action ... MUST NOT be a manipulation of the viewport...") then seems at odds with the first note in [9.1 The touch-action CSS property](https://w3c.github.io/pointerevents/#the-touch-action-css-property)

```
As noted previously, in the case of user agents that allow default behaviors (such as panning or zooming) for other pointer types...
```

In that note, I think we need to drop the mention of "default behaviors" since we just said previously that panning/zooming isn't counted as a default behavior.

---

Now also wondering about the second note at the start of the whole [9. Declaring candidate regions for default touch behaviors](https://w3c.github.io/pointerevents/#declaring-candidate-regions-for-default-touch-behaviors) section

```
While the issue of pointers used to manipulate the viewport is generally limited to touch input (where a user's finger can both interact with content and pan/zoom the page), certain user agents may also allow the same types of (direct or indirect) manipulation for other pointer types. For instance, on mobile/tablet devices, users may also be able to pan using a stylus. This section applies to these scenarios as well (despite the specification's use of "touch").
```

Should we make this normative and more prominent somehow (not sure if we have a style/semantic for it...like a warning boxout that doesn't make it informative only), and if that's sufficient to then simply forego #350

index.html
smaug----: ok, so we don't need this change at all, but only move the paragraph below.

patrickhlauke:
> A few more thoughts to discuss:

decided in our call today that these will be addressed separately (so as not to hold up this PR)

patrickhlauke:
@mustaqahmed @smaug---- please have a look at https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/349.html#details-of-touch-action-values and, if it's all ok, add a positive review

index.html
```diff
@@ -982,24 +982,24 @@ <h2>Determining supported touch behavior</h2>
...
         <section>
           <h2>Details of <code>touch-action</code> values</h2>
+          <p><code>touch-action</code> covers touch behaviors related to viewport panning/scrolling and zooming. Any additional user agent behaviors, such as text selection/highlighting, or activating links and form controls, MUST NOT be affected by this CSS property.</p>
           <dl>
             <dt>auto</dt>
-            <dd>The user agent MAY determine any permitted touch behaviors, such as panning and zooming of the viewport, for touches that begin on the element.</dd>
+            <dd>The user agent MAY consider any permitted touch behaviors related to viewport panning/scrolling and zooming for touches that begin on the element.</dd>
             <dt>none</dt>
-            <dd>Touches that begin on the element MUST NOT trigger default touch behaviors.</dd>
+            <dd>Touches that begin on the element MUST NOT trigger touch behaviors related to viewport panning/scrolling and zooming.</dd>
```
smaug----, line 995: "related to viewport panning/scrolling and zooming" still hints that some other touch behaviors could happen. I thought there would be text before the list to explain that these values don't affect to other behavior outside touch actions.

patrickhlauke: there is text at line 990 just before the `<dl>`. and yes this means that `none` disallows panning/scrolling/zooming, but reinforces/reminds once more that this is just about panning/scrolling/zooming, as already mentioned before the list.

patrickhlauke: i.e. this might be seen as redundant, but i think it's not doing any harm and helps prevent casual readers from just seeing "no touch behaviors allowed" and then being surprised.

index.html
```diff
@@ -982,24 +982,24 @@ <h2>Determining supported touch behavior</h2>
...
           </ul>
 
-          <div class="note">Some user agents support touch actions triggered by interactions of multiple concurrent pointers (e.g. multi-touch). Methods for processing or associating the <code>touch-action</code> values of multiple concurrent pointers is out of scope for this specification.</div>
+          <div class="note">Some user agents support touch actions triggered by interactions of multiple concurrent pointers (e.g. multi-touch). Methods for processing or associating the <code>touch-action</code> values of multiple concurrent pointers are out of scope for this specification.</div>
         </section>
 
         <section>
           <h2>Details of <code>touch-action</code> values</h2>
+          <p><code>touch-action</code> covers touch behaviors related to viewport panning/scrolling and zooming. Any additional user agent behaviors, such as text selection/highlighting, or activating links and form controls, MUST NOT be affected by this CSS property.</p>
```
mustaqahmed: Optional: Starting a section with a lower case word seems a bit odd to me.  If you agree, please consider saying something like:
> The <code>touch-action</code> values below cover touch behaviors related to ...

mustaqahmed (commented):
LGTM modulo @smaug----'s approval.

index.html
patrickhlauke: change to "The `touch-action` property..."

patrickhlauke:
@plehegar any idea why the IPR check comes back as failing? Is my user not in the repository's group anymore? (related, any chance the repo could be changed so that I can merge smaller things without an approval, or is that not recommended procedure?)

mustaqahmed (approved):

PR #350, opened 2021-02-16
Major refactoring: refer to "direct manipulation" rather than "touch"

### LINKED ISSUE #202: Normatively make touch-action apply for direct-manipulation pen devices
We have this note at the top of [section 9](https://w3c.github.io/pointerevents/#declaring-candidate-regions-for-default-touch-behaviors):

> While the issue of pointers used to manipulate the viewport is generally limited to touch input (where a user's finger can both interact with content and pan/zoom the page), certain user agents may also allow the same types of (direct or indirect) manipulation for other pointer types. For instance, on mobile/tablet devices, users may also be able to pan using a stylus. This section applies to these scenarios as well (despite the specification's use of "touch").

We discussed [the PR](https://github.com/w3c/pointerevents/pull/188) a bit [on Twitter](https://twitter.com/patrick_h_lauke/status/841028283533799425).  I think there should be normative text that REQUIRES that touch-action apply to a device if and only if  [InputDeviceCapabilities.pointerMovementScrolls](https://wicg.github.io/InputDeviceCapabilities/#dom-inputdevicecapabilities-pointermovementscrolls) is true.

In particular, I think doing anything else would cause a web compat and interop problem.  Eg. if we decided to change ChromeOS so that pen dragging scrolled but DIDN'T honor touch-action, then tons of sites like Google Maps would be broken when you drag with pen (since you'd get a pointercancel).

The simplest way to do that is to probably refactor the text at the top of [10.3](https://w3c.github.io/pointerevents/#implicit-pointer-capture) into a section that defines the term "direct manipulation pointing device", and then reference that term from both the touch-action and implicit capture sections.

@patrickhlauke @mustaqahmed @NavidZ WDYT?

patrickhlauke:
This is admittedly slightly awkward, but necessary - as `touch-action` does not simply apply to touch, but any other situation where an input serves to directly scroll/zoom (e.g. stylus on touchscreen). Before, it was just a non-normative note, but  if we want to make this normative and require (for web compat) that UAs do this, it's probably the least-bad approach. (still leaves a bit of cognitive dissonance with the fact that the CSS property is called `touch-action`, but hey...)

Closes https://github.com/w3c/pointerevents/issues/202

Please take your time reviewing this, as it's probably quite a fundamental decision here. We should discuss this in the upcoming call(s).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/350.html" title="Last updated on Apr 30, 2021, 4:50 PM UTC (947f883)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/350/dc31af5...947f883.html" title="Last updated on Apr 30, 2021, 4:50 PM UTC (947f883)">Diff</a>


patrickhlauke:
Setting this to draft, as we likely need to wordsmith this / discuss this a bit more than other changes, so avoiding accidental merging.

index.html
```diff
@@ -261,6 +261,8 @@ <h1>Glossary</h1>
                 <dd>The bounding box of an input (most commonly, touch) on a digitizer. This typically refers to devices with coarser pointer input resolution than a single pixel. Some devices do not report this data at all.</dd>
             <dt><dfn>digitizer</dfn></dt>
                 <dd>A type of input sensing device in which a surface can detect input which is in contact and/or in close proximity. Most commonly, this is the surface that senses input from the touch contact or a pen stylus.</dd>
+            <dt><dfn>direct manipulation</dfn></dt>
+                <dd>Certain input devices (such as touchscreens) implement a "direct manipulation" metaphor where a pointer is intended to act primarily on the UI element it became active upon (providing a physical illusion of direct contact, instead of indirect contact via a cursor that conceptually floats above the UI), and where the pointer not only serves to interact with controls, but is also used to directly pan, scroll or zoom the current page when it is active (e.g. when a finger or stylus touches the touchscreen).</dd>
```
mustaqahmed: Let me try to reword your points and mention one difference from the indirect manipulation case for clarity (text selection):

> A pointer interaction metaphor where the input mechanism controlling the pointer (e.g. a finger or stylus touching a touchscreen) appears to be in a direct contact with an UI element on screen, and a cursor is not needed to mark the position of the pointer (unlike a mouse or trackpad).   Dragging direct manipulation pointers on screen usually cause a "sticky" effect with the UI elements under the pointers to convey the illusion of a direct physical contact, like panning/scrolling or zooming in such a way that the relative positions of the UI elements with respect to the pointers remain unchanged.  This is different from a mouse-drag (an indirect manipulation) that changes text selection highlight without moving the position of the UI element under the cursor.

Hmm, not sure if the use of "mechanism" vs "pointer" in my suggestion above look ambiguous enough!

index.html
```diff
@@ -957,94 +959,94 @@ <h2>The <code>touch-action</code> CSS property</h2>
...
 
-            <div class="note">As noted previously, in the case of user agents that allow default behaviors (such as panning or zooming) for other pointer types, these user agents MUST apply the same consideration for those pointer types. For instance, if a user agent allows panning/zooming with a stylus, the user agent must take into account the relevant <code>touch-action</code> value when determining which default behaviors it should handle.</div>
+            <div class="note">As noted previously, in the case of user agents that allow default behaviors (such as panning or zooming) for pointer types other than touch, these user agents MUST apply the same consideration for those pointer types. For instance, if a user agent allows panning/zooming with a stylus, the user agent must take into account the relevant <code>touch-action</code> value when determining which default behaviors it should handle.</div>
```
mustaqahmed: I think this old note about  "(input) other than touch" is no longer needed.  Wdyt?

mustaqahmed (commented):
This are my first pass comments.  I am still not done with the whole PR, will complete next week.

index.html
patrickhlauke: made another / similar stab at a rewording, by just giving some examples. i realise also that we're kind of using "direct manipulation" in a very narrow sense - as even some GUI interactions with a mouse are sometimes referred to as "direct manipulation" https://www.nngroup.com/articles/direct-manipulation/ ... which is why i tried to be a bit more direct and explicit here in my rewording by, essentially, calling a spade a spade and saying that touch/stylus are, but mouse isn't, for the purposes of our definition.

patrickhlauke:
any more thoughts on this?

index.html
mustaqahmed: Thanks for the reference, I now see the perspective behind the current definition.  I think this needs some discussion because IMHO this definition is a bit too broad to be able to differentiate between mouse and touch.

Let me summarize the two different definitions of "direct menipulation" we have mentioned here:

A. The nngroup page linked above defines "directness" exactly like [this Wikipedia page](https://en.wikipedia.org/wiki/Direct_manipulation_interface): if anything in the UI responds dynamically to a pointer, it qualifies as a direct manipulation, regardless of whether we have an explicit cursor (e.g. for mouse or trackpad nput) or not (onscreen touch or pen input).

B. In the issue discussion, the "directness" is like a classification of the pointing device: the input device can be "mouse-like" which is off-screen hence require an explicit cursor, or "touch-like" which comes in contact with the screen hence doesn't need a cursor.  In Chrome this is what we mean by direct manipulation.  I just discovered this ["guideline for developers"](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/ms698539(v=vs.85)) from Microsoft which exactly matches our interpretation.

My mental model for the two deifnitions is as follow: the "direct manipulation" in [A] needs the UI target to stick to an explicit or implicit cursor, while in [B] the UI target needs to stick to the input device (apparently of course).  So it seems [A] covers all pointer input devices while [B] is only for "on-screen" devices.

Given the long-term impact our spec definition could possibly have on the developer community, we have to be extra careful with the term "direct menipulation".

Or maybe we can find an alternate term to avoid all the confusion?  It's still not too late!

patrickhlauke:
tl;dr now wondering if instead of talking about things like `direct manipulation behaviors (such as panning or zooming)` it should be more `direct manipulation behaviors *for* panning or zooming`. this would aknowledge that there are different types of "direct manipulation", including the definition that also covers GUIs where a user can drag/drop elements with a mouse pointer, and hones in specifically on the direct manipulation for panning/zooming which *is* unique to the types of scenarios we mean (finger / stylus that is used to pan/zoom)

patrickhlauke:
Took another stab at making the language more comprehensible. Maybe went a bit overboard, but:

* to simplify matters, just settled on talking about "scrolling" rather than always clarifying it's "panning/scrolling". we do still have a note that says we're intending "pan" to mean the same as "scroll"
* as discussed in our call, instead of just saying "direct manipulation", specified, where necessary, that we're talking about "direct manipulation for scroll/zoom" or similar. this bridges the weird gap of having "direct manipulation" mean different things to different people (touch/stylus only for us, but in some quarters even a mouse GUI is called a "direct manipulation" interface ... for the latter though, scrolling/zooming is generally NOT done via this direct manipulation, so by being specific this now makes sense regardless of which way you interpret "direct manipulation")

this will still need a refactor/rebase/potential more work once https://github.com/w3c/pointerevents/pull/349 lands

patrickhlauke:
Did a rebase, tweaked some things along the way, and turned out I needed to force-push and rewrite history on this. Apologies. This should now however be in line with the recently merged stuff from the other PR.

index.html
```diff
@@ -908,10 +908,10 @@ <h2>Extensions to the `Navigator` interface</h2>
...
-        <p>For touch input, the default action of any and all pointer events MUST NOT be a manipulation of the viewport (e.g. panning or zooming).</p>
-        <div class="note">Touch manipulations are intentionally not a default action of pointer events. Removing this dependency on the cancellation of events facilitates performance optimizations by the user agent.</div>
-        <div class="note">While the issue of pointers used to manipulate the viewport is generally limited to touch input (where a user's finger can both interact with content and pan/zoom the page), certain user agents may also allow the same types of (direct or indirect) manipulation for other pointer types. For instance, on mobile/tablet devices, users may also be able to pan using a stylus. This section applies to these scenarios as well (despite the specification's use of "touch").</div>
+        <h1>Declaring candidate regions for default direct manipulation behaviors</h1>
+        <p>For <a>direct manipulation</a> interactions, the default action of any and all pointer events MUST NOT be a manipulation of the viewport (e.g. scrolling or zooming).</p>
```
smaug----: is it panning or scrolling? I'll let native English speakers to decide :)

index.html
```diff
@@ -925,99 +925,98 @@ <h2>The <code>touch-action</code> CSS property</h2>
                 <tr><th>Computed value:</th><td>Same as specified value.</td></tr>
             </table>
 
-            <p>The <code>touch-action</code> CSS property determines whether touch input MAY trigger the user agent's default behavior. This includes, but is not limited to, behaviors such as panning or zooming. See the section on <a href="#details-of-touch-action-values"><code>touch-action</code> values</a>.</p>
+            <p>The <code>touch-action</code> CSS property determines whether <a>direct manipulation</a> interactions (which are not limited to touch pointers, despite the property's name) MAY trigger the user agent's scrolling and zooming behavior. See the section on <a href="#details-of-touch-action-values"><code>touch-action</code> values</a>.</p>
```
smaug----: touch pointer? That sounds a bit odd and not used elsewhere, I think

smaug---- (commented):
Minor nit and the question about scrolling or panning.

index.html
patrickhlauke: we use(d) the terms interchangeably, but yes we should stick to one ... think when/if we merge the other PR for the new section (which messes things around more fundamentally) and rebase/merge it into this, we can revisit that. my preference is on panning

index.html
patrickhlauke: yup, good point. in my head, it was along the lines of "`touch` pointers" (as in the `pointerType`), but it reads awkward, you're right

index.html
mustaqahmed: After our WG meeting discussion today, I am now convinced that we don't need to explicitly isolate "mouse" from `touch-action` and therefore sticking the broader definition of "direct manipulation" works fine.

index.html
patrickhlauke: changed to just "touch"

patrickhlauke:
Made some further changes based on today's PEWG call - changed most instances of "scrolling" to "panning", and turned a few sentences around a bit to make them easier to read (hopefully without introducing any substantially new interpretations)

patrickhlauke:
Sorry, one more tweak (and now i'll leave this alone to give sufficient time to review) - after @mustaqahmed's comment in today's call, I revisited the glossary definition for "direct manipulation". I think I see what you meant now ... so turned it around a bit.

index.html
```diff
@@ -1349,6 +1347,11 @@ <h1>Glossary</h1>
...
             <dt><dfn>digitizer</dfn></dt>
                 <dd>A type of input sensing device in which a surface can detect input which is in contact and/or in close proximity. Most commonly, this is the surface that senses input from the touch contact or a pen stylus.</dd>
+            <dt><dfn>direct manipulation</dfn></dt>
+                <dd>Certain user agents (such as browsers on a touchscreen device) implement a "direct manipulation" metaphor where a pointer not only interacts with controls, but
+                is also used to directly pan or zoom the current page, providing the illusion of direct physical contact. As an example, users on a touchscreen device are generally
+                able to use a finger or a stylus to "grab" a page and pan it by moving the pointer, directly manipulating the page. Contrast this with a mouse pointer on a regular
+                desktop/laptop, where panning is done by using a scrollbar, rather than by "dragging" the page.</dd>
```
flackr: Note that on many laptops the touchpad provides a bit of a mixed experience, where you are physically manipulating (indirectly) the page, in the same way that an external stylus is also indirectly manipulating the page. The difference here seems to be that the touchpad touches are not sent directly to the page but produce actions through the mouse pointer.

TLDR; it may be worth mentioning that touchpads are similarly indirect, perhaps by pointing out that they only serve to produce actions through the indirect mouse pointer?

index.html
```diff
@@ -739,7 +739,7 @@ <h3>The <dfn><code>pointercancel</code> event</dfn></h3>
                     <li>The user agent has determined that a pointer is unlikely to continue to produce events (for example, because of a hardware event).</li>
                     <li>After having fired the <code>pointerdown</code> event, if the pointer is subsequently used to manipulate the page viewport (e.g. panning or zooming).
                      <div class="note">User agents can trigger panning or zooming through multiple pointer types (such as touch and pen),
-                     and therefore the start of a pan or zoom action may result in the cancellation of various pointers, including pointers with different pointer types.
+                     and therefore the start of a scroll or zoom action may result in the cancellation of various pointers, including pointers with different pointer types.
```
flackr: I think we decided we should go with pan consistently even though the two should be interchangeable?

index.html
```diff
@@ -755,7 +755,7 @@ <h3>The <dfn><code>pointercancel</code> event</dfn></h3>
                         <li>A device's screen orientation is changed while a pointer is active.</li>
                         <li>The user inputs a greater number of simultaneous pointers than is supported by the device.</li>
                         <li>The user agent interprets the input as accidental (for example, the hardware supports palm rejection).</li>
-                        <li>The user agent interprets the input as a pan or zoom gesture.</li>
+                        <li>The user agent interprets the input as a scroll or zoom gesture.</li>
```
flackr: Here too.

index.html
patrickhlauke: ah yeah, i believe touchpads fake scroll wheel events, if i'm not misremembering. yeah, will try and put in some wording to that effect, good catch

index.html
patrickhlauke: odd, i thought i had caught all instances of where i made a change and reverted them...good spot (here and in the other place)

index.html
mustaqahmed: I think in PE world, touchpad is mouse because it "shares" the mouse cursor on screen and appears as (pointerId=mouseId, pointerType="mouse").  Wheel events from trackpad form a completely separate event stream---same as a keyboard scroll---the mouse cursor position remains unchanged and doesn't affect any PointerEvent stream.  I mean we don't `pointercancel` any pointerId (eventTarget changes though).

flackr (approved):

smaug---- (approved):

PR #351, opened 2021-02-16
Add a SHOULD clause about pan gestures and what happens if a gesture …

### LINKED ISSUE #303: touch-action and scrolling directional lock
I’m looking into an issue reported against Safari in iOS 13 related to Pointer Events. I've created a [test case](https://gist.github.com/graouts/e7390caa34f08549f95a0ae8b3d10a93) that reduces the issue to a simple usage scenario where a page is designed to only ever scroll vertically but contains an interactive area where the user may scroll horizontally to go through various panels (think an image gallery). The interactive container can be scrolled by listening to pointer events and applying a CSS `transform` to reflect the pan gesture. That interactive container also has `touch-action: pan-y`.

In Safari on iOS 13, since the page scrolls vertically, the interactive area will get a `pointercancel` event during a horizontal panning gesture unless the user is careful not to stray from a very straight horizontal panning gesture.

However, in Chrome using the Pixel 2 XL in the Developer Tools, if the user starts a panning gesture that is predominantly horizontal, it seems to lock the gesture recognition to the horizontal direction and any change of direction while in flight will *not* trigger the page to scroll horizontally.

Additionally, in the same Chrome Pixel 2 XL setup, removing the `touch-action: pan-y` property from the interactive container will trigger a `pointercancel` event even when panning  predominantly in the horizontal direction. This doesn't happen in Safari on iOS 13 since the page *cannot* scroll horizontally.

This test shows some interoperability issues and I'd like to understand if these are due to bugs in Safari and/or Chrome, or because of some lacking specifications.

Specifically, I have the following questions:

1. Does Chrome apply a directional lock of sorts for scrolling such that starting panning in a predominantly horizontal direction prevents *any* scrolling in the vertical direction? Cc @NavidZ and @RByers from Google who may know more about this Chrome behavior.

1. If there is such a directional lock, is there a specification that defines this behavior? Under what conditions does that happen? From anecdotal testing, it would appear that, in Chrome, this is tied to the presence of a unidirectional panning restriction using the `touch-action` property.

1. If there *isn't* a specification defining this behavior, should there be additional behavior to let page authors enable such directional locks? (As an aside, iOS offers an API on `UIScrollView` has the [`directionalLockEnabled`](https://developer.apple.com/documentation/uikit/uiscrollview/1619390-directionallockenabled?language=objc) property that allows for just that.)

Hopefully all of the Chrome behaviors are already specified, and in this case I'd appreciate some pointers. Otherwise, I think we need to propose some ways to make such content interoperable.

patrickhlauke:
…suddenly changes direction

Closes https://github.com/w3c/pointerevents/issues/303

/cc @graouts

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/351.html" title="Last updated on Feb 17, 2021, 8:32 PM UTC (b672a19)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/351/d770a96...b672a19.html" title="Last updated on Feb 17, 2021, 8:32 PM UTC (b672a19)">Diff</a>


patrickhlauke:
Opted for a "SHOULD" rather than "MUST", as user agents may have a valid user-experience type reason why they choose not to do the effective "railing" of gestures once they start

patrickhlauke:
This is a fairly naive first pass. Happy for some further wordsmithing if needed / if unclear.

index.html
```diff
@@ -980,6 +980,7 @@ <h2>Determining supported touch behavior</h2>
...
+            <li>Similarly, in the case of the various <code>touch-action</code> values of <code>pan-*</code>, once the user agent has determined whether to handle an action directly or not at the start of the action, a subsequent change in the direction of the same gesture/action SHOULD be ignored by the user agent for as long as that pointer is active. For instance, if an element has been set to <code>touch-action: pan-y</code> (meaning that only vertical scrolling is handled by the user agent), and a gesture starts of horizontally, no vertical scrolling should occur if the user changes the direction of their action to be vertical at some point during the same gesture.</li>
```
mustaqahmed: Do you see any way to separate out action vs gesture?  The easiest way seems to be to replace "gesture/action" with "gesture" based on the interpretation that a gesture is a series of horizontal/vertical actions.  But may be others interpret the "action" vs "gesture" diff in some other way.  Thoughts?

mustaqahmed: Same idea: "for as long as that pointer is active" suggests a single gesture to me, so may be we can remove this phrase?  Or move it to somewhere else as a definition of a gesture?

mustaqahmed (commented):
Thanks Patrick for the PR, I agreed to take a look last year then missed it :disappointed:!

Your edit looks good to me.  I have two comments around how I interpret gesture vs action.  Please give it a thought.

index.html
patrickhlauke: i kind of interpreted action and gesture as pretty much interchangeable, admittedly. i might avoid confusion and just stick with "gesture" (though wasn't sure if that was a no-no word for us to use or not)

patrickhlauke: made a stab at a rewording, embracing the word "gesture" a bit more...

plehegar:
Because the sentence doesn't attempt to define a gesture but merely describe the consequence of a gesture on touch-action, I believe the pull request is fine.

patrickhlauke:
thanks @plehegar

mustaqahmed (approved):

mustaqahmed:
We need WPTs, perhaps by adding new actions for switched gesture directions in [pointerevents/pointerevent_touch-action*.html](https://wpt.fyi/results/pointerevents?label=master&label=experimental&aligned&view=subtest&q=pointerevent_touch-action).

mustaqahmed:
> [pointerevents/pointerevent_touch-action*.html](https://wpt.fyi/results/pointerevents?label=master&label=experimental&aligned&view=subtest&q=pointerevent_touch-action)

These WPTs enforce the expected behavior already, and in a way contradicting this PR.  My bad that I missed one important point with railing!  I will comment in the issue.

PR #354, opened 2021-03-10
gotpointercapture/lostpointercapture might never fire

liviutinta:
While working on wpt tests I found a case where the lostpointercapture event
was never firing because there was no other synthetic pointer event after
calling releasePointerCapture for that frame.
Looking into that I found about [pointerevents issue 32](https://github.com/w3c/pointerevents/issues/32) where it was decided gotpointercapture
and lostpointercapture will be delayed until the next pointer event.
This wasn't clear to me when I read the spec, so I thought I'll propose a
note to specifically call this out.

***
<a href="https://pr-preview.s3.amazonaws.com/liviutinta/pointerevents/pull/354.html" title="Last updated on Mar 10, 2021, 9:02 PM UTC (066d791)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/354/46346e9...liviutinta:066d791.html" title="Last updated on Mar 10, 2021, 9:02 PM UTC (066d791)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -694,6 +694,11 @@ interface PointerEvent : MouseEvent {
                         </li>
                         <li>Set the <dfn>pointer capture target override</dfn> to the <a>pending pointer capture target override</a>, if set. Otherwise, clear the <a>pointer capture target override</a>.</li>
                     </ol>
+                    <div class="note">Firing <code>gotpointercapture</code>, <code>lostpointercapture</code> is
+                        delayed until the next pointer event. If there are no pointer events after calls to
+                        <code>element.setPointerCapture(pointerId)</code> or
+                        <code>element.releasePointerCapture(pointerId)</code>, the events
+                        <code>gotpointercapture</code>, <code>lostpointercapture</code> will never fire.</div>
                 </section>
             </section>
         </section>


NavidZ:
I don't think this is right unless I misunderstood what you are saying.
The scenario you are referring is called [implicit release of pointer capture](https://w3c.github.io/pointerevents/#implicit-release-of-pointer-capture) which in turn should cause the lostpointercapture to fire (if the pointer was captured already). So if you don't see the lostpointercapture it might be a bug in Chromium.

liviutinta:
The wpt test I am referring to is [pointerevent_pointercapture_in_frame.html](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/web_tests/external/wpt/pointerevents/pointerevent_pointercapture_in_frame.html;drc=e5f599f2726269ef8984469488a3e6e1bdd0c7d2;l=144). The scenario where I observed the lostpointercapture event not firing:
- when setPointerCapture is called on pointerdown and releasePointerCapture on first pointermove: specifically when there would be no other pointer event for that frame (test 4: Test mousepointer capture in same-origin frame: Pointerdown with set capture at outerframe, then release on next pointermove). The lostpointercapture event is not lost all the time, in particular it happens under load. 

You are correct that I didn't explain correctly in the original post. I will update to better reflect the situation in the test.

NavidZ:
I see now what you are saying. I can imagine what causes this in our code. Although if I recall correctly @LanWei22  might have done some fixes around this. Note that this situation might differ in Chromium implementation when the iframe is OOPIF or whether you are running the input injection using test_driver as the test or just really manually moving mouse on the browser.

But regardless of the implementation and looking purely at the spec, I don't think it is clear that developers can get a gotpointercapture and never receive their lostpointercapture. We should strive to always match them without creating special cases.

liviutinta:
Thank you for clarifications @NavidZ. I am convinced now the scenario in which `gotpointercapture` or `lostpointercapture` events might not be fired because of a lack of a "next" pointer event that would execute [process pending pointer capture](https://www.w3.org/TR/pointerevents/#process-pending-pointer-capture) (not in the case of [implicit release of pointer capture](https://w3c.github.io/pointerevents/#implicit-release-of-pointer-capture)) will not happen for normal users.

Issue #215, opened 2017-08-10
Difficulty understanding getCoalescedEvents definition


gsnedders:
To me, the paragraph defining `getCoalescedEvents ` seems like a bit of mess. It makes a lot of statements without clearly delineating between them, and without defining what events get coalesced (and hence what properties they have are essentially undefined) and hence doesn't really define what the coalesced events even are. I'm also not clear on which sentences are declaring something normatively, and which are stating something informatively based on other definitions.

In short, I can understand each sentence individually, but the sum of them seems to leave a lot of behaviour undefined as far as I can tell (what are these events? what are their attributes?).

I can make a guess that we're meant to create the coalesced events ordinarily and then just change various attributes, but that's not clear. Or are we not changing them because we've coalesced them in such a way that they've already got those values?

I'd suggest something more like:

> Returns a sequence of all PointerEvents that were coalesced into the dispatched pointermove event. 
>
> Events must be coalesced into a single dispatched event grouped by [XXX], and the dispatched event's attributes set to a best representative summary of the coalesced events.
>
> Coalesced events must:
> * have the same pointerId, pointerType, isPrimary, isTrusted, target as the dispatched event,
> * have cancelable and bubbles set to false [NOTE: this is because they will never be dispatched], and
> * have all attributes mutated during [event dispatch](https://dom.spec.whatwg.org/#concept-event-dispatch) set to their default value. [NOTE: all related functions, including `stopPropagation` and `stopImmediatePropagation`, will do nothing as a result]
>
> The sequence must be in chronological order, sorted by `timeStamp`.

I think everything else in the paragraph is informative? You can add them back, but they should be clearly delineated from the normative requirements.

patrickhlauke:
@RByers et al, mind having a closer look at this proposed rewrite suggestion?

NavidZ:
@gsnedders the way you wrote it here sounds much better and clearer that it is now. Thanks. The only problem is that I'm not sure what we should use for the **XXX** part in _Events must be coalesced into a single dispatched event grouped by [XXX]_. The coalescing generally happens for all the events that happened since the last animation frame. However, if the main thread is busy or something they can also get coalesced even beyond that.

dtapuska:
You might want to refer to something in here: https://html.spec.whatwg.org/#event-loop-processing-model

gsnedders:
[...]

PR #364, opened 2021-04-18
Add new section explaining coalesced and predicted events

### LINKED ISSUE #361: Add section explaining what coalesced and predicted events actually *are* in layperson's terms
While the spec explains the functions and the lists of events, if doesn't have any kind of soft introduction/explanation of what these actually *are*, maybe with a few tiny graphics/examples. Explaining how user agent/OS does this coalescing/predicting internally, that particularly for drawing/high precision applications authors will want the non-coalesced events, how the predicted events can help with apparent responsiveness (e.g. "draw ahead to the predicted one, though be prepared to backtrack/redraw when the actual event comes in"). This should be a high-level section, like the current "Compatibility Mapping with Mouse Events" for instance.

patrickhlauke:
Closes https://github.com/w3c/pointerevents/issues/361 and https://github.com/w3c/pointerevents/issues/215

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/364.html" title="Last updated on May 13, 2021, 12:37 AM UTC (88ce9a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/364/0bd62f6...88ce9a0.html" title="Last updated on May 13, 2021, 12:37 AM UTC (88ce9a0)">Diff</a>


patrickhlauke:
would appreciate a pair of eyes whenever your get a chance (as can't merge without one approving review) - this should be fairly non-controversial

mustaqahmed:
This is awesome!  Thanks for taking care to add this.

I have a few minor comments/questions, mostly around pulling some more text into your new section:

A. In 5.1, the big text explanation right after `getCoalescedEvents` and `getCoalescedEvents` seems to belong to the new Section 11.  Can we move the big text into it and add a link to the new section in each of `getCoalescedEvents` and `getCoalescedEvents`?

B. The image about prediction needs to show the direction of drawing, otherwise it would be very confusing.  Also, can we perhaps zoom in one part of the curve where we have two non-overlapping prediction-lines?  ("Too many ~~cooks~~points spoil the ~~broth~~image!" :smile:)

C. This is subjective, but in an introduction I prefer very focused code examples (no `mousemove`, `maxTouchPoints`, "keyboard events" etc).

patrickhlauke:
> A. In 5.1, the big text explanation right after getCoalescedEvents and getCoalescedEvents seems to belong to the new Section 11. Can we move the big text into it and add a link to the new section in each of getCoalescedEvents and getCoalescedEvents?

Yeah, I was considering this initially, but now wondering - should the new section here for coalesced/predicted events actually come under the (now renumbered) section 4 as a subsection, similar to 4.1.1 button states? and then yes, move the big text into it, here.

> B. The image about prediction needs to show the direction of drawing, otherwise it would be very confusing. Also, can we perhaps zoom in one part of the curve where we have two non-overlapping prediction-lines?

Agree about indication the movement direction that caused these points. Have to admit, because I was using https://patrickhlauke.github.io/touch/coalesced-predicted-events/ it ended up being surprisingly difficult to "swoosh" a good representative gesture. I might have to tweak my source script and possibly just zoom in on the end of the gesture (where the predictions become a bit clearer/better than the often quite random predictions that Chrome seems to give me during a drawing operation currently).

> C. This is subjective, but in an introduction I prefer very focused code examples (no mousemove, maxTouchPoints, "keyboard events" etc).

fair point yes, especially since we already stress the need for considering keyboard etc elsewhere. will tweak

patrickhlauke:
> can we perhaps zoom in one part of the curve where we have two non-overlapping prediction-lines?

checking back on this, just to mention that in my demo, i don't actually draw any prediction lines per se, just the regular lines between confirmed and non-coalesced raw points. the predictions just show as dots (as otherwise i'd need to start tweaking my drawing logic to start looking at timestamps, and start drawing line segments from the last known good point to the next predicted point only, etc). I'll see if i can make it clearer though

patrickhlauke:
@mustaqahmed  I chose a much cleaner example image for the predicted points and clarified in the caption what the movement that led to that line was.

patrickhlauke:
@mustaqahmed I've gone with your proposed idea of moving the content about coalesced and predicted events into the section (plus made a few tiny tweaks). Initially tried moving that whole section back up (to be a sibling section to "Button states", "Primary pointer", etc), but it felt odd / far too wordy / got in the way of getting to the actual list of "Pointer Event Types". I *think* this way here works ok.

Going to set this back from draft to ready for QA

index.html
```diff
@@ -1085,6 +942,188 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+    <section>
+        <h1><dfn>Coalesced and predicted events</dfn></h1>
+
+        <div class="note">This specification does not define how user agents should coalesce or
+        predict pointer movement data. It only specifies the API for accessing this information.</div>
+
+        <section>
+            <h2>Coalesced events</h2>
+
+            <p>For performance reasons, user agents may choose not to send a move event every time
```
mustaqahmed: Perhaps better to say "...not to send a _pointermove_ event..."?

index.html
```diff
@@ -1085,6 +942,188 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+            the position of a pointer is updated. Instead, they will coalesce (combine/merge) multiple
+            position changes into a single, averaged move event. While this approach helps in reducing
+            the amount of event handling the user agent must perform, it will naturally reduce the
+            granularity and fidelity when tracking a pointer position, particularly for fast and large
+            movements. Using the <code>getCoalescedEvents</code> method it is possible for applications
```
mustaqahmed: Link `getCoalescedEvents` back to the event interface definition?

index.html
```diff
@@ -1085,6 +942,188 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+                <figcaption>Example of a curve in a drawing application – using only the coalesced
+                coordinates from <code>pointermove</code> events (the grey dots), the curve is noticeably
+                angular and jagged; the same line drawn using the more granular points provided by
+                <code>getCoalescedEvents</code> (the red circles) results in a smoother approximation
+                of the pointer movement.</figcaption>
+            </figure>
+
+            <p>A <a>PointerEvent</a> has an associated <dfn>coalesced event list</dfn> (a list of
+            zero or more <code>PointerEvents</code>). If this event is a <code>pointermove</code>
+            or <code>pointerrawupdate</code> event, it is a sequence of all <code>PointerEvent</code>
```
mustaqahmed: Nit: Replace "it" with "the list"?

index.html
```diff
@@ -1085,6 +942,188 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+                                paint(coalesced_event); // Paint all raw/non-coalesced points
+                            }
+                        } else {
+                            paint(e); // Paint the final coalesced point
+                        }
+                    });
+                
+                    function paint(event) {
+                        if(event.buttons>0)
+                        context.fillRect(event.clientX, event.clientY, 5, 5);
```
mustaqahmed: We need to indent this block.

index.html
```diff
@@ -1085,6 +942,188 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+
+        </section>
+
+        <section>
+            <h2>Predicted events</h2>
+            
+            <p>Some user agents have built-in algorithms which, after a series of confirmed pointer movements,
+            can make a prediction (based on past points, and the speed/trajectory of the movement) what
+            the position of future pointer movements may be. Applications can use this information with
+            the <code>getPredictedEvents</code> method to speculatively "draw ahead" to a predicted position
```
mustaqahmed: Like before: link `getPredictedEvents` back to the event interface definition?

index.html
```diff
@@ -1085,6 +942,188 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+                    // Draw the current predicted points to reduce the perception of latency.
+                    predicted_points = event.getPredictedEvents();
+                    for (let e of predicted_points) {
+                        drawPoint(e.pageX, e.pageY);
+                    }
+                });
+                </code>
+            </pre>
+                
+            <p>When a <a>PointerEvent</a> is created, run the following steps for each event in the
```
mustaqahmed, line 1091: The rest of Sec10.2 is related to both lists, so it seems we can start a new section here:
> 10.3 Targets of coalesced and predicted events

Then the algorithmic steps in Sec4.1 for `getCoalescedEvents` and `getPredictedEvents` entries can be moved here, to the bottom of (current 10.2) new 10.3.

After this, the definitions in 4.1 would be one liners (like "Returns the events held back by the user agent since last dispatch for performance reasons.  See section_ref" etc).

Please let me know your thoughts.  The name I proposed for the new 10.3 could be better but I have no other ideas at this moment.

index.html
mustaqahmed: Related: some of the existing links from `getCoalescedEvents` can point to the new section 10 or one of its subsections now.  I spotted two: in `pointermove` and in `pointerrawupdate` sections.

mustaqahmed (commented):
This is very nice and organized now, thanks again!  I never realized until now how the coalesced/predicted concepts were mingled with unrelated text before!!

I have one big suggestion only, about a new 10.3 below.  The rest are minor comments.

index.html
patrickhlauke: Good point, yes.

index.html
patrickhlauke: Good catch.

index.html
patrickhlauke: Done

index.html
patrickhlauke: I thought just doing `<a><code>getPredictedEvents</code></a>` would work, and I tried variations on this theme. based on what little i know of respec, this should work, (as there's a `<dfn>getPredictedEvents</dfn>`), but for some obtuse reason it's not and throws a respec error. Any ideas?

patrickhlauke: @marcoscaceres any ideas on why just using `<a>` here isn't working?

patrickhlauke: ah, never mind, counterintuitively enough, it needed specifying here somehow that the references were to this very spec here.. i.e. `<code><a data-lt="PointerEvent.getPredictedEvents">getPredictedEvents</a></code>` etc

index.html
patrickhlauke: Made a stab at this, but not 100% convinced it all makes sense now. Made a new "Populating coalesced and predicted event lists" section, and crammed everything in there.

One aspect of the algorithmic steps that makes no sense to me, now that i look at it...what happens if the dirty flag is false? what's the target of coalesced/predicted events supposed to be then?

index.html
```diff
@@ -1085,6 +924,213 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+            </ol>
+
+            <p>When the <code><a data-lt="PointerEvent.getPredictedEvents">getPredictedEvents</a></code> method is called:</p>
+            <ol>
+                <li>If the <a>predicted events targets dirty</a> is true:
+                for each event in the <a>predicted event list</a>, set the event's
+                <a href="https://dom.spec.whatwg.org/#event-target"><code>target</code></a> to
+                this <code>PointerEvent</code>'s <code>target</code>.
+                <li>Set the <a>predicted events targets dirty</a> to false.
+                <li>Return the <a>predicted event list </a>
```
mustaqahmed: Nit: missing period at line end (typo from past).

index.html
mustaqahmed: Hmm, to me having all "list maintenance" steps in one place is way better than before.

Re what happens when the dirty bit is not set, I think it means the event goes to where it was originally meant to go.  I agree this is not clear.  But better fix that separately, it's a separate problem.  Resolving the comment.

index.html
```diff
@@ -1085,6 +924,213 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+                    for (let e of predicted_points) {
+                        drawPoint(e.pageX, e.pageY);
+                    }
+                });
+                </code>
+            </pre>
+        </section>
+        
+        <section>
+            <h2><dfn>Populating the coalesced and predicted event lists</dfn></h2>
```
mustaqahmed: "Maintaining"?  Not sure!

mustaqahmed (approved):
LGTM.  There are some unclear things from existing text but let's do it separately.

index.html
patrickhlauke: Good spot

index.html
patrickhlauke: While a bit awkward, I tried to combine this ... "Populating and maintaining ...". Think that's workable...

index.html
```diff
@@ -359,31 +301,13 @@ <h2><code>PointerEvent</code> Interface</h2>
...
-                            <ol>
-                                <li>If the <a>coalesced events targets dirty</a> is true:
-                                for each event in the <a>coalesced event list</a>, set the event's
-                                <a href="https://dom.spec.whatwg.org/#event-target"><code>target</code></a>
-                                to this <code>PointerEvent</code>'s target.</li>
-                                <li>Set the <a>coalesced events targets dirty</a> to false.</li>
-                                <li>Return the <a>coalesced event list</a>.</li>
-                            </ol>
-                            </p>
+                            <p>Getter method for <a>coalesced events</a>.</p>
```
smaug----: Is Getter method some explicitly defined thingie? at least it isn't a getter in JS sense. Should this be "A method to access coalesced events." or some such

index.html
```diff
@@ -1085,6 +924,213 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+
+        <div class="note">This specification does not define how user agents should coalesce or
+        predict pointer movement data. It only specifies the API for accessing this information.</div>
+
+        <section>
+            <h2><dfn>Coalesced events</dfn></h2>
+
+            <p>For performance reasons, user agents may choose not to send a <code>pointermove</code>
+            event every time the position of a pointer is updated. Instead, they will coalesce
+            (combine/merge) multiple position changes into a single, averaged move event. While
```
smaug----: Not sure 'averaged' is what is actually happening. The event which is fired contains the most recent information about the position and such. It is just that some events which weren't fired may be put to the coalesced events list.

smaug---- (dismissed):

index.html
patrickhlauke: ah, good point. i can probably just remove the ", averaged" and be done with it (as this spec doesn't define how these coalesced/non-coalesced events are divined, only how they're accessed)

index.html
patrickhlauke: good point, i prefer that (was only going by the previous language that was there). makes it much more understandable.

patrickhlauke: changed to "A method that returns the list of ..." in both cases.

index.html
```diff
@@ -1085,6 +924,213 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+                <code>isTrusted</code>.</li>
+                <li>Set the event's <code>cancelable</code> and <code>bubbles</code> attributes to false.</li>
+                <li>Set the event's <a>coalesced event list</a> and <a>predicted event list</a> to an empty list.</li>
+                <li>Initialize the rest of the attributes the same way as <a>PointerEvent</a>.</li>
+            </ol>
+               
+            <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events targets dirty</dfn> and an associated
+            <dfn>predicted events targets dirty</dfn> flag. When an event is created they must be initialized to false.</p>
+                
+            <p>When the <a>PointerEvent</a>'s target is changed, set <a><code>coalesced events targets dirty</code></a>
```
flackr, line 1106: This seems like an implementation detail to say that the predicted and coalesced points' target must be the same as the main PointerEvent. Could we instead say that we always set the target for each coalesced /predicted event (and leave it as an implementation detail to do this lazily) or is it possible for a developer to observe when the target of these events change?

index.html
```diff
@@ -1085,6 +924,213 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+            </ol>
+
+            <p>When the <code><a data-lt="PointerEvent.getPredictedEvents">getPredictedEvents</a></code> method is called:</p>
+            <ol>
+                <li>If the <a>predicted events targets dirty</a> is true:
+                for each event in the <a>predicted event list</a>, set the event's
+                <a href="https://dom.spec.whatwg.org/#event-target"><code>target</code></a> to
+                this <code>PointerEvent</code>'s <code>target</code>.
+                <li>Set the <a>predicted events targets dirty</a> to false.
+                <li>Return the <a>predicted event list.</a>
```
flackr: I think you have to leave the period out of the link for this link to work.

index.html
patrickhlauke: Agree, this would simplify things here as well, bascially just boiling down to the statement that the target of coalesced and predicted events is always the same as its "parent" event, and that if the coalesced or predicted event was over another target, to ignore this but to set the relevant dirty flag. I'll think of a clean way to word this...

index.html
patrickhlauke: well spotted. corrected.

index.html
patrickhlauke: > is it possible for a developer to observe when the target of these events change?

it further makes me think: is the dirty flag only set when the `getCoalescedEvents` / `getPredictedEvents` methods get called (which is what the current text suggests), or is this happening as the original / high-level / coalesced pointer event gets fired?

additionally, i've never been quite clear...where do i "find" or access this flag? is this something internal to the browser itself and not meant to be actually exposed to JS authors? this concept of the flag seems rather vague to me...

patrickhlauke: (incidentally, we can either tackle this here in this PR, or make a separate issue/PR for this aspect ... whatever folks think is cleaner)

flackr: Yes, as far as I can tell the flag is an internal implementation optimization, which is why I suggest we remove it from the spec and just say that we always copy the current target to the coalesced / predicted points.

I'm okay with leaving this for a new PR since it's orthogonal to the changes being made here.

patrickhlauke: so purely from a JS author's point of view, what matters really is, in essence: all events in the coalesced event list and the predicted event list MUST have the same target as their "parent" pointer event (and the rest is left up to implementers on how that is achieved). maybe the whole flag business can be shuffled into a separate sub-section aimed at implementers here? (just to make the distinction clearer - at least mentioning the existence of the flag, but not going into any more detail)

patrickhlauke: would also be good to come up with some name for what i call "parent" pointer event above. not sure what the generally accepted naming for this sort of thing is...

patrickhlauke: @mustaqahmed @smaug---- any thoughts on these aspects? should we tackle them in a separate PR?

patrickhlauke:
I believe that, at least in part, this PR also addresses https://github.com/w3c/pointerevents/issues/215 - please see if there's anything else that's missed out/unclear

flackr (approved):

index.html
```diff
@@ -1085,6 +924,213 @@ <h3>Implicit Release of Pointer Capture</h3>
...
+                        }
+                    }
+                
+                &lt;/script&gt;</code>
+            </pre>
+            
+            <div class="note">The PointerEvent's attributes will be initalized in a way that best represents
+            all events in the coalesced event list.
+            For example its <a href="https://www.w3.org/TR/pointerlock/#widl-MouseEvent-movementX">
+            movementX</a> and
```
smaug----, line 1001: For non-movement, shouldn't coordinates be the most recent ones, not "best represents
all events in the coalesced event list"

patrickhlauke: i wonder if this is an implementation detail? i.e. could user agents choose to do coalescing in terms of averaging, rather than just picking the last/most recent one? do any do that? [edit: ah, i missed the "for non-movement" part]

patrickhlauke: Made a tiny tweak changing "all events" to "the events". Taking a step back though, I'm actually unclear on what this note is trying to say. Right from the start, "The PointerEvent's attributes" ... which pointer event is it referring to? The coalesced "parent" pointer event that has the list of non-coalesced pointer events? If so, is this trying to define (in a non-normative note, but still) what a user agent could do? The spec already says that it isn't trying to define this. Wonder if the whole note should not be removed? I think it's worth opening a new issue for this though, as this question is essentially about existing text (rather than about the reshuffling around of that into a new section)

patrickhlauke:
For the question relating to initialization (https://github.com/w3c/pointerevents/pull/364#discussion_r631141148 and https://github.com/w3c/pointerevents/pull/364#discussion_r631477745), I opened a separate issue https://github.com/w3c/pointerevents/issues/374

index.html
patrickhlauke: we'll tackle this separately from https://github.com/w3c/pointerevents/issues/370

PR #376, opened 2021-05-15
Clarify the rationale and purpose of touch-action

### LINKED ISSUE #366: Clarify/expand introduction to the touch-action section
Currently, the "Declaring candidate regions..." section starts with

>For touch input, the default action of any and all pointer events MUST NOT be a manipulation of the viewport (e.g. panning or zooming).
>
>NOTE
>Touch manipulations are intentionally not a default action of pointer events. Removing this dependency on the cancellation of events facilitates performance optimizations by the user agent.

This could do with being expanded a bit, as it may not be immediately obvious to a layperson what this *actually* means ... that cancelling a pointer event (with `preventDefault()` for instance) can't be used to stop/suppress scrolling or zooming, the same way that, say, this can be done with touch events. And that *that* is also the reason why you need to use `touch-action` instead to declaratively say whether or not the user agent can do scrolling/zooming or not (since you can't prevent it from the event handler).

Just an extra sentence or so that explains this would help make this fact more immediately clear, and give some context that then leads naturally into the meat of this section.

Suggest tackling this *after* https://github.com/w3c/pointerevents/pull/350 lands

### LINKED ISSUE #368: disambiguate default *action* and default *behavior* in prose
split from https://github.com/w3c/pointerevents/pull/349#issuecomment-811092269

[edit: modified this issue text to renumber the links to sections, for clarity, as the structure has changed since this was opened]

at the start of [8. Declaring candidate regions for default touch behaviors](https://w3c.github.io/pointerevents/#declaring-candidate-regions-for-default-touch-behaviors), it reads:

```
<p>For touch input, the default action of any and all pointer events MUST NOT be a manipulation of the viewport (e.g. panning or zooming).</p>
```

followed by the note:

```
Touch manipulations are intentionally not a default action of pointer events. Removing this dependency on the cancellation of events facilitates performance optimizations by the user agent.
```

It's unclear to me, as a reader, what this actually means/refers to. What concept is this trying to convey? I'm guessing it's essentially saying here that panning/zooming is special and not counted as a "default action" in the sense of [4.1.3.1 Attributes and Default Actions](https://w3c.github.io/pointerevents/#attributes-and-default-actions), and that for this reason panning/zooming also can't be cancelled by cancelling/preventing the event. This may need a bit more clarification/expansion. Perhaps even foreshadowing what we say later about `pointercancel` being fired. And then leading into the actual reason for this whole section: "instead, authors can control how user agents should handle panning/zooming using the `touch-action` CSS property instead".

That first sentence (about "default action ... MUST NOT be a manipulation of the viewport...") then seems at odds with the first note in [8.1 The touch-action CSS property](https://w3c.github.io/pointerevents/#the-touch-action-css-property)

```
As noted previously, in the case of user agents that allow default behaviors (such as panning or zooming) for other pointer types...
```

In that note, I think we need to drop the mention of "default behaviors" since we just said previously that panning/zooming isn't counted as a default behavior.

patrickhlauke:
- moved the note about viewport manipulation from the "Declaring candidate regions for default direct manipulation behaviors" section directly to the "Attributes and Default Actions" section, and made the text normative rather than informative (by not having it as a note, but actual part of the normative prose)
- expanded the sentence to more clearly convey what this means - by not having viewport manipulations treated/counted as a default action for pointer events, panning/zooming can't be prevented by cancelling a pointer event, and that instead authors must use `touch-action`
- tweaked the intro section to "Declaring candidate regions for default direct manipulation behaviors" to reiterate this concept/rationale
- removed one confusing instance of "default behavior" and changed to just "behavior" in a note in the touch-action section

Closes https://github.com/w3c/pointerevents/issues/366, closes https://github.com/w3c/pointerevents/issues/368

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/376.html" title="Last updated on May 19, 2021, 12:18 AM UTC (b1f888f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/376/284ec3f...b1f888f.html" title="Last updated on May 19, 2021, 12:18 AM UTC (b1f888f)">Diff</a>


index.html
```diff
@@ -756,8 +758,7 @@ <h2>Extensions to the `Navigator` interface</h2>
     </section>
     <section>
         <h1>Declaring candidate regions for default direct manipulation behaviors</h1>
```
mustaqahmed: Perhaps remove the "default" in this section header too?

mustaqahmed (approved):
This makes Sec8 well-connected with the rest of the spec at last.  Thanks for looking into this!

I checked all usage of "default...behavior".  I found two other spots where the word "default" could be removed: Sec1 para3 and (marked below) Sec8 header.  Please consider removing them if makes sense to you too.

index.html
patrickhlauke: good idea, done

PR #377, opened 2021-05-15
Simplify/clarify coalesced and predicted events

### LINKED ISSUE #215: Difficulty understanding getCoalescedEvents definition

### LINKED ISSUE #370: Clarify the coalesced/predicted targets dirty stuff, and clarify better how these events are generated/populated and when
It's currently a bit implicit (and probably not made super clear by the language we use), but my layperson's understanding is: it can happen that during a movement of the pointer, the pointer moves over various different targets. so in the case of the non-coalesced/raw events inside the coalesced event list, and the predicted events in the predicted event list, it can happen that those individual events belong/point to different targets. *but* when the `getCoalescedEvents` and `getPredictedEvents` methods are called, this gets sanitised so that even though the pointers were above different targets, they are all made to appear as if they were on the original target of the `pointermove`/`pointerrawmove` event that the method was called on.

patrickhlauke:
* clearly list out properly what properties the coalesced/predicted events should have (more closely matching the suggested structure from https://github.com/w3c/pointerevents/issues/215)
* remove the "Populating and maintaining the coalesced and predicted event lists" part, which is more confusing than anything if the various steps (with the dirty flag etc) are essentially just browser-internal flags, and where it doesn't really matter where/when the targets are changed as authors can't observe this change. leave it up to implementations how they want to do it. they can just decide wholesale to set the target to the event where the `getCoalescedEvents`/`getPredictedEvents` methods are being called on - see https://github.com/w3c/pointerevents/pull/364#discussion_r623925897

Closes https://github.com/w3c/pointerevents/issues/370, closes https://github.com/w3c/pointerevents/issues/215

Superseded by #395

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/377.html" title="Last updated on Jul 8, 2021, 2:44 PM UTC (707d9fd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/377/847813d...707d9fd.html" title="Last updated on Jul 8, 2021, 2:44 PM UTC (707d9fd)">Diff</a>


index.html
```diff
@@ -1057,10 +1069,21 @@ <h2><dfn>Predicted events</dfn></h2>
...
-            so the first event will have the smallest <code>timeStamp</code> which should still be greater than
-            the last event in the <a>coalesced event list</a>.</p>
+            <p>Predicted events MUST:</p>
+            <ul>
+                <li>have the same <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>,
+                    <a href="https://dom.spec.whatwg.org/#dom-event-istrusted"><code>isTrusted</code></a>,
+                    and <a href="https://dom.spec.whatwg.org/#event-target"><code>target</code></a> as
+                    the dispatched "parent" event</li>
+                <li>have <code>cancelable</code> and <code>bubbles</code> set to <code>false</code>
+                    (as these events will never be dispatched in isolation)</li>
```
smaug----, line 1079: Not sure (as these events will never be dispatched in isolation) is particularly useful

index.html
```diff
@@ -1085,48 +1108,6 @@ <h2><dfn>Predicted events</dfn></h2>
...
-                <li>Initialize the rest of the attributes the same way as <a>PointerEvent</a>.</li>
-            </ol>
-               
-            <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events targets dirty</dfn> and an associated
-            <dfn>predicted events targets dirty</dfn> flag. When an event is created they must be initialized to false.</p>
-                
-            <p>When the <a>PointerEvent</a>'s target is changed, set <a><code>coalesced events targets dirty</code></a>
-            and <a><code>predicted events targets dirty</code></a> to true.</p>
-
-            <p>When the <code><a data-lt="PointerEvent.getCoalescedEvents">getCoalescedEvents</a></code> method is called:</p>
```
smaug----, line 1126: How can we remove this? There is now no explanation what happens when getCoalescedEvents() is called.
Same with getPredictedEvents()

index.html
```diff
@@ -957,8 +957,20 @@ <h2><dfn>Coalesced events</dfn></h2>
...
             that were coalesced into this event; otherwise it is an empty list.</p>
 
-            <p>The events in the coalesced event list will have increasing
-            <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a>
+            <p>Coalesced events MUST:</p>
+            <ul>
+                <li>have the same <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>,
+                    <a href="https://dom.spec.whatwg.org/#dom-event-istrusted"><code>isTrusted</code></a>,
+                    and <a href="https://dom.spec.whatwg.org/#event-target"><code>target</code></a> as
+                    the dispatched "parent" event</li>
```
smaug----: 'parent' feels a bit weird. The old text tries to more programmatically initialize the events, that feels less error prone approach.

index.html
patrickhlauke: in the API definition it does say that these two methods return the relevant list https://w3c.github.io/pointerevents/#dom-pointerevent-getcoalescedevents / https://w3c.github.io/pointerevents/#dom-pointerevent-getpredictedevents - which seems all the info that's really needed?

this is based on the assumption that the "steps" we had before were just a browser-internal process to normalise the target, and it seems it's irrelevant when the target is actually set (either as part of the `getCoalescedEvents()` / `getPredictedEvents()` calls, or at the time these are gathered/generated initially) as long as the end result is indeed the same `target` as the main/parent/whatever we want to call it pointerevent, it seems  this can be removed?

smaug----: Hmm, the links aren't right in the API.
They should point to ' coalesced event list' and ' predicted event list'
And if *dirty flags aren't used, how do the events in the lists get retargeted when event is patched to shadow DOM?

patrickhlauke: the events will always have the target of the pointer event that the method was called on, if i understand it correctly. so whether the browser does them all in a oner (just goes through the list, sets the target for each event in the list to the caller/parent/whatever event's, then returns the list) or does some internal dirty flag setting to decide if this step is needed or not seems more like a browser-internal optimisation. unless i'm misunderstanding the shadow DOM aspect here?

and/or is the dirty flag used for anything else? where/how can authors check this flag? or is it something that the browser can use somewhere else internally? it might be that i've been overzealous in removing this, but mainly due to a lack of understanding what, in practice, the use of this dirty flag would even be

smaug----: Each time the target of the pointer event changes, also target of the events the get*Events() methods return need to change.
And while event is propagating through shadow + light DOM, the target may change several times.

patrickhlauke: i'm assuming (not knowing about the specific of how it's actually done in-engine) that for each pointer event that is dispatched, the lists are "fresh" and relating only to that specific event, and the target only would always be the target at the exact moment the event's method was called, i.e. (as @flackr asked in the other discussion) as an author you wouldn't be able to observe the retargeting, right? and then if so, it's more an implementation choice when/how the target for all events in the list that is returned when calling the methods is changed to match the target of the event where the method was called?

i mean i'm happy to reinclude the bit that i ripped out here, of course. but it still makes me wonder if it's anything specifically aimed at JS authors, or more of a "how user agents SHOULD implement this", or even "one way the MAY do it, but there may be other ways, as long as the end result is that the target of these events in the lists has the same target as the event the method was called on"

patrickhlauke:
Setting this to draft, as me ripping out the "Populating and maintaining..." https://w3c.github.io/pointerevents/#populating-and-maintaining-the-coalesced-and-predicted-event-lists section may have been premature. Admittedly, this may be due to me not grokking the central point the section tries to make. My most immediate questions are here https://github.com/w3c/pointerevents/issues/370#issuecomment-842088123

Beyond that, I *think* it would still be valuable to keep those bullet lists of the various properties of coalesced/predicted events in their respective sections though, as they outline what a JS author can expect from those events. But maybe the "Populating..." part should also be kept, for user agents.

index.html
smaug----: The flag makes the algorithm rather clear. And in fact, the proposed spec change isn't doing exactly the same as the current spec. Currently if one calls getCoalescedEvents() (just once) and then keeps references to those events and checks their target, .target never changes, but with the proposed change it is unclear whether the target does change at some point if the main event propagates through shadow DOM.

patrickhlauke: but how do you keep a reference to one of the coalesced/predicted events in the list? Aren't they fungible? As soon as a new pointerevent is actually fired, it'd have a whole set of new coalesced/predicted events, no? or are they sort of recycled/kept?

patrickhlauke:
thanks to @smaug---- for taking time today to have a chat about this with me. I'll mull things over some more and propose some further changes that will hopefully steer this in a cleaner direction without losing anything

index.html
patrickhlauke: kept the word `"parent"` in quotes, but expanded to make it clear what is meant (`dispatched "parent" pointer event that the <code>getCoalescedEvents()</code> method was called on`)

patrickhlauke:
@smaug---- i rejigged the sections with the "steps" for the UA, hopefully addressing what we discussed the other day about the need to explicitly say that the retargeting/setting the target needs to happen when the methods are called. also, tried to explain a bit better what the colloquial "parent" I used means there. let me know if this is better, then i can set this back to ready for review

smaug---- (approved):
I guess this is fine. We'll need to tweak this a tiny bit, probably, when fixing constructor handling (if one passes coalesced events to the constructor.)

index.html
```diff
@@ -961,13 +961,26 @@ <h2><dfn>Coalesced events</dfn></h2>
...
 
-            <p>The events in the coalesced event list will have increasing
-            <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a>
+            <p>When the <code>getCoalescedEvents()</code> method is called, user agents MUST apply the following
+               steps to each event in the coalesced events list, before returning the list:</p>
+            <ul>
+                <li>set <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>,
+                    <a href="https://dom.spec.whatwg.org/#dom-event-istrusted"><code>isTrusted</code></a>,
+                    and <a href="https://dom.spec.whatwg.org/#event-target"><code>target</code></a> to match the
+                    dispatched "parent" pointer event that the <code>getCoalescedEvents()</code> method was called on</li>
```
flackr, line 979: In the points which are coalesced I think the only thing that could vary is the target (as the pointer moves over different targets). We should never coalesce events from other pointers into a different pointer's event. As such, I think we only need to call out target here.

index.html
```diff
@@ -961,13 +961,26 @@ <h2><dfn>Coalesced events</dfn></h2>
...
+                <li>set <code>cancelable</code> and <code>bubbles</code> to <code>false</code>
+                    (as these events will never be dispatched in isolation)</li>
+                <li>set <a>coalesced events list</a> and <a>predicted events list</a> to an empty list</li>
```
flackr, line 982: Does this mean the developer can't get the predicted events if they call getCoalescedEvents first? This seems surprising. Why do we need to set an empty list when the user calls getCoalescedEvents? It seems like they should be able to get the same coalesced event list multiple times if they choose to. Otherwise, you could have multiple event listeners conflict with each other if one inadvertently clears the coalesced event list.

index.html
```diff
@@ -1059,17 +1072,29 @@ <h2><dfn>Predicted events</dfn></h2>
...
+            <p>When the <code>getPredictedEvents()</code> method is called, user agents MUST apply the following
+               steps to each event in the predicted events list, before returning the list:</p>
+            <ul>
+                <li>set <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>,
+                    <a href="https://dom.spec.whatwg.org/#dom-event-istrusted"><code>isTrusted</code></a>,
+                    and <a href="https://dom.spec.whatwg.org/#event-target"><code>target</code></a> to match the
+                    dispatched "parent" pointer event that the <code>getPredictedEvents()</code> method was called on</li>
+                <li>set <code>cancelable</code> and <code>bubbles</code> to <code>false</code>
+                    (as these events will never be dispatched in isolation)</li>
+                <li>set <a>coalesced events list</a> and <a>predicted events list</a> to an empty list</li>
```
flackr, line 1104: Same comments above apply here.

index.html
patrickhlauke: these are the coalesced and predicted lists properties of the coalesced/predicted events. otherwise, you get some weird recursion type stuff here... "the events that were coalesced to get the coalesced event"?

patrickhlauke: would adding "set ... for the event to an empty list" help or seem redundant, as these are already steps scoped to "apply the following  steps to each event in the coalesced events list" ?

index.html
patrickhlauke: but then it would need to say somewhere (else?) that the coalesced / predicted events have the same id/type/etc as their "parent" event that the method was called on (otherwise authors don't know what those values would be set to?)

patrickhlauke: adding here that, from what @smaug---- was saying, in the Gecko implementation the non-coalesced "raw" events are not stored as full-fledged pointer events internally, but some more general kind of event. only when `getCoalescedEvents` is called are they turned into pointer events for the list that is then returned. so probably from that point of view, these steps make more sense. if an engine instead already stores them up as if they were real pointer events, and then only needs to ensure that they're retargeted to the "parent" pointer event where the method is actually called, then yes only retargetting step here would be needed. but then we may need to split out somehting about all the other properties? a separate list that explains what authors can expect `pointerId` etc to be (that they'll also match the "parent" event)

flackr: > but then it would need to say somewhere (else?) that the coalesced / predicted events have the same id/type/etc as their "parent" event that the method was called on (otherwise authors don't know what those values would be set to?)

This seems reasonable to add. We already state that "The events in the coalesced event list will have increasing timeStamps," we could add to this that those other properties will be the same as the PointerEvent (the root one).

> adding here that, from what @smaug---- was saying, in the Gecko implementation the non-coalesced "raw" events are not stored as full-fledged pointer events internally, but some more general kind of event. only when `getCoalescedEvents` is called are they turned into pointer events for the list that is then returned. so probably from that point of view, these steps make more sense. if an engine instead already stores them up as if they were real pointer events, and then only needs to ensure that they're retargeted to the "parent" pointer event where the method is actually called, then yes only retargetting step here would be needed. but then we may need to split out somehting about all the other properties? a separate list that explains what authors can expect `pointerId` etc to be (that they'll also match the "parent" event)

The spec already seems to imply that the PointerEvent keeps a list of full PointerEvents in the coalesced events list, "A PointerEvent has an associated coalesced event list (a list of zero or more PointerEvents)." Implementations can of course optimize and build this on the fly as long as the side effects (i.e. calls to getCoalescedEvents) return a list that is consistent with this assumption. That does raise the question of whether you should get the same list of events with multiple calls - I'd say the spec sort of implies that you should right now, but perhaps it shouldn't.

index.html
```diff
@@ -961,13 +961,26 @@ <h2><dfn>Coalesced events</dfn></h2>
...
 
-            <p>A <a>PointerEvent</a> has an associated <dfn>coalesced event list</dfn> (a list of
+            <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events list</dfn> (a list of
             zero or more <code>PointerEvent</code>s). If this event is a <code>pointermove</code>
             or <code>pointerrawupdate</code> event, the list is a sequence of all <code>PointerEvent</code>s
             that were coalesced into this event; otherwise it is an empty list.</p>
 
-            <p>The events in the coalesced event list will have increasing
-            <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a>
+            <p>When the <code>getCoalescedEvents()</code> method is called, user agents MUST apply the following
```
mustaqahmed, line 974: I think it's easier if we "finalize" both lists at the moment the event was dispatched, instead of the moment when `get*Events()` is called.  The target is known at dispatch time, so do we need to delay the target fixup step at all?

If there is a reason to delay it (performance?), then the first call to any of `get*Events()` can finalize both lists to avoid complications.

flackr: I don't think the spec should say when this happens, and just say that the coalesced events will have the same target.

However, I think an implementation is free to do this whenever it chooses. It shouldn't be developer observable when this happens.

flackr: Nevermind, per discussion it is observable when the target changes so we should retarget here.

patrickhlauke:
reminder for @flackr https://www.w3.org/2021/05/26-pointerevents-minutes.html#a01

patrickhlauke:
> reminder for @flackr https://www.w3.org/2021/05/26-pointerevents-minutes.html#a01

happy to bash out some of this in a comment here as well if you're not inspired, and can then wordsmith it some more

index.html
flackr: Thanks for the explanation, I think this is fine as is and I just failed to read the context correctly.

patrickhlauke:
As this has become too unwieldy and out of sync with other work that has happened since, closing this - superseded by https://github.com/w3c/pointerevents/pull/395

PR #379, opened 2021-05-23
Expand explanation for non-coalesced events

### LINKED ISSUE #375: Do user agents only coalesce pointermove events relating to changes in position?
https://w3c.github.io/pointerevents/#coalesced-events states that

> For performance reasons, user agents may choose not to send a `pointermove` event *every time the position of a pointer is updated.*

Does this really only apply to *position* updates? Or do changes in other attributes ("i.e. button state, coordinates, pressure, tangential pressure, tilt, twist, or contact geometry", to borrow from the `pointerrawupdate` description) also get coalesced (and the non-coalesced events would have the more granular changes from those attributes)?

patrickhlauke:
- a user agents may coalesce events not just if the position is updated, but any of the attributes (as it's not just position changes that can result in an overwhelmingly large number of events)
- tweaked the matching sentence in the `pointerrawupdate` slightly, to use `pointer's attributes` rather than the awkward sounding `pointing device attributes`
- added glossary definition for "measurable properties"
- minor: `in contrast with` > `in contrast to`

Closes https://github.com/w3c/pointerevents/issues/375

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/379.html" title="Last updated on Jul 7, 2021, 5:53 PM UTC (f3693ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/379/cdb4a64...f3693ac.html" title="Last updated on Jul 7, 2021, 5:53 PM UTC (f3693ac)">Diff</a>


index.html
```diff
@@ -1410,6 +1412,15 @@ <h1>Glossary</h1>
                 <div class="note">In some cases, touchpads (like those found on a laptop) will allow the user to scroll by "dragging" on the touchpad. However, this is generally achieved
                 by the touchpad generating "fake" mouse wheel events, so this wouldn't count as a direct manipulation.</div>
                 </dd>
+            <dt><dfn>discrete and continuous properties</dfn></dt>
+                <dd><p>Discrete properties are those properties that can only have a finite set of values. In the context of pointer events, <code>pointerId</code>,
+                    <code>pointerType</code>, <code>isPrimary</code>, and the [[UI-EVENTS]] Mouse Event model properties <code>button</code>, <code>buttons</code>,
+                    <code>ctrlKey</code>, <code>shiftKey</code>, <code>altKey</code>, and <code>metaKey</code> are discrete.</p>
+                    <p>Continuous properties are those properties whose value can be expressed as a real number over a specific range. For pointer events,
```
mustaqahmed: Wondering how we can avoid the fact that `tiltX`, `tiltY` and `twist` have type `long` which is (mathematically speaking) a finite set matching the definition of "discrete" above!

Looking at the [Wikipedia entry](https://en.wikipedia.org/wiki/Continuous_or_discrete_variable), it seems to me "measuring" vs "counting" perspective could help, something like this:

> A continuous property represents some measured value expressed using a real number or an integer from a large domain...

index.html
```diff
@@ -1410,6 +1412,15 @@ <h1>Glossary</h1>
                 <div class="note">In some cases, touchpads (like those found on a laptop) will allow the user to scroll by "dragging" on the touchpad. However, this is generally achieved
                 by the touchpad generating "fake" mouse wheel events, so this wouldn't count as a direct manipulation.</div>
                 </dd>
+            <dt><dfn>discrete and continuous properties</dfn></dt>
+                <dd><p>Discrete properties are those properties that can only have a finite set of values. In the context of pointer events, <code>pointerId</code>,
+                    <code>pointerType</code>, <code>isPrimary</code>, and the [[UI-EVENTS]] Mouse Event model properties <code>button</code>, <code>buttons</code>,
+                    <code>ctrlKey</code>, <code>shiftKey</code>, <code>altKey</code>, and <code>metaKey</code> are discrete.</p>
```
mustaqahmed: During last PEWG meeting, I proposed mentioning `ctrlKey`, `shiftKey`, `altKey`, and `metaKey` as "discrete properties" but didn't realize it would be problematic from `PointerEvent` perspective: a `PointerEvent` (or `MouseEvent`) is fired in response to a button-change on a pointing device, and never in response to a key-change (or "button-chance on a keyboard").  If we mention these four as discrete properties, it could be interpreted as, say, "a change in `ctrlKey` fires a `PointerEvent`".

Sorry for the oversight!

Not mentioning them here solves the problem.  Or we can say these are not pointer properties but just current states of the keyboard.  Wdyt?

patrickhlauke: as it would seem odd to not mention those properties, since they *are* discrete (and it would look more like a mistaken omission if we left them out), i've expanded the sentence where it talks about when `pointermove` may not be sent to explicitly single them out

```
<p>For performance reasons, user agents may choose not to send a <code>pointermove</code>
            event every time a <a data-lt="discrete and continuous properties">continuous property</a>
            (such as coordinates, pressure, tangential pressure, tilt, twist, or contact geometry) or a
            <a data-lt="discrete and continuous properties">discrete property</a> relating to keyboard
            modifiers (<code>ctrlKey</code>, <code>shiftKey</code>, <code>altKey</code>,
            and <code>metaKey</code> from the [[UI-EVENTS]] Mouse Event model) of a pointer
            is updated.
```

makes the sentence a bit long, but more accurate i think. and more logical to mention this there, rather than muddying up (by omission) the definition of what is and isn't discrete

index.html
patrickhlauke: Ah, this does throw a spanner in the works overall. Forgot that `tiltX`, `tiltY`, `twist` are `long` ... darn it. This almost calls into question the whole PR, as i'm not sure we want to redefine the common terms of "continuous" and "discrete" to bend them to "our" convenience.

patrickhlauke: perhaps awkward, but i pivoted to defining "measurable properties" instead, and re-jigged the explanation for coalesced events accordingly. see if this makes sense.

patrickhlauke:
@flackr @smaug---- if you could have a look at this so far. it's not super elegant, but hopefully accurate at least. open to any better suggestions too of course.

index.html
```diff
@@ -941,8 +941,12 @@ <h1><dfn>Coalesced and predicted events</dfn></h1>
             <h2><dfn>Coalesced events</dfn></h2>
 
             <p>For performance reasons, user agents may choose not to send a <code>pointermove</code>
-            event every time the position of a pointer is updated. Instead, they will coalesce
-            (combine/merge) multiple position changes into a single move event. While
+            event every time a <a data-lt="measurable properties">measurable property</a>
+            (such as coordinates, pressure, tangential pressure, tilt, twist, or contact geometry) or a
+            property relating to keyboard modifiers (<code>ctrlKey</code>, <code>shiftKey</code>,
+            <code>altKey</code>, and <code>metaKey</code> from the [[UI-EVENTS]] Mouse Event model)
```
mustaqahmed: Because no pointer event is fired for changes in `*Key` modifiers (see the comment in the glossary), let's drop them from here too.  (Sorry again for suggesting them initially).

mustaqahmed (commented):
The glossary entry looks good to me but please wait for a second opinion because I may be biased towards "measurable".

index.html
patrickhlauke: doh, you're right, no need to even mention them here anymore.

mustaqahmed (approved):
LGTM, thanks,

I would still suggest a second review in case I am too biased towards "measurable".

index.html
```diff
@@ -1419,6 +1421,17 @@ <h1>Glossary</h1>
                 </dd>
             <dt><dfn>hit test</dfn></dt>
                 <dd>The process by which a user agent determines a target element for a pointer event. Typically, this is determined by considering the pointer's location and also the visual layout of elements in a document on screen media.</dd>
+            <dt><dfn>measurable properties</dfn></dt>
+                <dd><p>Measurable properties represent values relating to pointer sensor data that is expressed using a real number or an
+                    integer from a large domain. For pointer events, <code>width</code>, <code>height</code>, <code>pressure</code>,
```
smaug----, line 1423: what does "a large domain" mean exactly here, or does it really matter here?

smaug---- (approved):
I would possibly just drop that "domain" thing

index.html
patrickhlauke: if something only had a handful of values (like...`button` and `buttons` for instance, which use a discrete set of integer values https://w3c.github.io/pointerevents/#button-states), it wouldn't be "measurable" in this definition, but countable/discrete.

the domain/large domain tries to square the circle of "yes, `tiltX`/`tiltY` are integers, and yes they can only have a specific discrete number of values...but we do treat them as measurable because that number of values is so large"

if we dropped domain/large domain, `button` and `buttons` would suddenly also count as "measurable" (unless we then want to argue "ah, but they don't relate to sensor data", but then...what's a sensor? etc)

index.html
```diff
@@ -1419,6 +1421,17 @@ <h1>Glossary</h1>
...
+                    <code>tangentialPressure</code>, <code>tiltX</code>, <code>tiltY</code>, <code>twist</code>,
+                    <code>altitudeAngle</code>, <code>azimuthAngle</code>, and the [[UI-EVENTS]] Mouse Event model properties
+                    <code>screenX</code>, <code>screenY</code>, <code>clientX</code>, <code>clientY</code> are measurable properties.</p>
+                    <p>In contrast<code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code>, and the
```
mustaqahmed: Need a space after "contrast".

patrickhlauke: well spotted. fixed.

smaug---- (approved):

index.html
```diff
@@ -942,8 +941,10 @@ <h1><dfn>Coalesced and predicted events</dfn></h1>
             <h2><dfn>Coalesced events</dfn></h2>
 
             <p>For performance reasons, user agents may choose not to send a <code>pointermove</code>
-            event every time the position of a pointer is updated. Instead, they will coalesce
-            (combine/merge) multiple position changes into a single move event. While
+            event every time a <a data-lt="measurable properties">measurable property</a>
+            (such as coordinates, pressure, tangential pressure, tilt, twist, or contact geometry)
+            of a pointer is updated. Instead, they will coalesce (combine/merge) multiple changes into
```
flackr: I think this should say "Instead, they may coalesce".

index.html
flackr: We could argue that the underlying data for tiltX/tiltY is a real number whereas buttons is inherently an integer value.

flackr (approved):

index.html
patrickhlauke: good point yes, that would gel with the "may" in the previous sentence. will make that change in a sec.

index.html
patrickhlauke: perhaps...to square this particular circle (and remembering your original idea/mention  of "continuous" values) this could be modified along the lines of `Measurable properties represent values relating to continuous pointer sensor data`, or `Measurable properties represent continuous values relating to pointer sensor data`. thoughts? this would again clarify with button/buttons aren't there...

patrickhlauke:
Agreed on call to add "continuous" as per https://github.com/w3c/pointerevents/pull/379#discussion_r665399239 (first option) and merge.

index.html
patrickhlauke: done

PR #381, opened 2021-05-27
Refreshing references using latest respec magic

plehegar:
Note that "secure context" now comes from HTML (instead of Secure Contexts)

It's using more data-cite than I'd like.

HTML exports missing?
- html/#task-queue
- html/#initiate-the-drag-and-drop-operation
- html/#event-loop-processing-model | animation frame callbacks:  not sure if we need a hook for step 12 in https://html.spec.whatwg.org/#event-loop-processing-model

once this gets reviewed, I'll open bugs in whatwg/html for the needed exports.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/381.html" title="Last updated on Jun 10, 2021, 9:15 AM UTC (09aa1f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/381/ae7cd4e...09aa1f4.html" title="Last updated on Jun 10, 2021, 9:15 AM UTC (09aa1f4)">Diff</a>


index.html
```diff
@@ -335,8 +338,7 @@ <h2><code>PointerEvent</code> Interface</h2>
                     the default value.</p>
                     <p>When the user agent calculates <code>tiltX</code> / <code>tiltY</code> from
                     <code>azimuthAngle</code> / <code>altitudeAngle</code> it SHOULD round the final integer values using
-                    <a href="https://tc39.es/ecma262/#sec-math.round"><code>Math.round</code></a> ([[ECMA-262]]) rules.</p>
-                    <p></p>
+                    <a data-cite="ECMASCRIPT/numbers-and-dates.html#sec-math.round">Math.round</a> ([[ECMASCRIPT]]) rules.</p>
```
marcoscaceres: If it's possible, it might be safer to go with: 
 
```suggestion
                    <a data-cite="ECMASCRIPT#sec-math.round">Math.round</a> ([[ECMASCRIPT]]) rules.</p>
```

In case they move the definition elsewhere in the future.

index.html
```diff
@@ -495,9 +497,9 @@ <h3>Attributes and Default Actions</h3>
...
+                        <p>Similar to {{MouseEvent}} [[UI-EVENTS]] the <code>relatedTarget</code> should be initialized to the element whose bounds the pointer just left (in the case of a <code>pointerover</code> or <code>pointerenter</code> event) or the element whose bounds the pointer is entering (in the case of a <code>pointerout</code> or <code>pointerleave</code>). For other pointer events, this value will default to null. Note that when an element receives the pointer capture all the following events for that pointer are considered to be inside the boundary of the capturing element.</p>
```
marcoscaceres: `relatedTarget` is on {{FocusEvent}}, so:

(as an aside... "should be initialized" seems dubious here... wouldn't it be must?)

```suggestion
                        <p>Similar to {{MouseEvent}}, the {{FocusEvent/relatedTarget}} should be initialized to the element whose bounds the pointer just left (in the case of a <code>pointerover</code> or <code>pointerenter</code> event) or the element whose bounds the pointer is entering (in the case of a <code>pointerout</code> or <code>pointerleave</code>). For other pointer events, this value will default to null. Note that when an element receives the pointer capture all the following events for that pointer are considered to be inside the boundary of the capturing element.</p>
```

index.html
```diff
@@ -967,8 +969,8 @@ <h2><dfn>Coalesced events</dfn></h2>
             that were coalesced into this event; otherwise it is an empty list.</p>
 
             <p>The events in the coalesced event list will have increasing
-            <a href="https://dom.spec.whatwg.org/#dom-event-timestamp"><code>timeStamps</code></a>
-            ([[!WHATWG-DOM]]), so the first event will have the smallest <code>timeStamp</code>.</p>
+            {{Event/timeStamp}}s
+            ([[!WHATWG-DOM]]), so the first event will have the smallest {{Event/timeStamp}}.</p>
```
marcoscaceres: General note to the Editors: It's best not to sprinkle [[SPEC]] everywhere. Sometimes things get moved to other specs, so it's best just to let ReSpec figure out where something is coming from (it will put the right thing in the bibliography). 


```suggestion
            {{Event/timeStamp}}s, so the first event will have the smallest {{Event/timeStamp}}.</p>
```

marcoscaceres:
> html/#task-queue

Depending on usage, it might be best not to export "html/#task-queue", but the specific task queue should be exported instead (if it's not already!).

> html/#event-loop-processing-model | animation frame callbacks: not sure if we need a hook for step 12 in
 
Yeah, that might need some discussion with the HTML folks - or something might not be layered correctly. I didn't look tho.

patrickhlauke:
anything we need/can do to move this forward?

marcoscaceres:
I think you could merge it and we can fix up more things as a followup.

patrickhlauke:
@plehegar leaving it up to you if you want to merge now and open a separate issue/reminder to follow up, or just leaving this open

PR #390, opened 2021-07-06
Update targets of predicted and coalesced events when trusted event target changes.

flackr:
This allows removing the dirty flag, as trusted events keep the target of the PointerEvent in sync with their coalesced and predicted events.

***
<a href="https://pr-preview.s3.amazonaws.com/flackr/pointerevents/pull/390.html" title="Last updated on Jul 6, 2021, 7:02 PM UTC (ae66c7c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/390/cdb4a64...flackr:ae66c7c.html" title="Last updated on Jul 6, 2021, 7:02 PM UTC (ae66c7c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1104,13 +1104,13 @@ partial interface Navigator {
 
         <section>
             <h2><dfn>Populating and maintaining the coalesced and predicted event lists</dfn></h2>
-            <p>When a trusted <a>PointerEvent</a> is created, run the following steps for each event in the
+            <p>When a <a>PointerEvent</a> is created, run the following steps for each event in the
             <a>coalesced event list</a> and <a>predicted event list</a>:</p>
             <ol>
-                <li>Set the event's <code>pointerId</code>, <code>pointerType</code>, <code>target</code>,
+                <li>Set the event's <code>pointerId</code>, <code> pointerType</code>,
                 <code>isPrimary</code> and <code>isTrusted</code> to the {{PointerEvent}}'s
-                <code>pointerId</code>, <code>pointerType</code>, <code>target</code>,
-                <code>isPrimary</code> and <code>isTrusted</code>.</li>
+                <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code> and
+                <code>isTrusted</code>.</li>
                 <li>Set the event's <code>cancelable</code> and <code>bubbles</code> attributes to false.</li>
                 <li>Set the event's <a>coalesced event list</a> and <a>predicted event list</a> to an empty list.</li>
                 <li>Initialize the rest of the attributes the same way as <a>PointerEvent</a>.</li>
--- a/index.html
+++ b/index.html
@@ -1104,40 +1104,24 @@ partial interface Navigator {
 
         <section>
             <h2><dfn>Populating and maintaining the coalesced and predicted event lists</dfn></h2>
-            <p>When a <a>PointerEvent</a> is created, run the following steps for each event in the
+            <p>When a trusted <a>PointerEvent</a> is created, run the following steps for each event in the
             <a>coalesced event list</a> and <a>predicted event list</a>:</p>
             <ol>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1104,40 +1104,24 @@ <h2><dfn>Predicted events</dfn></h2>
...
-            <p>When a <a>PointerEvent</a> is created, run the following steps for each event in the
+            <p>When a trusted <a>PointerEvent</a> is created, run the following steps for each event in the
             <a>coalesced event list</a> and <a>predicted event list</a>:</p>
             <ol>
-                <li>Set the event's <code>pointerId</code>, <code> pointerType</code>,
+                <li>Set the event's <code>pointerId</code>, <code>pointerType</code>, <code>target</code>,
                 <code>isPrimary</code> and <code>isTrusted</code> to the {{PointerEvent}}'s
-                <code>pointerId</code>, <code>pointerType</code>, <code>isPrimary</code> and
-                <code>isTrusted</code>.</li>
+                <code>pointerId</code>, <code>pointerType</code>, <code>target</code>,
```
smaug----: I wonder how precise we want to be here. When a PointerEvent is created, it doesn't have a target at all. target is set during event dispatch. https://dom.spec.whatwg.org/#concept-event-listener-invoke

index.html
```diff
@@ -1104,40 +1104,24 @@ <h2><dfn>Predicted events</dfn></h2>
...
             </ol>
 
-            <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events targets dirty</dfn> and an associated
-            <dfn>predicted events targets dirty</dfn> flag. When an event is created they must be initialized to false.</p>
-
-            <p>When the <a>PointerEvent</a>'s target is changed, set <a><code>coalesced events targets dirty</code></a>
-            and <a><code>predicted events targets dirty</code></a> to true.</p>
-
-            <p>When the <code><a data-lt="PointerEvent.getCoalescedEvents">getCoalescedEvents</a></code> method is called:</p>
+            <p>When a trusted <a>PointerEvent</a>'s target is changed:</p>
```
smaug----, line 1119: ...in fact, isn't this enough. target changes when event is dispatched.

patrickhlauke:
assume this can then be grafted back in/onto https://github.com/w3c/pointerevents/pull/377

going to trust @smaug---- on the whether or not this sufficiently addresses our concern there

index.html
flackr: Right, this is much simpler. Updated.

index.html
```diff
@@ -1116,28 +1116,12 @@ <h2><dfn>Populating and maintaining the coalesced and predicted event lists</dfn
...
             <ol>
-                <li>If the <a>coalesced events targets dirty</a> is true:
-                for each event in the <a>coalesced event list</a>, set the event's {{Event/target}}
+                <li>for each event in the <a>coalesced event list</a>, set the event's {{Event/target}}
                 to this <code>PointerEvent</code>'s target.</li>
```
mustaqahmed, line 1122: Nit: <code>target</code> instead of plain "target"?

mustaqahmed, line 1122: This was there before but never came to me: the "this" here seems unclear to me.  The first point in the above para (source line 1110) doesn't use "this", but also unclear to me :( .
- Can we make both paras consistent?
- Can we perhaps use something like "container event"?

patrickhlauke: keep in mind that this PR will then be followed by https://github.com/w3c/pointerevents/pull/377 which completely rejigs this stuff....

smaug----:
I couldn't find a test for this

smaug----:
I'm writing a test for this.

smaug----:
A super simple manual wpt is here https://github.com/web-platform-tests/wpt/pull/46485
It needs to be manual, since testing predicted events isn't really possibly through non-manual WPTs

PR #395, opened 2021-07-08
Tweaks to coalesced/predicted events section

### LINKED ISSUE #215: Difficulty understanding getCoalescedEvents definition

patrickhlauke:
* clearly list out properly what properties the coalesced/predicted events should have (more closely matching the suggested structure from https://github.com/w3c/pointerevents/issues/215)
* further minor tweaks to the coalesced/predicted/populating... sections to hopefully split out more cleanly what is intended for authors, and what is intended for user agent developers
* add `()` when talking about `getCoalescedEvents()` and `getPredictedEvents()` methods
* consistency: use em-dashes `—` instead of minus or en-dash `–`

Supersedes https://github.com/w3c/pointerevents/pull/377

Closes #215

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/395.html" title="Last updated on Jul 21, 2021, 3:50 PM UTC (3e9b49c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/395/25ec0af...3e9b49c.html" title="Last updated on Jul 21, 2021, 3:50 PM UTC (3e9b49c)">Diff</a>


index.html
```diff
@@ -1016,7 +1021,7 @@ <h2><dfn>Coalesced events</dfn></h2>
                 original events' order. For example if a <code>pointerdown</code> event causes the dispatch for the
                 coalesced <code>pointermove</code> events the user agent SHOULD first dispatch one <code>pointermove</code>
                 event with all those coalesced events of a <code>pointerId</code> followed by the <code>pointerdown</code> event.
-                Here is an example of the actual events happening with increasing <code>timestamps</code> and the
+                Here is an example of the actual events happening with increasing {{Event/timeStamp}} and the
```
mustaqahmed: I also find "`timestamp`s" a bit awkward to read.  But we need a plural here I think, right?  In that case, does "`timestamp` values" read better?

Whatever you chose, please consider making the same change everywhere in this CL.

index.html
```diff
@@ -959,19 +959,24 @@ <h2><dfn>Coalesced events</dfn></h2>
...
+            <code>pointerrawupdate</code> events, the list is a sequence of all <code>PointerEvent</code>s
             that were coalesced into this event; otherwise it is an empty list.
             Untrusted events have their <a>coalesced events list</a> initialized to the value passed to the
             constructor.</p>
 
-            <p>The events in the coalesced events list of a trusted event will have increasing
-            {{Event/timeStamp}}s, so the first event will have the smallest {{Event/timeStamp}}.</p>
+            <p>The events in the coalesced events list of a trusted event will have:</p>
+            <ul>
+                <li>Increasing {{Event/timeStamp}}s, so the first event will have the smallest {{Event/timeStamp}}.</li>
```
mustaqahmed: Awesome, the bullets make the description so clear!

- In particular, I can now see that the original description here was incomplete: the list will have monotonically increasing timestamps, and the parent event's timestamp will be be greater than all of them.

- Related note: perhaps we can make the first bullet in Sec10.1 and the first in Sec10.2 consistent with each other?

index.html
patrickhlauke: ah, good spot. yes, adding `values` should help here. will make that change

index.html
patrickhlauke: yeah, big fan of bullet lists myself where appropriate. i'll try to harmonise the wording/explanation in the first bullet of the two lists.

patrickhlauke: done

index.html
```diff
@@ -959,19 +959,27 @@ <h2><dfn>Coalesced events</dfn></h2>
...
+                <code>getCoalescedEvents()</code> (the red circles) results in a smoother approximation
                 of the pointer movement.</figcaption>
             </figure>
 
             <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events list</dfn> (a list of
-            zero or more <code>PointerEvent</code>s). If a trusted event is a <code>pointermove</code> or
-            <code>pointerrawupdate</code> event, the list is a sequence of all <code>PointerEvent</code>s
+            zero or more <code>PointerEvent</code>s). For trusted <code>pointermove</code> and
+            <code>pointerrawupdate</code> events, the list is a sequence of all <code>PointerEvent</code>s
             that were coalesced into this event; otherwise it is an empty list.
```
mustaqahmed: Nit (also in Sec10.2): without the old "if", the meaning of "otherwise" becomes less obvious.  I guess better to exclude untrusted events clearly, like:
> For all other trusted event types, it is an empty list.

index.html
```diff
@@ -959,19 +959,27 @@ <h2><dfn>Coalesced events</dfn></h2>
...
             Untrusted events have their <a>coalesced events list</a> initialized to the value passed to the
             constructor.</p>
 
-            <p>The events in the coalesced events list of a trusted event will have increasing
-            {{Event/timeStamp}}s, so the first event will have the smallest {{Event/timeStamp}}.</p>
+            <p>The events in the coalesced events list of a trusted event will have:</p>
+            <ul>
+                <li>Monotonically increasing {{Event/timeStamp}} values [[DOM]] - all coalesced events have a
```
mustaqahmed: Nit (also in Sec10.2): emdash instead of dash?

mustaqahmed (approved):
Thanks, this looks great now.

I missed a few nits last time, sorry.  They are optional.

index.html
patrickhlauke: done (and checked throughout document that we use it consistently ... had a few minus signs, and one or two en-dashes)

index.html
patrickhlauke: Ah yes, clearer. done

patrickhlauke:
@flackr @smaug---- would you mind having a look over this too? keen to merge this if possible so it doesn't hang around/need to be superseded again

index.html
```diff
@@ -948,30 +948,38 @@ <h2><dfn>Coalesced events</dfn></h2>
...
+            that were coalesced into this event. For all other trusted event types, it is an empty list.
             Untrusted events have their <a>coalesced events list</a> initialized to the value passed to the
             constructor.</p>
 
-            <p>The events in the coalesced events list of a trusted event will have increasing
-            {{Event/timeStamp}}s, so the first event will have the smallest {{Event/timeStamp}}.</p>
+            <p>The events in the coalesced events list of a trusted event will have:</p>
+            <ul>
+                <li>Monotonically increasing {{Event/timeStamp}} values [[DOM]] — all coalesced events have a
+                    {{Event/timeStamp}} that is smaller than the {{Event/timeStamp}} of the dispatched pointer event that the
```
flackr: I think this should be not greater than. It may be possible that there could be coalesced events with the same timestamp.

index.html
```diff
@@ -1072,10 +1080,17 @@ <h2><dfn>Predicted events</dfn></h2>
...
             the user agent and the prediction algorithm it uses.</p>
 
-            <p>The events in the predicted events list of a trusted event will have monotonically increasing {{Event/timeStamp}}s [[DOM]],
-            so the first event will have the smallest <code>timeStamp</code>. All predicted events have a <code>timeStamp</code>
-            that is greater than the <code>timeStamp</code> of the dispatched pointer event that the
-            <code><a data-lt="PointerEvent.getPredictedEvents">getPredictedEvents</a></code> method was called on.</p>
+            <p>The events in the predicted events list of a trusted event will have:</p>
+            <ul>
+                <li>Monotonically increasing {{Event/timeStamp}} values [[DOM]] — all predicted events have a
+                    {{Event/timeStamp}} that is greater than the {{Event/timeStamp}} of the dispatched pointer event that the
```
flackr: Less likely to occur than with coalesced events, but I think the constraint is technically not less than?

index.html
```diff
@@ -1109,24 +1124,21 @@ <h2><dfn>Predicted events</dfn></h2>
 
         <section>
             <h2><dfn>Populating and maintaining the coalesced and predicted events lists</dfn></h2>
-            <p>When a <a>PointerEvent</a> is created, run the following steps for each event in the
+            <p>When a <a>PointerEvent</a> is created, user agents SHOULD run the following steps for each event in the
```
flackr: I wonder if this should only be for trusted pointer events, or whether we should be doing this processing on user constructed PointerEvents' coalesced and predicted points lists.

index.html
patrickhlauke: so `smaller than, or equal to, the ...` ? or is that too handwavy sounding? (basically `<=`)

index.html
patrickhlauke: as above... `>=` ... `greater than, or equal to, the...` ?

index.html
patrickhlauke: my reading here at the moment is that it's currently requiring this both for trusted and untrusted events, right? are you proposing adding an explicit note about author-generated events that provide their own coalesced/predicted events lists, specifying that user agents SHOULD also go through the lists to ensure this is the case and sanitise whatever the author provided?

we had some previous conundrum with this when it came to the tiltX/tiltY/azimuthAngle/altitudeAngle, and there we decided in the end to require this too for untrusted events - see the note in 4.1

> When an untrusted (synthetic) Pointer Event is created programmatically using the constructor, and only one set of attributes is provided, the complementary set of attributes SHOULD be calculated and initialized by the user agent. If both sets of attributes are provided, no calculation should be performed. If only the value of one of the two attributes is provided, the other attribute SHOULD be initialized to the default value.

patrickhlauke: further, should user agents also sanitize the timestamps that an author has provided, to make sure they match the requirement about monotonically increasing / greater or smaller than the parent event's timestamp?

index.html
flackr: I think smaller than or equal to is also fine.

index.html
flackr: I would have thought we wouldn't attempt to sanitize untrusted events, given that it leads to things like needing to standardize things like the accessor and setter order since all of that is observable when you have constructed your own event.

patrickhlauke: Agreed in today's PEWG meeting (https://www.w3.org/2021/07/21-pointerevents-minutes.html) to add `trusted` here to make sure sanitisation only happens for trusted events. synthetic/untrusted events, we just take whatever the developer provided at face value and don't expect UA to attempt to rewrite/sanitise those lists.

patrickhlauke:
Agreed at today's PEWG meeting (https://www.w3.org/2021/07/21-pointerevents-minutes.html) to merge (following small tweaks done just now).

PR #397, opened 2021-07-21
Remove note about movementX/movementY and coalesced events

### LINKED ISSUE #374: Unclear note about PointerEvent initialization of attributes to reflect coalesced events
As part of the section relating mainly to coalesced events, the spec contains the following note:

> NOTE
> The PointerEvent's attributes will be initalized in a way that best represents all events in the coalesced event list. For example its `movementX` and `movementY` ([POINTERLOCK]) COULD be the sum of those of all the coalesced events.

I'm actually unclear on what this note is trying to say. Right from the start, "The PointerEvent's attributes" ... which pointer event is it referring to? The coalesced "parent" pointer event that has the list of non-coalesced pointer events? If so, is this trying to define (in a non-normative note, but still) what a user agent could do? The spec already says that it isn't trying to define this. Wonder if the whole note should not simply be removed?

If it's to be kept, it needs some disambiguation perhaps

patrickhlauke:
As our spec clearly says it's not trying to define HOW coalescing is done by UAs, this note (particularly as it relates to some other properties not of this spec) feels odd. If anything, this is the sort of detail that could/should be defined in the PointerLock spec?

Closes https://github.com/w3c/pointerevents/issues/374

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/397.html" title="Last updated on Jul 21, 2021, 3:43 PM UTC (1fc5211)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/397/a7ce1ff...1fc5211.html" title="Last updated on Jul 21, 2021, 3:43 PM UTC (1fc5211)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1008,8 +1008,8 @@ partial interface Navigator {
             </pre>
 
             <div class="note">The PointerEvent's attributes will be initalized in a way that best represents
-            the events in the coalesced events list. For example its {{MouseEvent/movementX}} and {{MouseEvent/movementY}}
-            ([[PointerLock]]) COULD be the sum of those of all the coalesced events.</div>
+            the events in the coalesced events list. The specific method by which user agents should do this is not covered by
+            this specification.</div>
 
             <div class="note">
                 <p>The order of all these dispatched events should resemble the actual order of the


mustaqahmed:
Ideally I would expect the removed line to point to a normative text in the PointerLock spec.  Should we try adding that first?

mustaqahmed (approved):
Based on today's discussion, I will file a PointerLock spec issue about adding the "sum of deltas" idea there.

patrickhlauke:
Agreed on today's PEWG call (https://www.w3.org/2021/07/21-pointerevents-minutes.html) to merge this. Ideally, PointerLock and/or UIEvents can tackle this topic, and reference back to PE spec when mentioning coalesced events.

PR #402, opened 2021-08-03
Change instances of "a user agent" to "the user agent"

### LINKED ISSUE #399: "A user agent" vs "the user agent"
The spec inconsistently uses "a user agent" vs "the user agent".  I spotted quite a few places like this:

> A user agent MUST fire a pointer event named ...

where the spec defines a normative behavior for _the_ specific user agent for the whole context.

One precise problem: we can only talk about a "shared" `pointerId` in a `pointerdown`-`pointermove`-`pointerup` event sequence when all these events are in _a single_ user agent implied by the context.  So I think we should say "The user agent MUST fire..." in (almost?) all the cases.

patrickhlauke:
To avoid confusion, any time a normative/clear statement is made, change to "the" (to not give the impression that a UA can somehow delegate things to another UA).

Closes https://github.com/w3c/pointerevents/issues/399

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/402.html" title="Last updated on Aug 3, 2021, 5:43 PM UTC (c138600)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/402/e445470...c138600.html" title="Last updated on Aug 3, 2021, 5:43 PM UTC (c138600)">Diff</a>


mustaqahmed (commented):
What about the "a"s in Section 4.2 and in the `pointerId` para?  In fact, the starting phrase in each subsection 4.2.[1-12] caught my attention originally because I felt like the whole discussion is tied implicitly to a specific UA ("the UA") except when we are comparing two UAs or ruling out exceptional behaviors etc.

patrickhlauke:
d'oh, case-sensitive find/replace fail...

mustaqahmed (approved):
Thanks for the careful fix.

Issue #360, opened 2021-04-14
Clarify that pointer event listeners are passive by default


patrickhlauke:
Split from https://github.com/w3c/pointerevents/issues/359#issuecomment-819828218

patrickhlauke:
@mustaqahmed @smaug---- is there some IDL-specific way of denoting this in the spec? or is this just something that can be done with prose somewhere?

smaug----:
It is not a WebIDL level thing. 
But I'm not sure where it has been said that the listeners would be passive.

(certain touch events and wheel event on the document and body and so are passive.)

patrickhlauke:
yeah wondering if this is more of a heuristic/implementation suggestion (i.e. an informative note), or if it's always true due to the fact that pointer events are non-blocking by their nature (unless there's some `touch-action` shenanigans I guess?)

mustaqahmed:
Idea of passive event listeners is defined [here](https://dom.spec.whatwg.org/#dictdef-eventlisteneroptions) in the DOM Standard.

I think the best statement in that [other discussion you linked](https://github.com/w3c/pointerevents/issues/359#issuecomment-819828218) would have been "pointerevent listeners are _always_ passive".  This is because (to guarantee browser responsiveness) a script can never `preventDefault` the browser behavior in response to a PointerEvent; the script has to declare a `touch-action` region instead.

Perhaps the `touch-action` intro section deserve a link to the spec?  Not sure though.

NavidZ:
Just to add to this regarding passiveness. I also commented something about this [there](https://github.com/w3c/pointerevents/issues/359#issuecomment-823274011).

I guess to be more accurate they are not passive (in the way that DOM spec [says](https://dom.spec.whatwg.org/#dom-addeventlisteneroptions-passive)) as you can prevent default a pointerdown and it will cause the corresponding mouse events not to fire. But the default action of pointerevents have nothing to do with scrolling as already [specified](https://www.w3.org/TR/pointerevents2/#attributes-and-default-actions) in the PointerEvent spec. Although not explicitly called out that scrolling is not part of the default action.
Just a note that if you do add a `passive: true` to pointerevent listener the `preventDefault` on that doesn't work and compat mouse events [always fire](https://jsbin.com/meyocuqugu/edit?html,console,output).

patrickhlauke:
> Just a note that if you do add a `passive: true` to pointerevent listener the `preventDefault` on that doesn't work and compat mouse events always fire.

[...]

PR #403, opened 2021-08-03
Expand prose about preventing compatibility mouse events and make it normative

### LINKED ISSUE #369: Add note explicitly about mouse compat events, preventDefault, and passive events
it's implicit at the moment, but would be good to just make explicit

https://github.com/w3c/pointerevents/issues/360#issuecomment-824407307

patrickhlauke:
* split the awkward "Also," into a separate paragraph (still inside the note)
* add an additional paragraph in the note about passive event listeners (which ignore `preventDefault` / where the default action can't be prevented) - see https://github.com/w3c/pointerevents/issues/360#issuecomment-824407307
* make the note itself normative / remove the "Note" around it

Closes https://github.com/w3c/pointerevents/issues/369

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/403.html" title="Last updated on Aug 17, 2021, 11:22 PM UTC (7866b69)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/403/cbd05e0...7866b69.html" title="Last updated on Aug 17, 2021, 11:22 PM UTC (7866b69)">Diff</a>


smaug---- (changes requested):
Calling preventDefault() from a passive event listener is defined already in the DOM spec so I think it is just a bit confusing to explain it here too 
https://dom.spec.whatwg.org/#set-the-canceled-flag

patrickhlauke:
Still wondering if there's a value in mentioning it - it caught me by surprise when it was mentioned, but then again I'm not hip with the DOM spec. Maybe rewording it slightly but keeping it as a nod/reminder? To make clear that compat events (when generated) are part of the default action, and that of course that default action won't be preventable in those situations? But yeah, not precious about this...if it's clear enough/obvious for others, happy to remove that para (but keep the slight tweak in structure for the note)

smaug----:
Perhaps some note could work here?

mustaqahmed (approved):

patrickhlauke:
As discussed in today's call https://www.w3.org/2021/08/04-pointerevents-minutes.html added an explicit `[[DOM]]` reference there to make it clearer that the concepts are defined in DOM

patrickhlauke:
However, with our newfound interest in normative/non-normative ... wondering if that second note for https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/403.html#compatibility-mapping-with-mouse-events should actually just be normal normative prose rather than a non-normative note. Thoughts?

flackr (approved):

mustaqahmed:
We have [pointerevent_mouse-pointer-preventdefault.html](https://wpt.fyi/results/pointerevents/compat/pointerevent_mouse-pointer-preventdefault.html?label=master&label=experimental&aligned&view=subtest) to cover almost all of this PR.  However, we still need a WPT for compat events for passive PointerEvent listeners.

smaug----:
https://github.com/web-platform-tests/wpt/pull/38525

mustaqahmed:
While working on a test on passive event listeners, I realized that this behavior is not normative!  It seems to me that the "note" we added here could be normative...thoughts?

mustaqahmed:
> While working on a test on passive event listeners, I realized that this behavior is not normative! It seems to me that the "note" we added here could be normative...thoughts?

In today's PEWG meeting we agreed to make this normative.

patrickhlauke:
> In today's PEWG meeting we agreed to make this normative.

done https://github.com/w3c/pointerevents/pull/465

mustaqahmed:
The WPT landed through https://github.com/web-platform-tests/wpt/pull/38895

PR #404, opened 2021-08-04
Add note about rounding coordinates for click, auxclick, contextmenu

### LINKED ISSUE #100: Should "click", "dblclick" and "contextmenu" events be PointerEvents?

patrickhlauke:
See the comment about web compat issues with the "upgraded" click/auxclick/contextmenu as pointer events https://github.com/w3c/pointerevents/issues/100#issuecomment-818870316

Closes https://github.com/w3c/pointerevents/issues/100

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/404.html" title="Last updated on Aug 25, 2021, 7:52 PM UTC (34864ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/404/6d33e08...34864ef.html" title="Last updated on Aug 25, 2021, 7:52 PM UTC (34864ef)">Diff</a>


patrickhlauke:
Currently set this as a note, as I'm not sure if we want to set this in stone as a requirement (though I guess then that invalidates the `SHOULD` I used ... not sure). Let's discuss at today's meeting.

patrickhlauke:
Related: should we also mention anything related to this explicitly in the mouse compatibility section? So user agents know that they should also round/convert to `long` when generating compat mouse events?

patrickhlauke:
Last tangential point: currently, the definition for `{{PointerEvent}}` links to https://w3c.github.io/pointerevents/#dom-pointerevent - should it not go to what is essentially the interface https://w3c.github.io/pointerevents/#pointerevent-interface ? (if so, just need to repoint the `<dfn>` appropriately). To me this would make more sense, but not sure what the conventions are for specs...

patrickhlauke:
And last additional point, I promise: wondering if the current note in 4.1

> The PointerEvent interface inherits from MouseEvent, defined in UI Events and extended by CSSOM View Module.

should actually be expanded to make clear what that last bit means ... fractional coordinates. took me a while to work out that nowhere in the PE spec we actually explicitly say anything about fractional coordinates/that they're `double`

patrickhlauke:
Made the changes discussed in today's meeting:

* expanded the note at the end of 4.1 (to foreshadow the topic a bit more about what CSSOM proposes, the fact that it's still only a proposal not a standard, and that there are UA requirements for those UAs that do already follow this)
* made the proposed note for 4.2.12 normative instead

While I was in 4.2.12, also cleaned up the structure of the prose there to make the distinction between events that came from pointing devices and those that didn't more obvious when scanning over the section.

smaug---- (approved):

patrickhlauke:
Just adding here, now that I thought it through some more: so the compat problems/issues would come about in user agents where PointerEvents have been implemented already using fractional coordinates, but the same implementation hasn't been extended to regular MouseEvents, right? If a browser decided to already implement fractional coordinates for MouseEvents, this would actually lead to problems again (essentially, the opposite type of problem this tries to address), right?

In short, the problem only appeared because Chrome (and others?) didn't apply the CSSOM approach for MouseEvents, but *just* for PointerEvents, right? So my proposed sentence

> For user agents that already implement this proposed extension, there are additional normative requirements when it comes to the <a><code>click</code>, <code>auxclick</code>, and <code>contextmenu</code> events</a>

should probably more specifically be about

> For user agents that already implement this proposed extension for PointerEvents, but <em>not</em> for MouseEvents, there are additional normative requirements when it comes to the <a><code>click</code>, <code>auxclick</code>, and <code>contextmenu</code> events</a>

likewise, in

> However, this change — when applied to <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code> — has proven to lead to web compatibility issues with legacy code.

it should probably read more along the lines of

> However, this change — when applied to PointerEvents, but not to regular MouseEvents — has proven to lead to web compatibility issues with legacy code when applied to <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>.

i.e. if the user agent hasn't implemented fractional coordinates anywhere (not in MouseEvents, not in PointerEvents), it's a non-issue. and it's also a non-issue if the user agent has applied this change to both MouseEvents and PointerEvents. it's only a problem when it's only implemented in one and not the other, right?

patrickhlauke:
Just pushed a change to this PR to clarify this further. I *think* this now more accurately explains where the problem lies, and when this requirement needs to be implemented.

mustaqahmed:
> Just pushed a change to this PR to clarify this further. I _think_ this now more accurately explains where the problem lies, and when this requirement needs to be implemented.

I am afraid it was more accurate before this last commit!  Let me add a few points to focus on, in case they call for more discussion:
- `Double` coordinates (as per CSSOM View spec) are known to be better but we can't move MouseEvent coordinates away from `long` for compat reasons.  We logically _expect_ that all UAs would pick `double` coordinates for PointerEvents (as per CSSOM) even if those UAs remain stuck with long coordinates in MouseEvents (as per UI Event spec).
- The _expectation_ above applied only to "original" PointerEvents (`pointerdown`, `pointermove` etc).  The "newly promoted" PointerEvents (`click`, `auxclick` and `contextmenu`) suffer from the similar compat issues like MouseEvents.
- The two points above are orthogonal to compat MouseEvents.
- The compat event section only defines the firing of followup MouseEvents, without touching any other details of the events---because those details belong to the UIEvent spec.  Adding more details here means more redundancy and may possibly call for more monkey-patching.

[EDIT: I mentioned these points only for discussion.  Most likely they don't deserve to be included in the PR here.]

patrickhlauke:
> We logically expect that all UAs would pick double coordinates for PointerEvents (as per CSSOM) even if those UAs remain stuck with long coordinates in MouseEvents (as per UI Event spec)

we don't normatively, explicitly, actually say this in our spec at the moment, right? A UA could implement PEs but with `long` rather than `double`? (as normatively the WebIDL only says we extend `MouseEvent`, and the CSSOM change is mentioned in an informative note only.

> we can't move MouseEvent coordinates away from long for compat reasons

but IF a user agent decided to apply the CSSOM fractional coordinates to their regular MouseEvent as well, then rounding click/auxclick/contextmenu to `long` would then cause the same weird mismatch, no? or is the expectation that the CSSOM change will never actually make it to regular MouseEvent ?

> The two points above are orthogonal to compat MouseEvents.

yes and my PR doesn't mention compat events. when I talk about MouseEvent, I mean the actual ones.

mustaqahmed:
> we don't normatively, explicitly, actually say this in our spec at the moment, right? A UA could implement PEs but with `long` rather than `double`? (as normatively the WebIDL only says we extend `MouseEvent`, and the CSSOM change is mentioned in an informative note only.
> 
> > we can't move MouseEvent coordinates away from long for compat reasons
> 
> but IF a user agent decided to apply the CSSOM fractional coordinates to their regular MouseEvent as well, then rounding click/auxclick/contextmenu to `long` would then cause the same weird mismatch, no? or is the expectation that the CSSOM change will never actually make it to regular MouseEvent ?

I ruled out this consideration because this is known to be impractical, therefore presumed all UAs ignore CSSOM extension for MouseEvents.  After seeing (again!) that this can't be normative in the PE spec, I am finally able to see the motivation of your last patch.  Phew!

> > The two points above are orthogonal to compat MouseEvents.
> 
> yes and my PR doesn't mention compat events. when I talk about MouseEvent, I mean the actual ones.

Looks like I misinterpreted "regular MouseEvent" as "compat MouseEvent", for no good reason!  I am taking back my compat event points.

mustaqahmed (approved):

patrickhlauke:
it's a tangled web that we weave...this CSSOM thing stretched my brain even more than that target discussion a few months ago...

index.html
```diff
@@ -634,11 +638,20 @@ <h3>The <dfn data-lt="lostpointercapture"><code>lostpointercapture</code> event<
...
+                    </ul>
+                <p>As noted in {{PointerEvent}}, [[[CSSOM-VIEW]]] proposes to redefine the various coordinate properties (<code>screenX</code>, <code>screenY</code>, <code>pageX</code>, <code>pageY</code>, <code>clientX</code>,
+                   <code>clientY</code>, <code>x</code>, <code>y</code>, <code>offsetX</code>, <code>offsetY</code>) as <code>double</code>, to allow for fractional coordinates.
+                   However, this change — when applied only to {{PointerEvent}}, but not to regular {{MouseEvent}} — has proven to lead to web compatibility issues with legacy code
+                   in the case of <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>. For this reason, user agents that have only implemented the proposed
+                   change in [[[CSSOM-VIEW]]] for {{PointerEvent}} MUST round the various coordinate properties for the <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>
```
flackr: *Rounding* isn't necessarily going to make the pointerevent equivalent to the mouse event. E.g. In chromium, we [floor the coordinates](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/events/mouse_event.h;l=140). Is there some wording we can use here to say they must be equivalent to the compatibility mouse events?

patrickhlauke: for compat mouse events, we don't say anything regarding coordinates at all at the moment. and generating compat mouse events is an optional feature that browser may or may not choose to implement. this may lead to a further conditional sentence here "if the user agent implements compatibility mouse events, do the same as there. otherwise..." and then that second case would still need to be explained. hmm

at the end of 4.1 we have a note with

> When the user agent calculates tiltX / tiltY from azimuthAngle / altitudeAngle it SHOULD round the final integer values using Math.round [ECMASCRIPT] rules.

maybe here we can explicitly say it should be Math.floor ? (and leave out any discussion of compat events)

patrickhlauke: i can already see us having to now also say something in the compat mouse events section about coordinates... got ourselves a bit tangled here...

flackr: It occurred to me after writing this that the mapping from floating point to integer is entirely developer visible ([can observe both the PointerEvent and the MouseEvent](https://jsbin.com/sagenev/edit?html,js,output)), so it probably should be well defined which way numbers are rounded. As far as I can tell from that jsbin chromium is the odd browser out to even use floating point coordinates for PointerEvents.

patrickhlauke: wondering if the WebIDL for PE can/should be changed to normatively define coordinates as `double` (assume it would be easier for PE to do this, as it's a new type of event, rather than what CSSOM tried to do with redefining existing MouseEvent

patrickhlauke: ...or if this is just something that we just leave as is (but which then leads to having these awkward "if you're following CSSOM, then..." type constructs)

patrickhlauke:
per the discussion at last week's meeting https://www.w3.org/2021/08/18-pointerevents-minutes.html changed the "round" to reference `Math.floor` and point to ECMAScript.

index.html
```diff
@@ -634,11 +638,20 @@ <h3>The <dfn data-lt="lostpointercapture"><code>lostpointercapture</code> event<
...
+                       <li>If the events are generated by a non-pointing device (such as voice recognition software or a keyboard interaction), <code>pointerId</code> MUST be <code>-1</code> and <code>pointerType</code> MUST be an empty string.</li>
+                    </ul>
+                <p>As noted in {{PointerEvent}}, [[[CSSOM-VIEW]]] proposes to redefine the various coordinate properties (<code>screenX</code>, <code>screenY</code>, <code>pageX</code>, <code>pageY</code>, <code>clientX</code>,
+                   <code>clientY</code>, <code>x</code>, <code>y</code>, <code>offsetX</code>, <code>offsetY</code>) as <code>double</code>, to allow for fractional coordinates.
+                   However, this change — when applied only to {{PointerEvent}}, but not to regular {{MouseEvent}} — has proven to lead to web compatibility issues with legacy code
+                   in the case of <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>. For this reason, user agents that have only implemented the proposed
```
mustaqahmed: Two optional nits:
- Does moving the word "only" to before "for PointerEvent" sound better?
- Please consider not using Math.floor as a verb.

With both suggestions, we would have this (changes in marked):
> user agents that have implemented the proposed change in CSSOM View Module **only** for PointerEvent MUST ~Math.floor [ECMASCRIPT]~ **convert** the coordinate properties for the click, auxclick, and contextmenu to long values (as defined in the original UI Events) **using Math.floor [ECMASCRIPT]**.

mustaqahmed (commented):
Two super picky comments!

index.html
patrickhlauke: good suggestions. done.

mustaqahmed:
I am looking into a WPT for this!

smaug----:
https://github.com/web-platform-tests/wpt/pull/38525

mustaqahmed:
Here I was modifying `pointerevent_fractional_coordinates.html` to cover integer coordinates in `click`-like events, and I think we still need that.  Did I miss anything?

smaug----:
oops, I posted the link to wrong issue.

mustaqahmed:
No worries, I updated the issue you meant: https://github.com/w3c/pointerevents/pull/456

mustaqahmed:
A WPT has been added.

Issue #405, opened 2021-08-04
Notes with normative (SHOULD/MUST) statements


patrickhlauke:
> all authoring guidelines, diagrams, examples, and notes in this specification are non-normative

however, we have a few instances of notes that contain MAY, SHOULD, SHOULD NOT, MUST.

- [x] https://w3c.github.io/pointerevents/#dom-pointerevent-pointerid > https://github.com/w3c/pointerevents/pull/411
- [x] first note in https://w3c.github.io/pointerevents/#dom-pointerevent-getpredictedevents > https://github.com/w3c/pointerevents/pull/410
- [x] second note in https://w3c.github.io/pointerevents/#coalesced-events > https://github.com/w3c/pointerevents/pull/412
- [x] second note in https://w3c.github.io/pointerevents/#mapping-for-devices-that-do-not-support-hover > https://github.com/w3c/pointerevents/pull/413
- [x] note in https://w3c.github.io/pointerevents/#chorded-button-interactions - make the note normative (about compat events and how the button/buttons should follow UIEVENTS) >  https://github.com/w3c/pointerevents/pull/414
- [x] note in https://w3c.github.io/pointerevents/#firing-events-using-the-pointerevent-interface ... contains "may" and "should", but not sure how strongly this is intended. they're not capitalised, but may give impression of being MAY and SHOULD in normative sense? more fundamentally, are we ok with *not* normatively defining if boundary events are fired or not? https://github.com/w3c/pointerevents/pull/419
- [x] first note in https://w3c.github.io/pointerevents/#the-touch-action-css-property right at the end contains a "should" - i'd reword that to avoid the word altogether > https://github.com/w3c/pointerevents/pull/415
- [x] first note in https://w3c.github.io/pointerevents/#compatibility-mapping-with-mouse-events is full of "may" and "should" - is it clear enough these are not used in the MAY and SHOULD sense? https://github.com/w3c/pointerevents/pull/418
- [x] second note in https://w3c.github.io/pointerevents/#dfn-active-pointer - i'd make the note normative text (and change "id" to "`pointerId`"). however, this is in the glossary, which by definition is non-normative. should this be moved somewhere else / into the normative text/interface definition? also, what does that second note *actually* try to say? https://github.com/w3c/pointerevents/pull/417
[...]

PR #411, opened 2021-09-13
Move elements of `pointerId` informative note to normative text

patrickhlauke:
x-ref https://github.com/w3c/pointerevents/issues/405 https://www.w3.org/2021/09/01-pointerevents-minutes.html#t02

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/411.html" title="Last updated on Oct 13, 2021, 3:28 PM UTC (b9ad01d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/411/76c9195...b9ad01d.html" title="Last updated on Oct 13, 2021, 3:28 PM UTC (b9ad01d)">Diff</a>


patrickhlauke:
felt that the part about the lifetime of the page etc was still quite important, so grafted it into the normative part. we can discuss this week at the meeting

index.html
```diff
@@ -228,10 +228,10 @@ <h2><code>PointerEvent</code> interface</h2>
...
+                            <p>A unique identifier for the pointer causing the event. This identifier MUST be unique from all other <a data-lt="active pointer">active pointers</a> in the <a>top-level browsing context</a> (as defined by [[HTML]]) at the time, and the identifier MUST NOT be influenced by any other top-level browsing context. The <code>pointerId</code> value of <code>-1</code> is reserved to indicate events that were generated by something other than a pointing device. The user agent MAY recycle previously retired values for <code>pointerId</code> from previous active pointers, or it MAY always reuse the same <code>pointerId</code> for a particular pointing device for the lifetime of the page. However, in the latter case, the <code>pointerId</code> MUST be randomized, to minimize the chance of fingerprinting and tracking across different pages.</p>
```
mustaqahmed: Does it look a bit less hand-wavy?

> ...MUST NOT be influenced by _existing pointers in_ any other... (emphasis marks the new text)

Not sure if this would leave something important out!

patrickhlauke:
discussed at today's meeting, needs a bit more finessing:

* need to clarify that mouse pointerId can be unique (and is, though the value differs across implementations)
* bring forward the rationale for doing it (tracking prevention)

patrickhlauke:
@mustaqahmed @smaug---- @flackr made an admittedly wordy stab at turning the whole explanation around to make it clearer, while also introducing the new idea that mouse's `pointerId` can be reserved and always the same (it's what Chrome and Firefox do) but without specifying any particular value. thoughts appreciated ... it still feels super dense, but maybe it IS clearer?

patrickhlauke:
note to self: this PR will need a more descriptive title when/if it gets merged

flackr (approved):

mustaqahmed:
The major part of this PR is covered by WPTs.  The few remaining points seem hard to test: most are optional, others need multiple top-level browsing contexts (needs discussion).  Below is a dump of the `pointerId` [section](https://w3c.github.io/pointerevents/#ref-for-dom-pointerevent-pointerid-1) with my comments:

> User agents MAY reserve a generic pointerId value of 0 or 1 for the primary mouse pointer.

Optional, hence untestable.

> The pointerId value of -1 MUST be reserved and used to indicate events that were generated by something other than a pointing device.

Covered in `*_is_a_pointerevent.html`.

> For any other pointers, user agents are free to implement different strategies and approaches in how they assign a pointerId value. However, all active pointers in the top-level browsing context must be unique, and the identifier MUST NOT be influenced by any other top-level browsing context (i.e. one top-level browsing context cannot assume that the pointerId of a pointer will be the same when the pointer moves outside of the browsing context and into another top-level browsing context).

Is there a way to have multiple top-level browsing contexts in a WPT?

> The user agent MAY recycle previously retired values for pointerId from previous active pointers, or it MAY always reuse the same pointerId for a particular pointing device (for instance, to uniquely identify particular pen/stylus inputs from a specific user in a multi-user collaborative application).

Optional, hence untestable.

> However, in the latter case, to minimize the chance of fingerprinting and tracking across different pages or domains, the pointerId MUST only be associated explicitly with that particular pointing device for the lifetime of the page / session, and a new randomized pointerId MUST be chosen the next time that particular pointing device is used again in a new session.

We can simulate a touch pointer crossing an `iframe` boundary multiple times and can swap `iframe` content during this interaction.  But this is complicated, and the MUST here is conditional!

mustaqahmed:
> Is there a way to have multiple top-level browsing contexts in a WPT?

In a past PEWG meeting, we agreed this may be testable using a same origin popup but it is not clear if "moving a pointer from one browsing context to another" is even supported by test-driver actions.

Removing the needs-wpt label because the major part of the PR is covered by WPTs already.

PR #413, opened 2021-09-13
Make note about UAs not firing compat events if they support touch events normative

patrickhlauke:
x-ref https://github.com/w3c/pointerevents/issues/405

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/413.html" title="Last updated on Sep 30, 2021, 11:13 PM UTC (e7b9ba6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/413/76c9195...e7b9ba6.html" title="Last updated on Sep 30, 2021, 11:13 PM UTC (e7b9ba6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1227,9 +1227,7 @@ partial interface Navigator {
                 </li>
                 <li>If the pointer event dispatched was <code>pointerup</code> or <code>pointercancel</code>, clear the <code>PREVENT MOUSE EVENT</code> flag for this <code>pointerType</code>.</li>
             </ol>
-            <div class="note">
-                <p>If the user agent supports both Touch Events (as defined in [[TOUCH-EVENTS]]) and Pointer Events, the user agent SHOULD NOT generate compatibility mouse events as described in this section as it is likely to introduce compatibility problems for sites that expect mouse events to be generated in accordance with the <a data-cite="touch-events/#mouse-events">model</a> outlined in [[TOUCH-EVENTS]].</p>
-            </div>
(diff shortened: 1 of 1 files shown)


mustaqahmed:
The linked normative text in the TouchEvent spec starts with a "may", so IMHO we need to avoid normative suppression of compat events here!  I suggest keeping the note as is but with a "MAY" instead.

patrickhlauke:
we're trying to avoid MAY, SHOULD, etc inside non-normative notes though. i missed that touch event mouse events were MAY. so, here's a thought: change this (i'll have a think how exactly) to basically say that UAs SHOULD NOT generate *both* and leave it at that (and leave it up to UAs to decide what to do ... or is that too wooly?)

mustaqahmed:
Sounds good.  Can we somehow highlight the possibility of duplicate event firing?  I think it is possible to cover both specs using a single set of compact events (even with multi-touch because we care only about the primary touch pointer here), but the "may" on the other side makes it impossible to be precise.

> If the user agent supports both Touch Events and Pointer Events, the user agent SHOULD guarantee not to fire duplicate compatibility events from a single touch interaction...

patrickhlauke:
Updated the wording per the above (noting that touch events aren't just for touch interactions, but also cover pen/stylus nowadays)

mustaqahmed (approved):
Awesome, this looks great!

patrickhlauke:
discussed in meeting today: change SHOULD NOT to MUST NOT, take out last sentence ", as a result of a touch or pen interaction." as it's redundant

patrickhlauke:
Made the edits discussed in meeting yesterday. Merging, as otherwise we were ok with this.

mustaqahmed:
The WPTS in [pointerevents/compat/](https://wpt.fyi/results/pointerevents/compat?label=master&label=experimental&aligned&view=subtest) folder asserts the correct number of compat mouse events, whether or not the UA supports Touch Events.

PR #419, opened 2021-10-27
Remove should from boundary events note and move to normative must.

flackr:
x-ref #405

Clarify in normative text that the target should be treated as if the pointing device moved over it, but defer whether boundary events should be fired to uievents.

***
<a href="https://pr-preview.s3.amazonaws.com/flackr/pointerevents/pull/419.html" title="Last updated on Dec 6, 2021, 10:02 AM UTC (6327b0f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/419/ef3ea93...flackr:6327b0f.html" title="Last updated on Dec 6, 2021, 10:02 AM UTC (6327b0f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -405,9 +405,9 @@ interface PointerEvent : MouseEvent {
                 <p>If the event is <code>pointerdown</code>, the associated device is a direct manipulation device, and the target is an {{Element}},
                    then <a href="#setting-pointer-capture">set pointer capture</a> for this <code>pointerId</code> to the target element as described in <a>implicit pointer capture</a>.
 
-                <p>Fire the event to the determined target.
+                <p>Fire the event to the determined target. The user agent SHOULD treat the target as if the pointing device has moved over it for the purpose of <a href="uievents/#events-mouseevent-event-order">ensuring event ordering</a>.
 
-                <div class="note">Using the <a>pointer capture target override</a> as the target instead of the normal hit-test result may fire some boundary events. This is the same as the pointer leaving its previous target and entering this new capturing target — and if they are different targets, boundary events should be dispatched first. When the capture is released, the same scenario may happen, as the pointer is leaving the capturing target and entering the hit-test target.</div>
+                <div class="note">Using the <a>pointer capture target override</a> as the target instead of the normal hit-test result may fire some boundary events. This is the same as the pointer leaving its previous target and entering this new capturing target. When the capture is released, the same scenario may happen, as the pointer is leaving the capturing target and entering the hit-test target.</div>
 
                 <section>
                   <h3>Attributes and default actions</h3>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -405,9 +405,9 @@ <h2>Firing events using the <code>PointerEvent</code> interface</h2>
...
+                <p>Fire the event to the determined target. The user agent SHOULD treat the target as if the pointing device has moved over it for the purpose of <a href="uievents/#events-mouseevent-event-order">ensuring event ordering</a>.
 
-                <div class="note">Using the <a>pointer capture target override</a> as the target instead of the normal hit-test result may fire some boundary events. This is the same as the pointer leaving its previous target and entering this new capturing target — and if they are different targets, boundary events should be dispatched first. When the capture is released, the same scenario may happen, as the pointer is leaving the capturing target and entering the hit-test target.</div>
+                <div class="note">Using the <a>pointer capture target override</a> as the target instead of the normal hit-test result may fire some boundary events. This is the same as the pointer leaving its previous target and entering this new capturing target. When the capture is released, the same scenario may happen, as the pointer is leaving the capturing target and entering the hit-test target.</div>
```
patrickhlauke: are we happy to leave this vague? is this aspect covered by UIEvents as well, and if so should we say as much (that we're not being more precise here, because that's down to UIEvents)?

patrickhlauke: or do we want to abdicate responsibility here and say "this aspect is not specified and left up to user agents" or some such?

flackr: The requirement to dispatch boundary events is up to the ui-events spec, and the note was meant to be informative that as a result of the capture being equivalent to moving the mouse over the capture element it may (as defined by ui-events) generate boundary events. Would it be better if I changed it to read "may fire some boundary events as defined by <a href="uievents/#events-mouseevent-event-order">ui-events</a>"?

patrickhlauke: yeah i think adding in a reference to [UI-EVENTS] would just make it clear that this *is* actually defined (somewhere else) and that we're not just handwaving it

patrickhlauke: (which then also excuses the use of "may" in that note, as we're clearly not making any kind of normative statement since we're just pointing somewhere else)

patrickhlauke:
any more updates for this @flackr / @mustaqahmed ? ref. last meeting's notes https://www.w3.org/2021/11/10-pointerevents-minutes.html#t01

index.html
patrickhlauke: added the reference to `as defined by [[UI-EVENTS]]` to this PR

patrickhlauke:
Going to merge this as is now, as I *think* it does what we set out to do. @mustaqahmed if you still have concerns about this not being specific enough, we can do a separate PR afterwards

mustaqahmed:
This change is covered by the WPTs [pointerevent_boundary_events_in_capturing.html](https://wpt.fyi/results/pointerevents?label=master&label=experimental&aligned&view=subtest&q=pointerevent_boundary_events_in_capturing.html).

PR #423, opened 2021-12-06
Improve illustrations

patrickhlauke:
- added correct axes to the `tiltX`/`tiltY` diagrams, to match the fact that the Y axis increases when going towards the user. this removes the weird confusion of the spec talking about positive `tiltY` values when it seemed that the Y axis was actually pointing in the other direction
- matched the extended axis lines for the `altitudeAngle`/`azimuthAngle` diagrams
- redrawn coalesced and predicted points diagrams - while the originals came straight from my little visualisation/test page, they were rather low-resolution and poor quality. these are now redrawn in vector format for a crisper look, and the coalesced one is rotated to have a landscape, rather than portrait, aspect ratio
- added all source files used to create the newer diagrams


mustaqahmed:
This is great: consistently putting the axis labels on the increasing directions makes the `tiltX/Y` diagrams much clearer!  This provides an additional visual reference for the "up direction" in the x-y plane--previously only the tablet camera served this purpose.

@patrickhlauke Can we possibly add similar axis labels also in the diagrams for `altitudeAngle` and `azimuthAngle`?  It will be super helpful when the reader is comparing two "angle systems" we have here.

patrickhlauke:
@mustaqahmed ah, good point. I was worried the axis labels might detract a bit from the "meat" of those diagrams, but yeah it couldn't hurt adding them (maybe in a slightly more subtle color/size). I'll have a play and make a new PR when ready

PR #429, opened 2022-01-19
Split touch-action bullet into two separate bullets, clarify ancestors

### LINKED ISSUE #319: touch-action:none and overflow:auto
In the following code:
```
<div id="container" style="touch-action:none; overflow: scroll">
  <div id="intermediate" style="overflow: auto">
    <div id="target"></div>
  </div>
</div>
```
If scroll action starts at `#target`, for the purposes of determining supported touch behaviour, should we consider `#intermediate` to be a scroller only if it is able to be scrolled (content is big enough) or in all cases?

This is in reference to [1]: _A touch behavior is supported if it conforms to the touch-action property of each element between the hit tested element and its nearest ancestor with the default touch behavior (including both the hit tested element and the element with the default touch behavior)._
 
[1] https://www.w3.org/TR/pointerevents3/#determining-supported-touch-behavior

patrickhlauke:
(also fixes reference to top-level browsing context)

Closes https://github.com/w3c/pointerevents/issues/319

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/429.html" title="Last updated on Jan 19, 2022, 5:59 PM UTC (b9099c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/429/c058225...b9099c9.html" title="Last updated on Jan 19, 2022, 5:59 PM UTC (b9099c9)">Diff</a>


flackr (approved):
Looks good to me!

Issue #409, opened 2021-09-07
Relationship between main pointer event and coalesced events


jakearchibald:
It isn't clear to me how the [coalesced event list](https://w3c.github.io/pointerevents/#dfn-coalesced-events-list) works.

If I have a trusted event from `pointermove`, should `event.getCoalescedEvents()` ever return an empty array? It does on Firefox Android.

Let's say `event.getCoalescedEvents()` returns an array of two items:

```js
const [one, two] = event.getCoalescedEvents();
```

What's the relationship between the single x/y points expressed by `event`, `one` and `two`? Is either `one` or `two` representing the same x/y point as `event`? If not, where does the `event` x/y point appear in the sequence?

jakearchibald:
It seems to me that `event.getCoalescedEvents()` should return a complete set of x/y points for that event, meaning that the last x/y point returned by `event.getCoalescedEvents()` will be the same x/y point expressed by `event` itself.

This means that `event.getCoalescedEvents()` shouldn't return an empty array, as it would always at least contain the x/y point expressed by `event`.

patrickhlauke:
my understanding so far was that the array contains all points *except* the one that was actually dispatched as a "real" pointer event - i.e. coalesced list includes all "dropped" points that ended up being coalesced/quietly ignored leading up to the actual one that was sent.

as for Firefox/Android, I believe the implementation of coalesced and predicted events lists there is only a stub at the moment (but @smaug----  may know more). indeed, testing using https://patrickhlauke.github.io/touch/coalesced-predicted-events/ shows me no entries in the coalesced events lists when trying on Firefox/Android (though Firefox/Windows it works nicely, though still no predicted events there either).

jakearchibald:
> my understanding so far was that the array contains all points _except_ the one that was actually dispatched as a "real" pointer event - i.e. coalesced list includes all "dropped" points that ended up being coalesced/quietly ignored leading up to the actual one that was sent.

So the order is `[one, two, event]`? That isn't clear to me from the spec, and I'm not sure it's developer-friendly either.

For example, if you want to do something with all the points, which seems like the common case, you end up doing something like this:

```js
ctx.beginPath();
ctx.moveTo(previousEvent.clientX, previousEvent.clientY);
for (const event of [...mainEvent.getCoalescedEvents(), mainEvent]) {
  ctx.lineTo(event.clientX, event.clientY);
}
ctx.stroke();
```

Rather than:

```js
ctx.beginPath();
[...]

PR #436, opened 2022-03-15
Clarify that coalesced events list contain also a clone of the parent event

smaug----:
Since the list below talks about "parent" event, perhaps just clarifying the sentence is actually enough here..

This PR is for https://github.com/w3c/pointerevents/issues/409

***
<a href="https://pr-preview.s3.amazonaws.com/smaug----/pointerevents/pull/436.html" title="Last updated on Apr 13, 2022, 11:49 AM UTC (866d077)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/436/614dc2c...smaug----:866d077.html" title="Last updated on Apr 13, 2022, 11:49 AM UTC (866d077)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -966,7 +966,7 @@ partial interface Navigator {
             <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events list</dfn> (a list of
             zero or more <code>PointerEvent</code>s). For trusted <code>pointermove</code> and
             <code>pointerrawupdate</code> events, the list is a sequence of all <code>PointerEvent</code>s
-            that were coalesced into this event, including a clone of the "parent" event itself.
+            that were coalesced into this event, or if no events were coalesced, it contains a clone of the "parent" event itself.
             For all other trusted event types, it is an empty list.
             Untrusted events have their <a>coalesced events list</a> initialized to the value passed to the
             constructor.</p>
--- a/index.html
+++ b/index.html
@@ -966,7 +966,8 @@ partial interface Navigator {
             <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events list</dfn> (a list of
             zero or more <code>PointerEvent</code>s). For trusted <code>pointermove</code> and
             <code>pointerrawupdate</code> events, the list is a sequence of all <code>PointerEvent</code>s
-            that were coalesced into this event. For all other trusted event types, it is an empty list.
+            that were coalesced into this event, including a clone of the "parent" event itself.
+            For all other trusted event types, it is an empty list.
             Untrusted events have their <a>coalesced events list</a> initialized to the value passed to the
             constructor.</p>
 


index.html
```diff
@@ -966,7 +966,8 @@ <h2><dfn>Coalesced events</dfn></h2>
             <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events list</dfn> (a list of
             zero or more <code>PointerEvent</code>s). For trusted <code>pointermove</code> and
             <code>pointerrawupdate</code> events, the list is a sequence of all <code>PointerEvent</code>s
-            that were coalesced into this event. For all other trusted event types, it is an empty list.
+            that were coalesced into this event, including a clone of the "parent" event itself.
```
patrickhlauke: do we need to maybe even specify that the clone is the *last* event in the list? (assuming we do intend it to be the last, and not just "buried somewhere in that list is also the parent")

smaug----: The order is defined in the text below.

patrickhlauke: ah, good point...the danger of looking at the one sentence here in isolation.

flackr: Minor nit: since the parent event can include slight interpolation / prediction it might be better to say something like including the event represented by the parent event?

flackr: Nevermind, it looks like the parent event should have the same coordinates

smaug----:
The current idea is to explicitly define parent event's movementX/Y to be relative to the previous parent event and the
movementX/Y of coalesced events to be relative to the previous event in a coalesced event list.

index.html
NavidZ: Yeah. I don't think the new statement is fully correct. The parent event might have different movementX/Y as @smaug---- also mentioned but it could also be resampled to what the user agent believes to be best representative of the current state (e.g. location, pressure and other similar stuff) of the pointer as @flackr mentioned. On the other hand the coalesced events are meant to be raw events received from the platfrom with no modifications on them.

smaug----:
I think one possible way to clarify this is to just say that if the coalesced events list was otherwise empty for trusted pointermove/pointerrawupdate, then it contains a clone of the parent event.

And then have totally separate definition how movementX/Y is computed for the events in the coalesced events list and for the parent events.

patrickhlauke:
so, naive idea, can the wording just say something along the lines of "the last event in the coalesced events list corresponds to the "parent" event (with the same values for the various PointerEvent properties). This means that for trusted pointer events, coalesced events list will always have at least one event." or something along those lines?

and yes, then adding an informative note that other properties, such as `movementX/movementY`, of that last event are likely to differ from the parent event. or is that too handwavy?

index.html
```diff
@@ -966,7 +966,8 @@ <h2><dfn>Coalesced events</dfn></h2>
             <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events list</dfn> (a list of
             zero or more <code>PointerEvent</code>s). For trusted <code>pointermove</code> and
             <code>pointerrawupdate</code> events, the list is a sequence of all <code>PointerEvent</code>s
-            that were coalesced into this event. For all other trusted event types, it is an empty list.
+            that were coalesced into this event, or if no events were coalesced, it contains a clone of the "parent" event itself.
```
smaug----, line 969: this is explicitly not talking about movement, since that seem to be separate issue altogether. Currently per specifications all the pointerevents should have movementX/Y 0 (whether or not the are in the coalesced events list). But the idea here is that this could work once movement is clarified elsewhere.

patrickhlauke:
naive idea (as usual from me): how about defining that *if the list does not have any events* (i.e. there were no extra events that were coalesced into a parent), then the list will contain a clone of the parent event, and otherwise leaving it as is? would that cover the scenario where a developer wants to *only* use the coalesced events list?

edit: conceptually, I could even rationalise this in terms of the coalesced list containing all events that led the browser to fire the "proper" parent event. in the case where no coalescing happened, it can kind of make sense that it contains just the pure event that is also the parent event?

Kuuy123 (approved):

index.html
patrickhlauke, line 969: From call just now, idea from @flackr Right after this line, something along the lines of: the parent event is the result of all these events, with a hint that UA may resample or otherwise average/modify

patrickhlauke: parent event is a representation of the events in the list, including resampling etc, to the point where the UA thinks the parent is the most representative of the raw coalesced event stream's position etc

patrickhlauke:
This is superseded by https://github.com/w3c/pointerevents/pull/440

PR #440, opened 2022-05-11
Explain relation of coalesced events to parent event (#409)

### LINKED ISSUE #409: Relationship between main pointer event and coalesced events

flackr:
This attempts to explain the relation of the coalesced events to the parent event without prescribing exactly how it is calculated with a note to clarify the implication for developers.

Supersedes https://github.com/w3c/pointerevents/pull/436

Closes https://github.com/w3c/pointerevents/issues/409

***

### :boom: Error: 502 Bad Gateway :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 25, 2022, 1:54 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fflackr%2Fpointerevents%2F062e7199015331a4fcc5567d2258625938ff9270%2Findex.html%3FisPreview%3Dtrue)

```
<html><body><h1>502 Bad Gateway</h1>
The server returned an invalid or incomplete response.
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/pointerevents%23440.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -983,9 +983,9 @@ partial interface Navigator {
             <code>pointerrawupdate</code> events, the list is a sequence of all <code>PointerEvent</code>s
             that were coalesced into this event. The "parent" trusted <code>pointermove</code> and
             <code>pointerrawupdate</code> event represents an accumulation of these coalesced events,
-            but may have additional processing to align with frame presentation. For all other trusted
-            event types, it is an empty list. Untrusted events have their <a>coalesced events list</a>
-            initialized to the value passed to the constructor.</p>
+            but may have additional processing (for example to align with the display refresh rate).
+            For all other trusted event types, it is an empty list. Untrusted events have their
+            <a>coalesced events list</a> initialized to the value passed to the constructor.</p>
 
             <div class="note">Since a trusted parent event is a summary or aggregation of the
             coalesced events, developers should only need to process either the parent events
--- a/index.html
+++ b/index.html
@@ -981,9 +981,16 @@ partial interface Navigator {
             <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events list</dfn> (a list of
             zero or more <code>PointerEvent</code>s). For trusted <code>pointermove</code> and
             <code>pointerrawupdate</code> events, the list is a sequence of all <code>PointerEvent</code>s
-            that were coalesced into this event. For all other trusted event types, it is an empty list.
-            Untrusted events have their <a>coalesced events list</a> initialized to the value passed to the
-            constructor.</p>
+            that were coalesced into this event. The "parent" trusted <code>pointermove</code> and
+            <code>pointerrawupdate</code> event represents an accumulation of these coalesced events,
+            but may have additional processing to align with frame presentation. For all other trusted
+            event types, it is an empty list. Untrusted events have their <a>coalesced events list</a>
+            initialized to the value passed to the constructor.</p>
+
+            <div class="note">Since a trusted parent event is a summary or aggregation of the
+            coalesced events, developers should only need to process either the parent events
+            or all of the coalesced events, but not both.
+            </div>
 
             <p>The events in the coalesced events list of a trusted event will have:</p>
             <ul>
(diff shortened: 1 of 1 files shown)


mustaqahmed (approved):
LGTM but do we have a better term to replace "frame representation"?  I would love to exclude any accidental interpretation around an `iframe`.

patrickhlauke (approved):

patrickhlauke:
@smaug please review this when you get a chance

flackr:
> LGTM but do we have a better term to replace "frame representation"? I would love to exclude any accidental interpretation around an `iframe`.

This is updated in the latest patch to refer to display refresh rate, also clearly identified as an example for why the values may be different but avoid implying that such alignment is spec'd.

mustaqahmed:
> > LGTM but do we have a better term to replace "frame representation"? I would love to exclude any accidental interpretation around an `iframe`.
> 
> This is updated in the latest patch to refer to display refresh rate, also clearly identified as an example for why the values may be different but avoid implying that such alignment is spec'd.

Thanks, still LGTM.

index.html
```diff
@@ -981,9 +981,16 @@ <h2><dfn>Coalesced events</dfn></h2>
...
             zero or more <code>PointerEvent</code>s). For trusted <code>pointermove</code> and
             <code>pointerrawupdate</code> events, the list is a sequence of all <code>PointerEvent</code>s
-            that were coalesced into this event. For all other trusted event types, it is an empty list.
-            Untrusted events have their <a>coalesced events list</a> initialized to the value passed to the
-            constructor.</p>
+            that were coalesced into this event. The "parent" trusted <code>pointermove</code> and
+            <code>pointerrawupdate</code> event represents an accumulation of these coalesced events,
+            but may have additional processing (for example to align with the display refresh rate).
+            For all other trusted event types, it is an empty list. Untrusted events have their
+            <a>coalesced events list</a> initialized to the value passed to the constructor.</p>
```
smaug----, line 989: I'm not sure it is clear from this that the coalesced events list contains always at least one event.

index.html
```diff
@@ -981,9 +981,16 @@ <h2><dfn>Coalesced events</dfn></h2>
...
+
+            <div class="note">Since a trusted parent event is a summary or aggregation of the
+            coalesced events, developers should only need to process either the parent events
+            or all of the coalesced events, but not both.
```
smaug----, line 993: ...this kind of hints that there should be at least one event, since otherwise this comment wouldn't work, but perhaps we should be very explicit about non-empty coalesced events list.

index.html
patrickhlauke: maybe to be overly explicit then, after `(for example to align with the display refresh rate).`, cram in an extra sentence.

```
(for example to align with the display refresh rate).
As a result, the coalesced events lists always contains one or more events.
For all other trusted event types, ...
```

patrickhlauke: `one or more events` could also be changed to `at least one event`. question of taste i guess

flackr: I've added @patrickhlauke's suggestion, with a slight clarification that it's for *these events* implying the events called out in the previous sentence. Language is complicated 😄

patrickhlauke: Thanks @flackr ... while a bit awkward ("events events events"), I think that's now quite clear

mustaqahmed: Suggesting one minor tweak: replace

> As a result, the coalesced events lists for these events always contain at least one event

with

> The coalesced events lists for these two event types always contain at least entry

Then the following sentence about "for all _other_ ..." becomes a bit clearer to me.  Wdyt?

PR #460, opened 2022-11-22
Add popover light dismiss integration

josepharhar:
I am adding a new HTML attribute called "popover" in https://github.com/whatwg/html/pull/8221

Part of this feature involves "light dismiss", where clicking outside of an element with the popover attribute causes it to be closed. This functionality was originally implemented by changing the event dispatching system of the DOM, but it was recommended to move it here instead:
https://github.com/whatwg/dom/pull/1117#issuecomment-1311928234
https://chromium-review.googlesource.com/c/chromium/src/+/4023021

***
<a href="https://pr-preview.s3.amazonaws.com/josepharhar/pointerevents/pull/460.html" title="Last updated on Feb 13, 2026, 8:50 PM UTC (fbc7da7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/460/4965307...josepharhar:fbc7da7.html" title="Last updated on Feb 13, 2026, 8:50 PM UTC (fbc7da7)">Diff</a>


patrickhlauke:
Briefly discussed at today's meeting. https://www.w3.org/2022/11/23-pointerevents-minutes.html#t03

Waiting for this to stabilise before adding to PE spec.

josepharhar:
Sounds good, I'll comment again when the popover HTML PR is done

index.html
```diff
@@ -409,6 +409,8 @@ <h2>Firing events using the <code>PointerEvent</code> interface</h2>
                 <p>If the event is {{GlobalEventHandlers/pointerdown}}, the associated device is a direct manipulation device, and the target is an {{Element}},
                    then <a>set pointer capture</a> for this <code>pointerId</code> to the target element as described in <a>implicit pointer capture</a>.
 
+                <p>Run <a href="html.spec.whatwg.org/#popover-light-dismiss">light dismiss open popovers</a> given the event and target.</p>
```
foolip: Since this is now crossing specs, the interface could be simplified to just giving the event, and HTML can get `event.target`, effectively.

foolip: In fact the algorithm in HTML already gets event target. @josepharhar do you think it makes sense to reduce it to just one argument?

josepharhar: I removed the target argument, thanks for the catch

josepharhar:
The HTML spec for this has been merged: https://html.spec.whatwg.org/multipage/popover.html#light-dismiss-open-popovers
Can we merge this PR now?

patrickhlauke:
@mustaqahmed @flackr @smaug---- as you're all closer to the "metal" here in terms of your technical experience/understanding, do you mind reviewing this PR to just double-check it's kosher and then merge?

index.html
```diff
@@ -409,6 +409,8 @@ <h2>Firing events using the <code>PointerEvent</code> interface</h2>
                 <p>If the event is {{GlobalEventHandlers/pointerdown}}, the associated device is a direct manipulation device, and the target is an {{Element}},
                    then <a>set pointer capture</a> for this <code>pointerId</code> to the target element as described in <a>implicit pointer capture</a>.
 
+                <p>Run <a href="html.spec.whatwg.org/#popover-light-dismiss">light dismiss open popovers</a> given the event.</p>
```
flackr: In the [chromium implementation, light dismiss from the escape key](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/input/keyboard_event_manager.cc;l=613?q=lightdismiss%20file:blink%5C%2F) seems to only be invoked if the default wasn't prevented[1]. However, calling it here before the event has been fired means that this will happen regardless of whether the default was prevented for pointer events. This indeed [matches the implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/input/pointer_event_manager.cc;l=211?q=lightdismiss%20file:blink%5C%2F) but are we intentionally inconsistent? Has there been discussion about this? I put together a demo at https://jsbin.com/bakoxan/edit?html,js,output to show a page which calls preventDefault on all events after a popover or dialog is open to test the differences.

Our implementation [only calls the light dismiss steps if the event is pointerdown or pointerup](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/input/pointer_event_manager.cc;l=211?q=lightdismiss%20file:blink%5C%2F). I see the steps in the html spec only care about pointerdown and pointerup, but might be worth limiting this here?

I have some concerns about the algorithm in the html spec. It seems like https://html.spec.whatwg.org/#popover-light-dismiss is replicating a form of click detection, however not taking into account many of the cases in which clicks are not generated.
- Do you really want to light dismiss popovers on the pointerup event if the finger was moved sufficiently such that a click wouldn't be generated?
- The check of whether the down target is the same as the up target seems a bit overly naive, in complex pages it's possible for the pointer to drift slightly and be over a different target element. Should this really result in not dismissing the popover even though it still generates a click?

I wonder whether it would be better to look for click events.

Apologies if these things have already been discussed I just couldn't find them with the searching that I did.

josepharhar: The reason I'm putting this here instead of in the HTML spec is that I got feedback during the review that we shouldn't be using an event listener to do the light dismiss behavior or really interacting with the event system at all, and that instead we should be going where the events are fired and instrumenting everything there. The lack of preventDefault support seems consistent with that idea.

Perhaps the escape key behavior should do the same if that is really what's preferred. I'm not sure if [CloseWatcher](https://github.com/WICG/close-watcher), which should replace the current escape key implementation, imposes any changes in behavior regarding preventDefault. @domenic

> It seems like https://html.spec.whatwg.org/#popover-light-dismiss is replicating a form of click detection, however not taking into account many of the cases in which clicks are not generated.

I agree, reusing the implementation of click events sounds better. @mfreed7 I believe that your initial implementation used mouseup and mousedown instead of click. Am I right? Was there any reason for that?

domenic: > Perhaps the escape key behavior should do the same if that is really what's preferred. I'm not sure if [CloseWatcher](https://github.com/WICG/close-watcher), which should replace the current escape key implementation, imposes any changes in behavior regarding preventDefault. @domenic

CloseWatcher works per https://wicg.github.io/close-watcher/#close-signals ; note steps 2 and 3. (Also note that per discussion in https://github.com/whatwg/html/issues/9132#issuecomment-1517053656, @emilio would prefer we be more explicit about it being keydown for Esc, instead of my current proposal's "Fire any relevant event ... If multiple such events are fired, the user agent must pick one for the purposes of the following steps.")

dontcallmedom:
This one-line PR is nearly two years old - is there any update on the discussion above that could help move it forward?

index.html
josepharhar: > I wonder whether it would be better to look for click events.

If we wanted to use the click event instead, should it be added here? https://w3c.github.io/uievents/#event-type-click

I think I have to find some place outside of the HTML spec to implement this, unless we can add some abstraction in the HTML spec for this kind of behavior. Potentially related: https://github.com/whatwg/html/issues/10762

josepharhar: I talked to Mason, and changing to click sounds OK as long as it preserves the user-facing behavior, which we have a lot of WPT coverage for. I'm not sure if it will or not. I can try experimenting with it by instrumenting the place where click events are fired from.

josepharhar:
I filed an issue about migrating from pointerdown/pointerup to the click event here: https://github.com/w3c/pointerevents/issues/542

josepharhar:
I updated this PR to use the click event instead of sending the pointerdown and pointerup events to the HTML spec.

index.html
```diff
@@ -904,6 +904,8 @@ <h4>Event dispatch</h4>
                             <div class="note">If |userEvent| was captured, |event| is dispatched to the capturing target of |userEvent| even though the {{GlobalEventHandlers/lostpointercapture}} event with the same {{PointerEvent/pointerId}} has been dispatched already.</div>
                         </li>
                     </ol>
+
+                    <p>Before a |click| event is dispatched, the <a href="html.spec.whatwg.org/#popover-light-dismiss">light dismiss open popovers</a> algorithm MUST be run given the target of the corresponding <code>pointerdown</code> event and the target of the corresponding <code>pointerup</code> event.</p>
```
mustaqahmed: I think this step should better move before (current) Step 4, saying something like:
> run the (html) algorithm with parameters <var>target</var>, the target of pointerdown and the target of pointerup.

Do we have a corresponding PR to update the HTML algorithm parameters?

josepharhar: I moved it before step 4, and I will work on a corresponding HTML spec PR

josepharhar:
I created a corresponding HTML spec PR to use click instead of pointerdown/pointerup here: https://github.com/whatwg/html/pull/11536

mustaqahmed (approved):
Looks good to me, thanks for addressing my comments.

mustaqahmed:
Because this PR depends on your the HTML PR (whatwg/html#11536), should we perhaps land that one first?

index.html
```diff
@@ -899,6 +899,10 @@ <h4>Event dispatch</h4>
                             <p>Otherwise (|event| is a <code>click</code> or <code>auxclick</code> event for which |userEvent| is a <code>pointerup</code> event that was dispatched uncaptured) let |target| be the nearest common inclusive ancestor of the corresponding <code>pointerdown</code> and <code>pointerup</code> targets in the DOM at the moment |event| is being dispatched.</p>
                         </li>
 
+                        <li>
+                          <p>If |event| is a <code>click</code> event, then <a href="https://html.spec.whatwg.org/#run-light-dismiss-activities">run light dismiss activities</a> given |event|.</p>
+                        </li>
+
```
smaug----, line 905: I don't understand how this setup works with the customizable select, which opens the popup using mousedown.

index.html
```diff
@@ -899,6 +899,10 @@ <h4>Event dispatch</h4>
                             <p>Otherwise (|event| is a <code>click</code> or <code>auxclick</code> event for which |userEvent| is a <code>pointerup</code> event that was dispatched uncaptured) let |target| be the nearest common inclusive ancestor of the corresponding <code>pointerdown</code> and <code>pointerup</code> targets in the DOM at the moment |event| is being dispatched.</p>
                         </li>
 
+                        <li>
+                          <p>If |event| is a <code>click</code> event, then <a href="https://html.spec.whatwg.org/#run-light-dismiss-activities">run light dismiss activities</a> given |event|.</p>
+                        </li>
```
smaug----, line 960: I don't quite understand how this works with customizable select's popup. That one is opened using mousedown, and then we'd a click, and that runs light dismiss. Maybe I'm just missing something

sb3nder: Hi @smaug----, elements associated with the popover would be excluded from light dismiss, much like the current implementation, I think.

josepharhar: Here is the fix for this: https://github.com/whatwg/html/pull/12008

josepharhar:
I just added a lot more text to pass more detailed information to HTML

index.html
```diff
@@ -899,6 +899,63 @@ <h4>Event dispatch</h4>
...
+
+                                <li>
+                                  <p>Set |pointerDownY| and |pointerUpY| to |event|'s <code>clientY</code>.</p>
+                                </li>
+                              </ol>
+                            </li>
+                          </ol>
+
+                          <p><a href="https://html.spec.whatwg.org/#run-light-dismiss-activities">Run light dismiss activities</a> given |pointerDownTarget|, |pointerDownX|, |pointerDownY|, |pointerUpTarget|, |pointerUpX|, and |pointerUpY|.</p>
+                        </li>
```
smaug----, line 960: Should this be step 4. Now it is kind of outside the main algorithm

josepharhar: Thanks, I moved up the run light dismiss step. How does it look?

PR #464, opened 2023-02-15
Add clarification about `pointercancel` coordinates/properties

### LINKED ISSUE #463: Coordinates of a `pointercancel` event
I think the spec should say that the coordinates of a `pointercancel` event are uninitialized (or zeros).

Chrome does not initialize the coordinates of a `pointercancel` event, which seems logical because the event is tied to a specific `pointerId` and not related to a specific pointer location.  For example, if a touch-drag causes scroll (and therefore the firing of a `pointercancel` event), the user agent absorbs the last `pointermove` events to determine the scroll amount etc and the page is not expected to know those event locations.  Developers can cache the last `pointermove` for that `pointerId` if there is any use.

We have a [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=724013) about it which we are planning to close soon.

patrickhlauke:
Initial update based on our discussion on 1 Feb https://www.w3.org/2023/02/01-pointerevents-minutes.html

Closes #463

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/464.html" title="Last updated on Feb 15, 2023, 3:59 PM UTC (1aefdf1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/464/ed733e3...1aefdf1.html" title="Last updated on Feb 15, 2023, 3:59 PM UTC (1aefdf1)">Diff</a>


patrickhlauke:
if/when merged, this will also need the `needs-wpt` label

patrickhlauke:
per the meeting today https://www.w3.org/2023/03/01-pointerevents-minutes.html we agreed to merge

mustaqahmed:
Working on a WPT now.

Issue #357, opened 2021-03-31
Clarify when lostpointercapture should fire


smaug----:
This is split from https://github.com/w3c/pointerevents/issues/75
The spec is contradicting itself whether lostpointercapture should fire before or after click.
[implicit-release-of-pointer-capture](https://w3c.github.io/pointerevents/#implicit-release-of-pointer-capture)

"if in an implicit release scenario both click and lostpointercapture events are fired, click SHOULD be fired before lostpointercapture."
(note, that is just "SHOULD")

vs

"Immediately after firing the pointerup or pointercancel events, a user agent MUST clear the pending pointer capture target override for the pointerId of the pointerup or pointercancel event that was just dispatched, and then run Process Pending Pointer Capture steps to fire lostpointercapture if necessary."
(Note, this is using "MUST")

Chromium seems to fire click first, Gecko and Webkit fire lostpointercapture first.
[Testcase](https://output.jsbin.com/zuwiwep) Click on blue.

EdgarChen:
My two cents: if the pointer capture affects the click targets, it seems to make sense to fire `click` event before `lostpointercapture`.

patrickhlauke:
Mentioned briefly in today's call. Parking this issue until #356 has been fleshed out more

smaug----:
Resolution is to fire lostpointercapture before the click, since click target is calculated already when firing pointerup - no matter whether it is implicit or explicit release.

smaug----:
This was fixed as part of https://github.com/w3c/pointerevents/pull/474

mustaqahmed:
Looks like `pointerevent_click_during_capture.html` needs to be updated: the spec now says `click` should be after `lostpointercapture` as per this [comment](https://github.com/w3c/pointerevents/issues/357#issuecomment-1601040840).

mustaqahmed:
Done with WPT.

PR #474, opened 2023-06-21
Define target and event-order for click, auxclick and contextmenu events

### LINKED ISSUE #356: Clarify what the target of the click event should be after capturing pointer events
This is split from https://github.com/w3c/pointerevents/issues/75
[Testcase](https://output.jsbin.com/zuwiwep)

We need some table here to compare the current behavior of various browser engines.
1. Mousedown on blue. Mouseup on blue.
2. Mousedown on green. Mouseup on green.
3. Mousedown on blue. Move to green. Mouseup on green.
4. Mousedown on green. Move to blue. Mouseup on blue.

mustaqahmed:
Closes #356 and #357.

***
<a href="https://pr-preview.s3.amazonaws.com/mustaqahmed/pointerevents/pull/474.html" title="Last updated on Jun 29, 2023, 9:17 PM UTC (0391b46)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/474/747bc69...mustaqahmed:0391b46.html" title="Last updated on Jun 29, 2023, 9:17 PM UTC (0391b46)">Diff</a>


index.html
```diff
@@ -650,21 +654,37 @@ <h3>The <dfn data-dfn-for="GlobalEventHandlers" data-dfn-type="event">lostpointe
...
+                      in the case of <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>. For this reason, user agents that have implemented the proposed
+                      change in [[[CSSOM-VIEW]]] only for {{PointerEvent}} MUST convert the various coordinate properties for the <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>
+                      to <code>long</code> values (as defined in the original [[[UIEVENTS]]]) using <a data-cite="ECMASCRIPT#sec-math.floor">Math.floor</a> [[ECMASCRIPT]].</p>
+                </section>
+                <section>
+                  <h4>Event dispatch</h4>
+                  <p>If a <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event <var>E</var> is fired in response to a {{GlobalEventHandlers/pointerup}} event, and
+                    the {{GlobalEventHandlers/pointerup}} event was disaptched while the pointer was captured, <var>E</var> MUST be dispatched to the target of the {{GlobalEventHandlers/pointerup}} event.
```
smaug----: disaptched

index.html
```diff
@@ -650,21 +654,37 @@ <h3>The <dfn data-dfn-for="GlobalEventHandlers" data-dfn-type="event">lostpointe
...
+                      However, this change — when applied only to {{PointerEvent}}, but not to regular {{MouseEvent}} — has proven to lead to web compatibility issues with legacy code
+                      in the case of <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>. For this reason, user agents that have implemented the proposed
+                      change in [[[CSSOM-VIEW]]] only for {{PointerEvent}} MUST convert the various coordinate properties for the <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>
+                      to <code>long</code> values (as defined in the original [[[UIEVENTS]]]) using <a data-cite="ECMASCRIPT#sec-math.floor">Math.floor</a> [[ECMASCRIPT]].</p>
+                </section>
+                <section>
+                  <h4>Event dispatch</h4>
+                  <p>If a <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event <var>E</var> is fired in response to a {{GlobalEventHandlers/pointerup}} event, and
```
smaug----: Capital E looks unusual, no?

index.html
```diff
@@ -650,21 +654,37 @@ <h3>The <dfn data-dfn-for="GlobalEventHandlers" data-dfn-type="event">lostpointe
...
+                    the {{GlobalEventHandlers/pointerup}} event was disaptched while the pointer was captured, <var>E</var> MUST be dispatched to the target of the {{GlobalEventHandlers/pointerup}} event.
+                    Otherwise, <var>E</var> MUST follow the dispatch process defined in [[UIEVENTS]].</p>
+                  <p>For the special case that a <code>contextmenu</code> event <var>E</var> is fired in response to a {{GlobalEventHandlers/pointerdown}} event (as happens on certain platforms for legacy reasons), and
```
smaug----: It is not a special case. It is just platform convention. On Linux/Gnome for example, contextmenu is triggered on down event.

index.html
```diff
@@ -650,21 +654,37 @@ <h3>The <dfn data-dfn-for="GlobalEventHandlers" data-dfn-type="event">lostpointe
...
+                    the {{GlobalEventHandlers/pointerdown}} event was disaptched while the pointer was captured, <var>E</var> MUST be dispatched to the target of the {{GlobalEventHandlers/pointerdown}} event.
```
smaug----: disaptched

index.html
```diff
@@ -650,21 +654,37 @@ <h3>The <dfn data-dfn-for="GlobalEventHandlers" data-dfn-type="event">lostpointe
...
+                    Otherwise, <var>E</var> MUST follow the dispatch process defined in [[UIEVENTS]].</p>
+                  <p>In any of the event capturing scenarios mentioned here, the <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event is dispatched to capturing target after the
```
smaug----: It is a bit hard to see how this sentence is related to the pointerdown handling. These are kind of contradicting.

index.html
mustaqahmed: Lowercased it, thanks.

To be frank, I wanted to avoid the first use of `<var>` in this spec but couldn't find a better way to point to "that event" both concisely and precisely.

index.html
mustaqahmed: Right, triggering at up is less common.  Rephrased, ptal.

index.html
mustaqahmed: Rephrased.

index.html
```diff
@@ -650,21 +654,31 @@ <h3>The <dfn data-dfn-for="GlobalEventHandlers" data-dfn-type="event">lostpointe
...
+                      change in [[[CSSOM-VIEW]]] only for {{PointerEvent}} MUST convert the various coordinate properties for the <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>
+                      to <code>long</code> values (as defined in the original [[[UIEVENTS]]]) using <a data-cite="ECMASCRIPT#sec-math.floor">Math.floor</a> [[ECMASCRIPT]].</p>
+                </section>
+                <section>
+                  <h4>Event dispatch</h4>
+                  <p>Let <var>e</var> be any <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event, and <var>t</var> be the user interaction event that caused the firing of <var>e</var>. The type of <var>t</var> depends on native platform convention: if <var>e</var> is a <code>contextmenu</code> event, <var>t</var> is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event; if <var>e</var> is a <code>click</code> or <code>auxclick</code> event, <var>t</var> is a {{GlobalEventHandlers/pointerup}} event.</p>
```
flackr: I'm not sure if this is true, or even should be specified here. I.e. isn't it up the ui-events spec what interactions can cause these events? ui-events is specific to say it must be after a mousedown if triggered by a right mouse button: https://www.w3.org/TR/uievents/#event-type-click:~:text=When%20the%20contextmenu%20event%20is%20triggered%20by%20right%20mouse%20button%2C%20the%20contextmenu%20event%20MUST%20be%20dispatched%20after%20the%20mousedown%20event.

Also doesn't the section before say that events can be generated by a non-pointing device.. if so must these fire pointerdown/pointerup?

Maybe it would be better to say if ther user interaction which caused the firing of e is a pointerevent, t, then ...

index.html
flackr, line 1: This also should update the [section on implicit release of pointer capture](https://w3c.github.io/pointerevents/#implicit-release-of-pointer-capture) where it says:

> If the user agent supports firing the click or auxclick event, and if in an implicit release scenario both that (click or auxclick) event and a [lostpointercapture](https://w3c.github.io/pointerevents/#dfn-lostpointercapture) event are fired, the click or auxclick event SHOULD be fired before [lostpointercapture](https://w3c.github.io/pointerevents/#dfn-lostpointercapture).

index.html
mustaqahmed: Done.

I left the platform convention as a non-normative text, with the thought that _some_ spec should mention this.  What do you think?

index.html
mustaqahmed: Good catch, thanks!  Done as follows:
- Removed the quoted para from Sec 9.5 because the previous para already has "run process pending pointer capture", and in this PR we are isolating click-like event dispatch as a special case.
- Excluded click-like events in Sec 4.1.3 algorithm.
- Mentioned click-like events in Sec 4.2.11.

index.html
```diff
@@ -951,10 +965,6 @@ <h3><dfn data-lt="implicitly release the pointer capture|implicitly releasing po
                After running <a>process pending pointer capture</a> steps,
                if the pointer supports hover, user agent MUST also send corresponding boundary events necessary
                to reflect the current position of the pointer with no capture.</p>
-            <p>If the user agent supports firing the <code>click</code> or <code>auxclick</code> event,
-               and if in an implicit release scenario both that (<code>click</code> or <code>auxclick</code>)
-               event and a {{GlobalEventHandlers/lostpointercapture}} event are fired,
-               the <code>click</code> or <code>auxclick</code> event SHOULD be fired before {{GlobalEventHandlers/lostpointercapture}}.</p>
             <p>When the <a>pointer capture target override</a> is no longer [=connected=] [[DOM]],
```
mustaqahmed, line 975: Just spotted that the "target override no longer connected" could apply to click-like events too!  I will update the PR shortly.

index.html
```diff
@@ -645,26 +649,41 @@ <h3>The <dfn data-dfn-for="GlobalEventHandlers" data-dfn-type="event">gotpointer
...
+                      in the case of <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>. For this reason, user agents that have implemented the proposed
+                      change in [[[CSSOM-VIEW]]] only for {{PointerEvent}} MUST convert the various coordinate properties for the <code>click</code>, <code>auxclick</code>, and <code>contextmenu</code>
+                      to <code>long</code> values (as defined in the original [[[UIEVENTS]]]) using <a data-cite="ECMASCRIPT#sec-math.floor">Math.floor</a> [[ECMASCRIPT]].</p>
+                </section>
+                <section>
+                  <h4>Event dispatch</h4>
+                  <p>If a <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event |e| is fired in response to a user interaction event |t|, and |t| is a <code>PointerEvent</code> which was dispatched while the corresponding pointer was captured:</p>
+                  <ul>
+                    <li>Event |e| MUST be dispatched to the capturing target of the |t|.  However, if the capturing target is no longer [=connected=] [[DOM]], |e| MUST be dispatched to the document.</li>
```
flackr: It seems odd to me that we dispatch the event to the document. Looks like without capture if the target node is removed we don't dispatch `click` or `auxclick` at all https://jsbin.com/joserud/edit?html,css,js,output ([spec](https://www.w3.org/TR/uievents/#events-mouseevent-event-order:~:text=If%20the%20event%20target%20(e.g.%20the%20target%20element)%20is%20removed%20from%20the%20DOM%20during%20the%20mouse%20events%20sequence%2C%20the%20remaining%20events%20of%20the%20sequence%20MUST%20NOT%20be%20fired%20on%20that%20element.))

Oddly, we do still calculate the new contextmenu target and dispatch the contextmenu.

TLDR; I feel like we should match this behavior - otherwise capturing to the same element will have different behavior even when the release event is over that element.

mustaqahmed: Thanks for noting this.

> Oddly, we do still calculate the new contextmenu target and dispatch the contextmenu.

By "calculate" you meant "hittesting" the `contextmenu` event, right?  That's what I found on both Chrome and Firefox.  On Windows, right-clicking near the lower edge of `#c#` sends the event to `log`!

> TLDR; I feel like we should match this behavior - otherwise capturing to the same element will have different behavior even when the release event is over that element.

Please elaborate the scope of "this behavior": did you mean `contextmenu` event would be hittested even for the capturing case?

flackr: The simplest least intrusive thing to spec would be that we only define special behavior when there is a capturing target that is still in the dom - otherwise we can back to the standard ui events targeting behavior - which is that we don't generate click and contextmenu is independently targeted.

I haven't tested yet - but I assume with regular pointerevents when the capturing element is removed any subsequent events fall back to the standard non-captured targeting behavior.

mustaqahmed: > The simplest least intrusive thing to spec would be that we only define special behavior when there is a capturing target that is still in the dom - otherwise we can back to the standard ui events targeting behavior - which is that we don't generate click and contextmenu is independently targeted.

Done, ptal.

> I haven't tested yet - but I assume with regular pointerevents when the capturing element is removed any subsequent events fall back to the standard non-captured targeting behavior.

Element removal causes firing of `lostpointercapture` as tested in `pointerevent_lostpointercapture_for_disconnected_node.html`.  This is defined in Sec9.5 but it should be clear that an implicit release could be caused by target removal!  I will handle this in a separate PR.

index.html
```diff
@@ -645,26 +649,43 @@ <h3>The <dfn data-dfn-for="GlobalEventHandlers" data-dfn-type="event">gotpointer
...
-                   to <code>long</code> values (as defined in the original [[[UIEVENTS]]]) using <a data-cite="ECMASCRIPT#sec-math.floor">Math.floor</a> [[ECMASCRIPT]].</p>
+                  <a data-cite="uievents/#event-type-auxclick">auxclick</a> and <a data-cite="uievents/#event-type-contextmenu">contextmenu</a>
+                  events defined in [[UIEVENTS]].  These events are typically tied to user interface activation, and are fired even from non-pointer input devices, such as keyboards.</p>
+                <p>These events MUST be of type <code>PointerEvent</code>, and are subject to the additional requirements mentioned in the rest of this section.</p>
+                <section>
+                  <h4>Attributes (other than coordinates)</h4>
+                  <p>For these events, all <code>PointerEvent</code> specific attributes (defined in this spec) other than <code>pointerId</code> and <code>pointerType</code> MUST have their default values. In addition:</p>
+                  <ul>
```
smaug----, line 663: But PointerEvent specific attributes aren't coordinates.  So we don't need "(other than coordinates)"

smaug---- (approved):
Perhaps we can land this and tweak more if needed.

index.html
```diff
@@ -395,7 +395,7 @@ <h2>The <dfn>primary pointer</dfn></h2>
             <section>
                 <h2>Firing events using the <code>PointerEvent</code> interface</h2>
                 <p>To <dfn>fire a pointer event</dfn> named |e| means to [=fire an event=] named |e| using <a>PointerEvent</a> whose attributes are set as defined in {{PointerEvent}} Interface and <a>Attributes and Default Actions</a>.</p>
-                <p>If the event is not {{GlobalEventHandlers/gotpointercapture}} or {{GlobalEventHandlers/lostpointercapture}}, run the <a>process pending pointer capture</a> steps for this <code>PointerEvent</code>.
+                <p>If the event is not a {{GlobalEventHandlers/gotpointercapture}}, {{GlobalEventHandlers/lostpointercapture}}, <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event, run the <a>process pending pointer capture</a> steps for this <code>PointerEvent</code>.
```
flackr, line 398: Given that we want to release pointer capture before click, auxclick or contextmenu, shoudln't we keep this here?

flackr (approved):
Looks good with nit.

index.html
mustaqahmed: I think excluding these events here is reasonable for a few reasons:
- As per the "event dispatch" subsection we added in this PR, the dispatch for click, auxclick and contextmenu is completely separate from all other PE types.
- Because this section implicitly covers PEs that are fired from non-pointer devices (such as keyboard click), we need to isolate these 3 events anyway.  Before this PR, the isolation was very vague (to be precise, "run the process pending pointer capture steps **for this PointerEvent**" doesn't apply because keyboard click has id=-1).
- Pending pointer capture gets resolved before we fire any of these 3 events I believe.

Do you see a scenario where the last point does not hold?

mustaqahmed: Landing it now, we can cover corner cases separately.

mustaqahmed:
The WPTs are covered through #356 and #357.

PR #479, opened 2023-08-30
Update README.md to point to latest editor's draft

mustaqahmed:
The spec link in README.md has been pointing to the published version.  With PEWG regularly updating the spec to fix or tune outdated text, the published version seems a bit stale most of the time.


patrickhlauke (approved):

patrickhlauke:
think this is safe to do, since the editor's draft also links back to the published version if somebody's keen to see that one

PR #480, opened 2023-08-30
Add clarification about coalesced events and secure/non secure context

### LINKED ISSUE #472: consider to change the behavior of getCoalescedEvents on non-SecureContext
https://github.com/w3c/pointerevents/pull/318 landed quite awhile ago, but it isn't implemented, at least not in Firefox nor in Chrome.
We could change that so that in non-SecureContext the method was still there, but it would always just return the event itself, and not any coalesced events. That would be hopefully less risky change than unshipping the API altogether from non-SecureContext.

patrickhlauke:
Closes https://github.com/w3c/pointerevents/issues/472

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/480.html" title="Last updated on Aug 30, 2023, 2:58 PM UTC (2cb27cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/480/978bfa2...2cb27cf.html" title="Last updated on Aug 30, 2023, 2:58 PM UTC (2cb27cf)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1123,13 +1123,18 @@ partial interface Navigator {
 
             <p>A <a>PointerEvent</a> has an associated <dfn>coalesced events list</dfn> (a list of
             zero or more <code>PointerEvent</code>s). For trusted {{GlobalEventHandlers/pointermove}} and
-            {{GlobalEventHandlers/pointerrawupdate}} events, the list is a sequence of all <code>PointerEvent</code>s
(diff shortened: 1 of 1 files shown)


patrickhlauke:
Discussed in meeting today, since usage is low, we don't even need to address this.

PR #485, opened 2023-10-05
Use `float` instead of `double` for `altitudeAngle` and `azimuthAngle`

jrandolf-google:
Altitude angle and azimuth angle are computed always computed with 32-bit floating points. For example, on macOS, https://developer.apple.com/documentation/appkit/nsevent/1534226-tilt. A priori, using a double doesn't offer more precision.

[EDIT: On macOS, the NSPoint struct uses `double` when the OS is x64, so this implies double accuracy is correct]

***
<a href="https://pr-preview.s3.amazonaws.com/jrandolf/pointerevents/pull/485.html" title="Last updated on Oct 9, 2023, 1:06 AM UTC (305c9ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/485/3943f69...jrandolf:305c9ac.html" title="Last updated on Oct 9, 2023, 1:06 AM UTC (305c9ac)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -205,8 +205,8 @@ dictionary PointerEventInit : MouseEventInit {
     long        tiltX;
     long        tiltY;
     long        twist = 0;
-    double      altitudeAngle;
-    double      azimuthAngle;
+    float       altitudeAngle;
+    float       azimuthAngle;
     DOMString   pointerType = "";
     boolean     isPrimary = false;
     sequence&lt;PointerEvent> coalescedEvents = [];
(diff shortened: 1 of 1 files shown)


mustaqahmed:
While I agree that the precision of a `double` may not be required here, it doesn't hurt either, right?  I am worried about the non-zero compat risk that some code in the wild could break because of some assumption around precision.

More on that compat point: we picked `double` here from the [TouchEvent spec](https://w3c.github.io/touch-events/#touch-interface) (where it has been like this for years).  What if there is a comparison between altitude angles of a TouchEvent and the corresponding PointerEvent that assumes an 8-digit precision (i.e. "finer" than `float`)?  This may sound impossible but consider this closely related experience: we failed to even [add precision to `click` coordinates](https://github.com/w3c/pointerevents/issues/100#issuecomment-818870316) because the integer coordinates of a MouseEvent was being compared with the coordinates of a `click` in a popular mapping website!

Without a clear benefit, we should avoid such a risk IMHO.

jrandolf:
I think there is a distinct difference here compared to the MouseEvent issue. In particular, what we are doing is truncating the `double` in `TouchEvent` which is only guaranteed to be accurate up to `float32` since we know browsers derive those values from OSes provide `float32`. In particular, if the `double` `TouchEvent` angles _does not_ match the `PointerEvent` `float` angles, then that means there is a problem with the implementation.

Issue #487, opened 2023-10-18
What happens when the pointer capture target is removed from DOM?


mustaqahmed:
DOM changes potentially affects both the variables [`pending pointer capture target override`](https://w3c.github.io/pointerevents/#dfn-pending-pointer-capture-target-override) and 
[`pointer capture target override`](https://w3c.github.io/pointerevents/#dfn-pointer-capture-target-override).  The current spec doesn't say what happens to them after a DOM change.

1. Should we change an element reference to `null` when the element gets disconnected so that the (pending) capture becomes ineffective?
2. Should we keep the element reference as-is, so any event dispatch to it fails silently until the element is re-attached?
3. Should we follow the recent consensus in https://github.com/web-platform-tests/interop/issues/380 that the reference will be updated to the still-connected ancestors?

flackr:
I think the most reasonable expectation is that we only clear the pending pointer capture target override and then in the next [Process pending pointer capture](https://w3c.github.io/pointerevents/#process-pending-pointer-capture) we will dispatch lostpointercapture.

mustaqahmed:
To add a bit more context from PEWG discussion today, we agreed that the second para in [Implicit release of pointer capture](https://w3c.github.io/pointerevents/#implicit-release-of-pointer-capture) section covers the issue except for the timing for the `lostpointercapture` event dispatch.  The current text seems to suggest the dispatch is synchronous with DOM updates, bypassing the lazy dispatch in [Process pending pointer capture](https://w3c.github.io/pointerevents/#process-pending-pointer-capture).

We will update that second para to utilize the lazy dispatch.

flackr:
BTW I think the state to track this may be slightly more complex. [Process pending pointer capture](https://w3c.github.io/pointerevents/#process-pending-pointer-capture) says:

> If the [pointer capture target override](https://w3c.github.io/pointerevents/#dfn-pointer-capture-target-override) for this pointer is set and is not equal to the [pending pointer capture target override](https://w3c.github.io/pointerevents/#dfn-pending-pointer-capture-target-override), then fire a pointer event named [lostpointercapture](https://w3c.github.io/pointerevents/#dfn-lostpointercapture) at the [pointer capture target override](https://w3c.github.io/pointerevents/#dfn-pointer-capture-target-override) node.

[...]

PR #490, opened 2023-10-27
Handle pointer capture node removal asynchronously

flackr:
This explains what happens when the current pointer capture target is removed or the pending pointer capture target is removed - addressing the concern raised in #487. The pointer capture is updated during the next process pending pointer capture which occurs before the dispatch of the next pointer event.

***
<a href="https://pr-preview.s3.amazonaws.com/flackr/pointerevents/pull/490.html" title="Last updated on Nov 2, 2023, 6:36 PM UTC (fc7616e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/490/24de93c...flackr:fc7616e.html" title="Last updated on Nov 2, 2023, 6:36 PM UTC (fc7616e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1096,14 +1096,14 @@ partial interface Navigator {
                 if the pointer supports hover, user agent MUST also send corresponding boundary events necessary
                 to reflect the current position of the pointer with no capture.</p>
             <p>When the <a>pointer capture target override</a> is no longer [=connected=] [[DOM]],
-                the <a>pending pointer capture target override</a> node SHOULD be cleared and
                 the <a>pointer capture target override</a> SHOULD be set to the document.</p>
-                <div class="note">
-                    This will result in a {{GlobalEventHandlers/lostpointercapture}} event corresponding to the captured pointer being fired at the document
-                    during the next <a>Process pending pointer capture</a>.
-                </div>
             <p>When the <a>pending pointer capture target override</a> is no longer [=connected=] [[DOM]],
                 the <a>pending pointer capture target override</a> node SHOULD be cleared.</p>
+            <div class="note">
+                The previous two paragraphs result in a {{GlobalEventHandlers/lostpointercapture}} event
+                corresponding to the captured pointer being fired at the document
+                during the next <a>Process pending pointer capture</a> after the capture node is removed.
+            </div>
             <p>When a pointer lock [[PointerLock]] is successfully applied on an element, the user agent MUST run the steps as if the <a data-lt='Element.releasePointerCapture'>releasePointerCapture()</a> method has been called if any element is set to be captured or pending to be captured.
 
         </section>
--- a/index.html
+++ b/index.html
@@ -1096,8 +1096,14 @@ partial interface Navigator {
                 if the pointer supports hover, user agent MUST also send corresponding boundary events necessary
                 to reflect the current position of the pointer with no capture.</p>
             <p>When the <a>pointer capture target override</a> is no longer [=connected=] [[DOM]],
-                the <a>pending pointer capture target override</a> and <a>pointer capture target override</a> nodes SHOULD be cleared
-                and also a PointerEvent named {{GlobalEventHandlers/lostpointercapture}} corresponding to the captured pointer SHOULD be fired at the document.</p>
+                the <a>pending pointer capture target override</a> node SHOULD be cleared and
+                the <a>pointer capture target override</a> SHOULD be set to the document.</p>
+                <div class="note">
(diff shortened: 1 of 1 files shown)


patrickhlauke:
this looks good to me (covers option 1 suggested by @mustaqahmed on #487)

index.html
```diff
@@ -1096,8 +1096,14 @@ <h2><dfn data-lt="implicitly release the pointer capture|implicitly releasing po
                 if the pointer supports hover, user agent MUST also send corresponding boundary events necessary
                 to reflect the current position of the pointer with no capture.</p>
             <p>When the <a>pointer capture target override</a> is no longer [=connected=] [[DOM]],
-                the <a>pending pointer capture target override</a> and <a>pointer capture target override</a> nodes SHOULD be cleared
-                and also a PointerEvent named {{GlobalEventHandlers/lostpointercapture}} corresponding to the captured pointer SHOULD be fired at the document.</p>
+                the <a>pending pointer capture target override</a> node SHOULD be cleared and
```
mustaqahmed: Clearing the `pending target override` here would drop a `setPointerCapture()` call that is made just before the deletion of current `target override`.  Not doing anything to the pending state here should work fine, right?

index.html
```diff
@@ -1096,8 +1096,14 @@ <h2><dfn data-lt="implicitly release the pointer capture|implicitly releasing po
...
+                the <a>pointer capture target override</a> SHOULD be set to the document.</p>
+                <div class="note">
+                    This will result in a {{GlobalEventHandlers/lostpointercapture}} event corresponding to the captured pointer being fired at the document
```
mustaqahmed: Another corner case: should we explicitly disallow `document.setPointerCapture`?

mustaqahmed: Alternatively, we can say here:

> This will result in a `lostpointercapture` event corresponding to the captured pointer being fired at the document during the next Process pending pointer capture, _provided that there is no pending setPointerCapture request for the document itself_.

mustaqahmed: Please ignore this thread, sorry: `document` is a non-element node, so it has no `setPointercapture`.

index.html
flackr: Good point, we should only clear the pending pointer capture target if it was the removed node - which we would do anyways from the following point. I can remove this here and move the note below both paragraphs as it only follows from the combination of updating both things.

flackr: Done.

index.html
```diff
@@ -1096,8 +1096,14 @@ <h2><dfn data-lt="implicitly release the pointer capture|implicitly releasing po
                 if the pointer supports hover, user agent MUST also send corresponding boundary events necessary
                 to reflect the current position of the pointer with no capture.</p>
             <p>When the <a>pointer capture target override</a> is no longer [=connected=] [[DOM]],
-                the <a>pending pointer capture target override</a> and <a>pointer capture target override</a> nodes SHOULD be cleared
-                and also a PointerEvent named {{GlobalEventHandlers/lostpointercapture}} corresponding to the captured pointer SHOULD be fired at the document.</p>
+                the <a>pointer capture target override</a> SHOULD be set to the document.</p>
+            <p>When the <a>pending pointer capture target override</a> is no longer [=connected=] [[DOM]],
+                the <a>pending pointer capture target override</a> node SHOULD be cleared.</p>
+            <div class="note">
+                The previous two paragraphs result in a {{GlobalEventHandlers/lostpointercapture}} event
```
mustaqahmed, line 1103: I think this note clarifies the outcome of _only_ the first paragraph modified here.  Regardless of what happens to the "pending state", the next invocation of "Process pending pointer capture" will fire a `lostpointercapture` to document.

I suggest making this a normative text at the end of the first para above.

flackr: This note clarifies the outcome of *both* paragraphs. If we only had the first paragraph, the pending pointer capture target override would still be the removed element, so we would attempt to re-apply pointer capture to the removed element.

> I suggest making this a normative text at the end of the first para above.

I don't think this is standard practice, this note is clarification as to what will happen as a result of the normative requirements. We don't normally clarify with normative statements as it could result in inconsistencies between the normative requirement and the clarifying paragraph and the paragraph.

In particular, this clarifying paragraph only describes one very specific case - but not the plethora of cases that can happen, e.g. if you request capture again on the node after re-adding the node, you'll still get a lostpointercapture.

mustaqahmed: Sounds good.

mustaqahmed (approved):

smaug---- (approved):

patrickhlauke:
Does this need a WPT?

flackr:
This should be tested by pointerevents/pointerevent_lostpointercapture_for_disconnected_node.html

Issue #477, opened 2023-07-21
Clarify `pointerleave` and `pointerout` events when first pointer move after removing an element under the pointer


masayuki-nakano:
I worked on fixing the failure in Firefox of `pointerevent_after_target_removed.html`. However, its expectation does not make sense. See <https://github.com/web-platform-tests/interop/issues/380> for the detail.

The main issue is, the spec does not define `pointerleave` and `pointerout` should be fired on which element(s) or should not be fired at the first move after an element under the pointer is removed.

Some definitions indicate that `pointerleave`, `pointerout`, `pointerenter` and `pointerover` should not be fired immediately (without a pointer move) when an element under the pointer is removed by a DOM API or hidden/moved by a style change. Then, web apps may want to listen to `pointerleave` and `pointerout` at next pointer move to clear something set at `pointerenter` or `pointerover`.

The spec does **not** disallow to dispatch `pointerout` on removed element (which is the last `pointerover`'ed element). Safari does that, and it seems reasonable.

The big problem here is, `pointerleave`. `pointerenter` must have been fired on the removed element and its ancestors. From web apps point of view, `pointerleave` should be fired on all elements which `pointerenter` was fired on and  are not ancestors of the element currently under the pointer. However, it sounds like terrible. To do that, browsers need to store all elements which the browser dispatched `pointerenter`, but this means that nearly root ancestors will be stored for a long time, e.g., the `<body>` element. And the dispatching code becomes complicated because DOM API may remove only some ancestors with re-connecting an ancestor to the same point.

@EdgarChen, @smaug----

mustaqahmed:
Below are my comments/questions to aid the discussion:

> I worked on fixing the failure in Firefox of `pointerevent_after_target_removed.html`. However, its expectation does not make sense. See [web-platform-tests/interop#380](https://github.com/web-platform-tests/interop/issues/380) for the detail.

Hmm, I added this test to match the expectation in a user reported [Chrome bug](https://crbug.com/1147998).  I haven't looked into all the details you mentioned above, but I can see that `pointerover/out` events were not mentioned in the Chrome bug, so I need to check if I accidentally made the test an "overfit"!

> The main issue is, the spec does not define `pointerleave` and `pointerout` should be fired on which element(s) or should not be fired at the first move after an element under the pointer is removed.

Yes, and one reason is that the text in PointerEvents depends on the hand-wavy [definition](https://w3c.github.io/uievents/#events-mouseevent-event-order) in UI Event Spec.

[...]

PR #491, opened 2023-10-31
Explain how a removed DOM node should be handled for boundary events.

flackr:
This explains how the previous target is tracked when the previous target is removed from the DOM for #477.

***
<a href="https://pr-preview.s3.amazonaws.com/flackr/pointerevents/pull/491.html" title="Last updated on Nov 20, 2023, 7:41 PM UTC (6a69eda)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/491/24de93c...flackr:6a69eda.html" title="Last updated on Nov 20, 2023, 7:41 PM UTC (6a69eda)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -423,7 +423,17 @@ interface PointerEvent : MouseEvent {
                 <p>If the event is {{GlobalEventHandlers/pointerdown}}, the associated device is a direct manipulation device, and the target is an {{Element}},
                     then <a>set pointer capture</a> for this <code>pointerId</code> to the target element as described in <a>implicit pointer capture</a>.
 
-                <p>Fire the event to the determined target. The user agent SHOULD treat the target as if the pointing device has moved over it for the purpose of <a data-cite="uievents/#events-mouseevent-event-order">ensuring event ordering</a> [[UIEVENTS]].</p>
+                <p>Fire the event to the determined target.
+                    The user agent SHOULD treat the target as if the pointing device has moved over it from the |previous target| (or if the |over child| flag is set, over an anonymous child of |previous target|)
+                    for the purpose of <a data-cite="uievents/#events-mouseevent-event-order">ensuring event ordering</a> [[UIEVENTS]].</p>
+
+                <p>Save the determined target as the |previous target| for the given pointer,
+                    and reset the |over child| flag to <code>false</code>.
+                    If the |previous target| at any point will no longer be [=connected=] [[DOM]],
+                    update the |previous target| to the nearest still [=connected=] [[DOM]] parent
+                    following the event path corresponding to dispatching events to the previous target,
+                    and set the |over child| flag to <code>true</code>.
+                </p>
 
                 <div class="note">Using the <a>pointer capture target override</a> as the target instead of the normal hit-test result may fire some boundary events, as defined by [[UIEVENTS]]. This is the same as the pointer leaving its previous target and entering this new capturing target. When the capture is released, the same scenario may happen, as the pointer is leaving the capturing target and entering the hit-test target.</div>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -423,7 +423,17 @@ <h3>Firing events using the <code>PointerEvent</code> interface</h3>
...
-                <p>Fire the event to the determined target. The user agent SHOULD treat the target as if the pointing device has moved over it for the purpose of <a data-cite="uievents/#events-mouseevent-event-order">ensuring event ordering</a> [[UIEVENTS]].</p>
+                <p>Fire the event to the determined target.
+                    The user agent SHOULD treat the target as if the pointing device has moved over it from the |previous target| (or if the |over child| flag is set, over an anonymous child of |previous target|)
+                    for the purpose of <a data-cite="uievents/#events-mouseevent-event-order">ensuring event ordering</a> [[UIEVENTS]].</p>
+
+                <p>Save the determined target as the |previous target| for the given pointer,
+                    and reset the |over child| flag to <code>false</code>.
+                    If the |previous target| at any point will no longer be [=connected=] [[DOM]],
+                    update the |previous target| to the nearest still [=connected=] [[DOM]] parent
+                    following the event path corresponding to dispatching events to the previous target,
```
flackr, line 436: Note that this following the event path part is a bit hand wavy. Maybe we need to expose a method which encapsulates step 5.9 of [dispatching events](https://dom.spec.whatwg.org/#dispatching-events). One of the complications to doing that is the [get the parent](https://dom.spec.whatwg.org/#get-the-parent) algorithm can depend on the particular event being dispatched. I think it is the case that for all pointer events (and mouse events) the path is expected to be the same, but technically nothing stops a particular event from having a different path.

patrickhlauke:
@smaug---- @mustaqahmed thoughts on this PR?

index.html
```diff
@@ -423,7 +423,17 @@ <h3>Firing events using the <code>PointerEvent</code> interface</h3>
                 <p>If the event is {{GlobalEventHandlers/pointerdown}}, the associated device is a direct manipulation device, and the target is an {{Element}},
                     then <a>set pointer capture</a> for this <code>pointerId</code> to the target element as described in <a>implicit pointer capture</a>.
 
-                <p>Fire the event to the determined target. The user agent SHOULD treat the target as if the pointing device has moved over it for the purpose of <a data-cite="uievents/#events-mouseevent-event-order">ensuring event ordering</a> [[UIEVENTS]].</p>
+                <p>Fire the event to the determined target.
+                    The user agent SHOULD treat the target as if the pointing device has moved over it from the |previous target| (or if the |over child| flag is set, over an anonymous child of |previous target|)
```
smaug----: What is an anonymous child?

flackr: The idea is to treat the pointer as if it is not *yet* over the parent. E.g. if you have
```html
<div id="a">
  <div id="b">Pointer over</div>
</div>
```

And you remove #b while the pointer is over it, the next boundary update *should* dispatch pointerover #a. If we implicitly set the pointer to be over #a then we will never dispatch this - so I'm proposing that we have some state that tracks that it is like being over a child of #a - but one which doesn't actually exist in the dom to receive events - for the sake of being able to dispatch the correct boundary events on the next update.

flackr: Happy to bikeshed on better ways to describe this / or define and explain a term for this.

flackr: I've replaced the "anonymous child" mention with a flag that directly states the over event is needed.

index.html
smaug----: I guess it is better to land this hand wavy patch and maybe file a followup to think about how to spec it in a more algorithmic way.

smaug---- (approved):

mustaqahmed:
Removing the need-wpt label because #477 has it already.

PR #494, opened 2024-01-04
Rewrite the dispatch algorithm for click-like events

### LINKED ISSUE #492: Clarify `mousedown` event target if the preceding `pointerdown` event listener removes the target
https://github.com/web-platform-tests/interop/issues/380#issuecomment-1787398235

Spinning out from the above issue.

Currently, only Chrome dispatches a `mousedown` event to parent element when a `pointerdown` event listener removed the event target.

**I** think that Chrome's behavior may lead odd behavior because user tried to click under the pointer, not its parent (or ancestor). However, the `mousedown` may cause running event listeners of them. This must be trouble if the element is not visually in the parent or ancestor element.

On the other hand, Safari dispatches a `mousedown` event on the removed child element, but I think that this is not required.

In my understanding, `mousedown` event should not be fired if the preceding `pointerdown` event removed the target. And I think this can be applied to other sets of a pointer event and a fallback mouse event.

I also found one odd thing of Chrome's behavior.
https://jsfiddle.net/d_toybox/m18jpnae/
```js
e.addEventListener("pointerdown", event => {
  event.target.remove();
});
addEventListener("click", () => alert("clicked"));
```
https://jsfiddle.net/d_toybox/m18jpnae/1/
```js
e.addEventListener("mousedown", event => {
  event.target.remove();
});
addEventListener("click", () => alert("clicked"));
```
In the first case, Chrome dispatches `mousedown` on the parent, and then, `mouseup` on the parent too. Therefore, `click` event should be fired from UI Events point of view. However, Chrome does not dispatch it. I guess Chrome fixes the candidate of the `click` event at dispatching `pointerdown` and the removal of the event target causes stopping dispatching it.

@mustaqahmed, @smaug----, @garykac

mustaqahmed:
Closes #492.

***
<a href="https://pr-preview.s3.amazonaws.com/mustaqahmed/pointerevents/pull/494.html" title="Last updated on Jan 31, 2024, 4:58 PM UTC (5b5016c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/494/dcca3e6...mustaqahmed:5b5016c.html" title="Last updated on Jan 31, 2024, 4:58 PM UTC (5b5016c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -813,13 +813,12 @@ function tilt2spherical(tiltX, tiltY) {
                     <h4>Event dispatch</h4>
                     <p>A <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event |e| MUST follow the dispatch process defined in [[UIEVENTS]] except when <em>all</em> of the following conditions are met:</p>
                     <ul>
-                        <li>event |e| is fired in response to a user interaction event |t|,</li>
-                        <li>event |t| is a <code>PointerEvent</code>,</li>
+                        <li>event |e| is fired in response to a user interaction event |t| which is a <code>PointerEvent</code>,</li>
                         <li>event |t| was dispatched while the corresponding pointer was captured, and</li>
                         <li>the capturing target of event |t| is [=connected=] [[DOM]] at the moment event |e| is being dispatched.</li>
                     </ul>
                     <p>When these conditions are met, event |e| MUST be dispatched to the capturing target of event |t|, and event |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</p>
-                    <div class="note">When |t| is a <code>PointerEvent</code>, its type depends on native platform convention: if |e| is a <code>contextmenu</code> event, |t| is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event; if |e| is a <code>click</code> or <code>auxclick</code> event, |t| is a {{GlobalEventHandlers/pointerup}} event.</div>
+                    <div class="note">When |e| is a <code>click</code> or <code>auxclick</code> event, |t| is always a {{GlobalEventHandlers/pointerup}} event.  When |e| is a <code>contextmenu</code> event, |t| is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event, depending on native platform convention.</div>
                 </section>
             </section>
         </section>
--- a/index.html
+++ b/index.html
@@ -811,14 +811,22 @@ function tilt2spherical(tiltX, tiltY) {
                 </section>
                 <section>
                     <h4>Event dispatch</h4>
-                    <p>A <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event |e| MUST follow the dispatch process defined in [[UIEVENTS]] except when <em>all</em> of the following conditions are met:</p>
-                    <ul>
-                        <li>event |e| is fired in response to a user interaction event |t| which is a <code>PointerEvent</code>,</li>
-                        <li>event |t| was dispatched while the corresponding pointer was captured, and</li>
-                        <li>the capturing target of event |t| is [=connected=] [[DOM]] at the moment event |e| is being dispatched.</li>
-                    </ul>
-                    <p>When these conditions are met, event |e| MUST be dispatched to the capturing target of event |t|, and event |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</p>
-                    <div class="note">When |e| is a <code>click</code> or <code>auxclick</code> event, |t| is always a {{GlobalEventHandlers/pointerup}} event.  When |e| is a <code>contextmenu</code> event, |t| is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event, depending on native platform convention.</div>
+                    <p>A <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event |e| MUST follow the dispatch process defined in the [[UIEVENTS]] spec except for the overrides defined in the algorithm below:</p>
+                    <ol>
+                        <li>Let |t| be the user interaction event that caused the firing of |e|.</li>
+                        <div class="note">
+                            <p>Event |t| could be a non-<code>PointerEvent</code>; for example, it is a <code>KeyboardEvent</code> when a <code>click</code> event dispatch is caused by hitting the spacebar on a checkbox element.<p>
+                            <p>When |t| is a <code>PointerEvent</code>, |t| is a {{GlobalEventHandlers/pointerup}} except for a <code>contextmenu</code> event where |t| is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event depending on native platform convention.</p>
+                        </div>
+
+                        <li>If |t| is not a <code>PointerEvent</code>, the dispatch of |e| MUST follow the [[UIEVENTS]] spec.</li>
+
+                        <li>Otherwise, if |t| was dispatched while the corresponding pointer was captured, and the capturing target of event |t| is [=connected=] [[DOM]] at the moment event |e| is being dispatched, then the target of |e| MUST be the capturing target of event |t|, and |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</li>
+
+                        <li>Otherwise, if |e| is a <code>contextmenu</code> event, the target of |e| MUST be the target of |t|.</li>
+
+                        <li>Otherwise (|e| is a <code>click</code> or <code>auxclick</code> event): let |td| and |tu| be the targets of the corresponding <code>pointerdown</code> and <code>pointerup</code> events respectively.  If both |td| and |tu| are [=connected=] [[DOM]] at the moment event |e| is being dispatched, the target of |e| MUST be the nearest common inclusive ancestor of |td| and |tu| in the DOM; otherwise, |e| is not disaptched at all.</li>
+                    </ol>
                 </section>
             </section>
         </section>
--- a/index.html
+++ b/index.html
@@ -825,7 +825,7 @@ function tilt2spherical(tiltX, tiltY) {
 
                         <li>Otherwise, if |e| is a <code>contextmenu</code> event, the target of |e| MUST be the target of |t|.</li>
 
-                        <li>Otherwise (|e| is a <code>click</code> or <code>auxclick</code> event): let |td| and |tu| be the targets of the corresponding <code>pointerdown</code> and <code>pointerup</code> events respectively.  If both |td| and |tu| are [=connected=] [[DOM]] at the moment event |e| is being dispatched, the target of |e| MUST be the nearest common inclusive ancestor of |td| and |tu| in the DOM; otherwise, |e| is not disaptched at all.</li>
+                        <li>Otherwise (|e| is a <code>click</code> or <code>auxclick</code> event): let |td| and |tu| be the targets of the corresponding <code>pointerdown</code> and <code>pointerup</code> events respectively.  If both |td| and |tu| are [=connected=] [[DOM]] at the moment event |e| is being dispatched, the target of |e| MUST be the nearest common inclusive ancestor of |td| and |tu| in the DOM; otherwise, |e| is not dispatched at all.</li>
                     </ol>
                 </section>
             </section>
--- a/index.html
+++ b/index.html
@@ -821,11 +821,15 @@ function tilt2spherical(tiltX, tiltY) {
                             </div>
                         </li>
 
-                        <li>If |u| is not a <code>PointerEvent</code>, the dispatch of |e| MUST follow the [[UIEVENTS]] spec (the remaining steps below are skipped).</li>
+                        <li>If |u| is not a <code>PointerEvent</code>, dispatch |e| following the [[UIEVENTS]] spec (the remaining steps below are skipped).</li>
 
-                        <li>If |e| is a <code>contextmenu</code> event, or |u| was dispatched while the corresponding pointer was captured, then let |t| be the target of |u|.</li>
+                        <li>
+                            <p>Define |t| as follows:</p>
 
-                        <li>Otherwise (|e| is a <code>click</code> or <code>auxclick</code> event for which |u| is a <code>pointerup</code> event that was dispatched uncaptured) let |t| be the nearest common inclusive ancestor of the corresponding <code>pointerdown</code> and <code>pointerup</code> targets in the DOM at the moment event |e| is being dispatched.</li>
+                            <p>If |e| is a <code>contextmenu</code> event, or |u| was dispatched while the corresponding pointer was captured, then let |t| be the target of |u|.</p>
+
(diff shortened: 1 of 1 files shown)


mustaqahmed:
ptal: simplifying the text that will be updated for #492.

mustaqahmed:
Never mind, this is not ready for review: the main PR will rewrite this section.

mustaqahmed:
This is ready for review now.

index.html
```diff
@@ -811,15 +811,22 @@ <h4>Event coordinates</h4>
...
+                        </div>
+
+                        <li>If |t| is not a <code>PointerEvent</code>, the dispatch of |e| MUST follow the [[UIEVENTS]] spec.</li>
+
+                        <li>Otherwise, if |t| was dispatched while the corresponding pointer was captured, and the capturing target of event |t| is [=connected=] [[DOM]] at the moment event |e| is being dispatched, then the target of |e| MUST be the capturing target of event |t|, and |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</li>
+
+                        <li>Otherwise, if |e| is a <code>contextmenu</code> event, the target of |e| MUST be the target of |t|.</li>
+
+                        <li>Otherwise (|e| is a <code>click</code> or <code>auxclick</code> event): let |td| and |tu| be the targets of the corresponding <code>pointerdown</code> and <code>pointerup</code> events respectively.  If both |td| and |tu| are [=connected=] [[DOM]] at the moment event |e| is being dispatched, the target of |e| MUST be the nearest common inclusive ancestor of |td| and |tu| in the DOM; otherwise, |e| is not disaptched at all.</li>
```
smaug----: disaptched

smaug----: Hmm, this doesn't seem work the same way as https://github.com/w3c/pointerevents/issues/492#issuecomment-1823553102

patrickhlauke:
```suggestion
                        <li>Otherwise (|e| is a <code>click</code> or <code>auxclick</code> event): let |td| and |tu| be the targets of the corresponding <code>pointerdown</code> and <code>pointerup</code> events respectively.  If both |td| and |tu| are [=connected=] [[DOM]] at the moment event |e| is being dispatched, the target of |e| MUST be the nearest common inclusive ancestor of |td| and |tu| in the DOM; otherwise, |e| is not dispatched at all.</li>
```

patrickhlauke: fixed it

patrickhlauke: did you mean the code comment that can be merged? or the actual substance of this line?

index.html
```diff
@@ -811,15 +811,22 @@ <h4>Event coordinates</h4>
...
-                    </ul>
-                    <p>When these conditions are met, event |e| MUST be dispatched to the capturing target of event |t|, and event |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</p>
-                    <div class="note">When |t| is a <code>PointerEvent</code>, its type depends on native platform convention: if |e| is a <code>contextmenu</code> event, |t| is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event; if |e| is a <code>click</code> or <code>auxclick</code> event, |t| is a {{GlobalEventHandlers/pointerup}} event.</div>
+                    <p>A <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event |e| MUST follow the dispatch process defined in the [[UIEVENTS]] spec except for the overrides defined in the algorithm below:</p>
+                    <ol>
+                        <li>Let |t| be the user interaction event that caused the firing of |e|.</li>
+                        <div class="note">
```
patrickhlauke: can't have a `<div>` as child of an `<ol>`. probably need to add this into the first list item

index.html
```diff
@@ -811,15 +811,22 @@ <h4>Event coordinates</h4>
...
+                        </div>
+
+                        <li>If |t| is not a <code>PointerEvent</code>, the dispatch of |e| MUST follow the [[UIEVENTS]] spec.</li>
+
+                        <li>Otherwise, if |t| was dispatched while the corresponding pointer was captured, and the capturing target of event |t| is [=connected=] [[DOM]] at the moment event |e| is being dispatched, then the target of |e| MUST be the capturing target of event |t|, and |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</li>
+
+                        <li>Otherwise, if |e| is a <code>contextmenu</code> event, the target of |e| MUST be the target of |t|.</li>
+
+                        <li>Otherwise (|e| is a <code>click</code> or <code>auxclick</code> event): let |td| and |tu| be the targets of the corresponding <code>pointerdown</code> and <code>pointerup</code> events respectively.  If both |td| and |tu| are [=connected=] [[DOM]] at the moment event |e| is being dispatched, the target of |e| MUST be the nearest common inclusive ancestor of |td| and |tu| in the DOM; otherwise, |e| is not dispatched at all.</li>
```
flackr: I think this should be computed when |t| was dispatched. I.e. even if the dispatch of |tu| changes the structure it will not affect the target of the click. This also makes the timing of this consistent with when we check for capture, i.e. right before the |tu| dispatch.

flackr: Also, I don't think we don't need to specifically call out |td| and |tu| being connected, as if the common ancestor is connected they must be connected.

index.html
mustaqahmed: Yes, I meant within the first item.  Done.

index.html
mustaqahmed: > Also, I don't think we don't need to specifically call out |td| and |tu| being connected, as if the common ancestor is connected they must be connected.

Done.

mustaqahmed: > I think this should be computed when |t| was dispatched. I.e. even if the dispatch of |tu| changes the structure it will not affect the target of the click. This also makes the timing of this consistent with when we check for capture, i.e. right before the |tu| dispatch.

I just realized that Sec4.2.12 talks only about `click`-like events, and the spec currently mentions nothing about any other events' dispatch "algorithms"!

Anyway, I rewrote the last three steps while avoiding adding a dispatch algorithm for `pointerup` or any other event.

index.html
```diff
@@ -811,15 +811,28 @@ <h4>Event coordinates</h4>
...
-                    </ul>
-                    <p>When these conditions are met, event |e| MUST be dispatched to the capturing target of event |t|, and event |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</p>
-                    <div class="note">When |t| is a <code>PointerEvent</code>, its type depends on native platform convention: if |e| is a <code>contextmenu</code> event, |t| is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event; if |e| is a <code>click</code> or <code>auxclick</code> event, |t| is a {{GlobalEventHandlers/pointerup}} event.</div>
+                    <p>A <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event |e| MUST follow the dispatch process defined in the [[UIEVENTS]] spec except for the overrides defined in the algorithm below:</p>
+                    <ol>
+                        <li>
+                            <p>Let |u| be the user interaction event that caused the firing of |e|.</p>
```
flackr: I think it would be clearer to use names like |user event|, |target|, and |event| instead of |u|, |t| and |e|.

index.html
```diff
@@ -811,15 +811,28 @@ <h4>Event coordinates</h4>
...
+                            <div class="note">
+                                <p>Event |u| could be a non-<code>PointerEvent</code>; for example, it is a <code>KeyboardEvent</code> when a <code>click</code> event dispatch is caused by hitting the spacebar on a checkbox element.</p>
+                                <p>When |u| is a <code>PointerEvent</code>, |u| is a {{GlobalEventHandlers/pointerup}} except for a <code>contextmenu</code> event where |u| is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event depending on native platform convention.</p>
```
flackr: Suggest the following rewrite to reduce the complexity of the sentence:

```html
<p>When |u| is a <code>PointerEvent</code>, |u| is a {{GlobalEventHandlers/pointerup}} for a <code>click</code> or <code>auxclick</code> event, and either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event depending on native platform convention for a <code>contextmenu</code> event.</p>
```

index.html
```diff
@@ -811,15 +811,28 @@ <h4>Event coordinates</h4>
...
-                        <li>event |t| was dispatched while the corresponding pointer was captured, and</li>
-                        <li>the capturing target of event |t| is [=connected=] [[DOM]] at the moment event |e| is being dispatched.</li>
-                    </ul>
-                    <p>When these conditions are met, event |e| MUST be dispatched to the capturing target of event |t|, and event |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</p>
-                    <div class="note">When |t| is a <code>PointerEvent</code>, its type depends on native platform convention: if |e| is a <code>contextmenu</code> event, |t| is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event; if |e| is a <code>click</code> or <code>auxclick</code> event, |t| is a {{GlobalEventHandlers/pointerup}} event.</div>
+                    <p>A <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event |e| MUST follow the dispatch process defined in the [[UIEVENTS]] spec except for the overrides defined in the algorithm below:</p>
```
flackr: what about `dblclick`?

https://w3c.github.io/uievents/#event-flow:~:text=SHOULD%20fire%20click%20and%20dblclick%20events%20on%20the%20nearest%20common%20inclusive%20ancestor%20when%20the%20associated%20mousedown%20and%20mouseup%20event%20targets%20are%20different.

I also wonder if it may be simpler to say this overrides the target, but that otherwise the dispatch process follows [[UIEVENTS]].

index.html
```diff
@@ -811,15 +811,28 @@ <h4>Event coordinates</h4>
...
+
+                        <li>
+                            <p>Define |t| as follows:</p>
+
+                            <p>If |e| is a <code>contextmenu</code> event, or |u| was dispatched while the corresponding pointer was captured, then let |t| be the target of |u|.</p>
+
+                            <p>Otherwise (|e| is a <code>click</code> or <code>auxclick</code> event for which |u| is a <code>pointerup</code> event that was dispatched uncaptured) let |t| be the nearest common inclusive ancestor of the corresponding <code>pointerdown</code> and <code>pointerup</code> targets in the DOM at the moment event |e| is being dispatched.</p>
+                        </li>
+
+                        <li>Dispatch |e| to |t| following the [[UIEVENTS]] spec.  If |u| was dispatched while the corresponding pointer was captured, |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</li>
```
flackr: We already require that lostpointercapture is dispatched before any "subsequent events for the pointer" in https://w3c.github.io/pointerevents/#the-lostpointercapture-event right? It probably shouldn't be normatively required in two places.

index.html
mustaqahmed: Done.  Put the phrase on platform dependence in parentheses for better readability, ptal.

index.html
mustaqahmed: > what about `dblclick`?

By default, it will follow the UIEvent spec as a MouseEvent.  Making it a PointerEvent is a separate problem.

> I also wonder if it may be simpler to say this overrides the target, but that otherwise the dispatch process follows [[UIEVENTS]].

Done, ptal.

index.html
mustaqahmed: Hmm.  The space in |user event| makes it look like two separate variables to me.  What about |user_event|?

index.html
```diff
@@ -811,15 +811,31 @@ <h4>Event coordinates</h4>
...
-                        <li>event |t| was dispatched while the corresponding pointer was captured, and</li>
-                        <li>the capturing target of event |t| is [=connected=] [[DOM]] at the moment event |e| is being dispatched.</li>
-                    </ul>
-                    <p>When these conditions are met, event |e| MUST be dispatched to the capturing target of event |t|, and event |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</p>
-                    <div class="note">When |t| is a <code>PointerEvent</code>, its type depends on native platform convention: if |e| is a <code>contextmenu</code> event, |t| is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event; if |e| is a <code>click</code> or <code>auxclick</code> event, |t| is a {{GlobalEventHandlers/pointerup}} event.</div>
+                    <p>A <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event |event| MUST follow the dispatch process defined in the [[UIEVENTS]] spec except that |event| target is overridden using the algorithm below:</p>
```
mustaqahmed: It's subjective but "event |event|" looks very odd to me in the rendered spec!  Thoughts?

index.html
mustaqahmed: > We already require that lostpointercapture is dispatched before any "subsequent events for the pointer" in https://w3c.github.io/pointerevents/#the-lostpointercapture-event right? It probably shouldn't be normatively required in two places.

Made it a non-normative text.  We added it few months ago because it is an important exception that deserves attention here.

index.html
flackr: You can just write "|event|" since the name already says that it's an event.

mustaqahmed: Rewrote to avoid this.

index.html
```diff
@@ -811,15 +811,31 @@ <h4>Event coordinates</h4>
...
-                    <p>When these conditions are met, event |e| MUST be dispatched to the capturing target of event |t|, and event |e| MUST be dispatched after the dispatch of the {{GlobalEventHandlers/lostpointercapture}} event with the same <code>pointerId</code>.</p>
-                    <div class="note">When |t| is a <code>PointerEvent</code>, its type depends on native platform convention: if |e| is a <code>contextmenu</code> event, |t| is either a {{GlobalEventHandlers/pointerdown}} or a {{GlobalEventHandlers/pointerup}} event; if |e| is a <code>click</code> or <code>auxclick</code> event, |t| is a {{GlobalEventHandlers/pointerup}} event.</div>
+                    <p>A <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event MUST follow the dispatch process defined in the [[UIEVENTS]] spec except that the event target is overridden using the algorithm below:</p>
+                    <ol>
+                        <li>
+                            <p>Let |event| be the <code>click</code>, <code>auxclick</code> or <code>contextmenu</code> event being dispatched, and |user_event| be the user interaction event that caused the firing of |event|.</p>
```
smaug----: This is a bit handwavy, but not sure we can have anything better easily.

smaug---- (approved):

flackr (approved):

patrickhlauke:
Per decision at today's meeting https://www.w3.org/2024/01/31-pointerevents-minutes.html I'm merging this. thanks @mustaqahmed @flackr @smaug----

patrickhlauke:
does this need a new WPT?

mustaqahmed:
We need to update `pointerevents/pointerevent_click_during_capture.html` to match this PR.

mustaqahmed:
The WPT was added [a while ago](https://chromium-review.googlesource.com/c/chromium/src/+/5301480)!

PR #496, opened 2024-01-21
Add logical/abstract values for `touch-action`

### LINKED ISSUE #272: Add logical dimension values for touch-action property, since logical is/has shipped
https://w3c.github.io/pointerevents/#the-touch-action-css-property
https://drafts.csswg.org/css-logical-1/
https://caniuse.com/#feat=pointer
https://caniuse.com/#feat=css-logical-props

With Webkit being the last major vendor to work on pointer events, initially deployed in STP Release 67, as well as Edge being the only vendor without impending support for logical properties and values, it seems the `touch-action` property needs to be updated to accept a set of logical values.

cc @mbrubeck @atanassov

patrickhlauke:
Adds `pan-inline`, `pan-inline-start`, `pan-inline-end`, `pan-block`, `pan-block-start`, `pan-block-end` values for `touch-action`, and expands the explanation to reference https://www.w3.org/TR/css-writing-modes-3/

Closes https://github.com/w3c/pointerevents/issues/272

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/496.html" title="Last updated on Jan 21, 2024, 4:28 PM UTC (fec4b39)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/496/8bf7955...fec4b39.html" title="Last updated on Jan 21, 2024, 4:28 PM UTC (fec4b39)">Diff</a>


patrickhlauke:
Discussed at last PEWG meeting (31 Jan). There's a few more tiny language tweaks I'd love to make, just to make it clear which element's writing mode needs to be taken into consideration (tl;dr: the element where the `touch-action` is defined, and not any of the child elements inside it). setting back to draft until I sorted that.

smaug----:
Removing needs-wpt for now, since this isn't even in the spec. We definitely should add a wpt at the same time when landing this

patrickhlauke:
discussed in today's PEWG meeting. want to include a graph/chart

need to also clarify explicitly what happens if you have `touch-action` defined on an element, and then a child element defines a different writing mode/direction (in short, only the one on the element that defined `touch-action` counts)

PR #524, opened 2024-10-23
Clarify isTrusted bits in coalesced/predicted event lists

### LINKED ISSUE #514: Coalesced and predicted event attributes within an untrusted event
The spec [Sec 10](https://w3c.github.io/pointerevents/#coalesced-and-predicted-events) mentions  the following about the attributes of the "contained events" for a JS-constructed (untrusted) event, which needs a clarification for at least `isTrusted` and `target` attributes:
> Untrusted events have their coalesced/predicted events list initialized to the value passed to the constructor.

This text seems to suggest that the constructor allows setting any arbitrary value to an attribute.  However, when I run the following code in Chrome and Firefox:
```JS
// Read default values:
let e0 = new PointerEvent("pointermove");
console.log(e0.isTrusted, e0.bubbles, e0.clientX, e0.pointerType, e0.target);
// Read assigned values:
let e1 = new PointerEvent("pointermove", {isTrusted: true, bubbles: true, clientX: 1234, pointerType: "hand", target: document.body});
console.log(e1.isTrusted, e1.bubbles, e1.clientX, e1.pointerType, e1.target);
```
both browsers emit the following:
```
false false 0 '' null
false true 1234 'hand' null
```
So `isTrusted` and `target` attributes do not allow arbitrary values while other attributes allow.  I didn't see any indication in IDLs to explain this behavior!

In any case, we need clarifications for those two special attributes _in coalesced and predicted event lists_:
- Within a JS-constructed "parent" event, will they always have their default values?
- When a trusted "parent" event is redispatched from JS, will they be reset to their default values?

mustaqahmed:
Closes #514.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/pull/524.html" title="Last updated on Oct 23, 2024, 2:48 PM UTC (acb3891)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/524/f722c92...acb3891.html" title="Last updated on Oct 23, 2024, 2:48 PM UTC (acb3891)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1212,6 +1212,10 @@ partial interface Navigator {
             or all of the coalesced events, but not both.
             </div>
 
+            <div class="note">
+                When a trusted event containing a <a>coalesced events list</a> is re-dispatched from JavaScript, the event dispatch algorithm sets the {{Event/isTrusted}} bit of the event to <code>false</code> but the same bits in the <a>coalesced events list</a> remain unchaged from their original <code>true</code> values.
+            </div>
+
             <p>The events in the coalesced events list of a trusted event will have:</p>
             <ul>
                 <li>Monotonically increasing {{Event/timeStamp}} values [[DOM]] — all coalesced events have a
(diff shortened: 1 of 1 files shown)


mustaqahmed:
@smaug----, Please take a look.

aprotyas (commented):
Is there a WPT for this behavior?

smaug---- (approved):
These are notes only, and seem to clarify nicely the behavior.

Obviously if those coalesced events themselves are dispatched using JS, they become untrusted, but that is true for all the events anyhow.

mustaqahmed:
> Is there a WPT for this behavior?

We have [this test](https://wpt.fyi/results/pointerevents?label=master&label=experimental&aligned&q=coalesced_events_attributes_on_redispatch.https) that will need to become non-tentative after this PR lands.

Issue #529, opened 2024-11-25
Mouse and Pointer events from a stationary mouse


mustaqahmed:
Forking off from #457 to highlight my question there:

If the mouse is held stationary, do we expect to see CSS hover effect and pointer/mouse events after a touch/wheel scroll ends?

There are important differences between major browsers here:

- [Repro for top frame](https://mustaqahmed.github.io/web/interop/event-from-non-moving-mouse.html):
  - After a short delay, Chrome fires pointer/mouse over/out events and no move event, and shows hover effect.
  - After a short delay, Firefox fires mouseover/out events only, and shows hover effect.
  - ~Safari fires no events, and shows hover effect.~ EDIT: Safari fires pointer/mouse move/over/out events, and shows hover effect.

- [Repro for sub-frame](https://codepen.io/mustaqahmed/full/pvzzxbe):
  - Chrome fires no events, and shows no hover effect.
  - After a short delay, Firefox fires mouseover/out events only, and shows hover effect.
  - ~[TODO]: need to check.~ EDIT: Safari fires pointer/mouse move/over/out events, and shows hover effect.

I think it is important to fire some events to let the developers know when the hover-effect is shown, and in that case both pointer and mouse over/out events need to be fired for consistency between these two classes of events, like what Chrome does.  But to be frank, I got confused upon discovering Chrome's subframe behavior above!

---

The PointerEvents spec expects no event in this case, but the UIEvent spec expects mouse events!

Related WPTs: [pointerevents/pointerevent_pointerout_no_pointer_movement.html](https://wpt.fyi/results/pointerevents/pointerevent_pointerout_no_pointer_movement.html?label=master&label=experimental&aligned) and [uievents/mouse/layout_change_should_fire_mouseover.html](https://wpt.fyi/results/uievents/mouse/layout_change_should_fire_mouseover.html?label=master&label=experimental&aligned)

mustaqahmed:
FYI: @flackr @smaug---- @masayuki-nakano

flackr:
> There are important differences between major browsers here:
> 
> * [Repro for top frame](https://mustaqahmed.github.io/web/interop/event-from-non-moving-mouse.html):
>   
>   * After a short delay, Chrome fires pointer/mouse over/out events and no move event, and shows hover effect.
>   * After a short delay, Firefox fires mouseover/out events only, and shows hover effect.
>   * Safari fires no events, and shows hover effect.

[...]

Issue #520, opened 2024-09-17
[PointerEvent algorithms] Remove 1:1 correspondence between Mouse and Pointer boundary events


mustaqahmed:
The UIEvents spec assumes there is "almost" a 1:1 correspondence between mouseenter and pointerenter (and same for other boundary events: mouseleave, mouseover, mouseout).  For example, Step 4.1.4 in [UIEvents Section 3.4.3.15. handle native mouse move](https://w3c.github.io/uievents/#handle-native-mouse-move-id) says “maybe send pointerenter event with mouseenter” where the “maybe” is vague and yet it is unable to cover the underlying complexity with multiple pointers.

### Possible solution
Solve #519 (Order of boundary events) first, then link the corresponding algorithms in UIEvents and PointerEvents through [PE Section 11.1 Tracking the effective position of the legacy mouse pointer](https://w3c.github.io/pointerevents/#dfn-tracking-the-effective-position-of-the-legacy-mouse-pointer).

mustaqahmed:
FYI: @garykac @flackr @smaug----

mustaqahmed:
FYI: we discussed the spec vagueness around boundary events in https://github.com/w3c/pointerevents/pull/532#discussion_r1932921615.

PR #532, opened 2024-12-17
Update boundary event behavior for a stationary pointing device.

mustaqahmed:
PEWG decided in #529 that the PointerEvents spec should match the UIEvent spec behavior for non-moving pointing device: boundary events should fire after a layout change, and no move events should fire.

***
<a href="https://pr-preview.s3.amazonaws.com/mustaqahmed/pointerevents/pull/532.html" title="Last updated on Jan 29, 2025, 8:54 PM UTC (3b7089b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/532/ec76bec...mustaqahmed:3b7089b.html" title="Last updated on Jan 29, 2025, 8:54 PM UTC (3b7089b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -612,7 +612,7 @@ interface PointerEvent : MouseEvent {
             <section>
                 <h3><dfn data-lt="events-from-layout-changes">Boundary events caused by layout changes</dfn></h3>
                 <p>A pointing device that moved relative to the screen surface or underwent some change in any of its properties fires various events as defined in <a>Pointer Event types</a>.  For a stationary pointing device (that neither moved relative to the screen surface nor underwent any change in any properties), the user agent MUST fire certain boundary events after a layout change that affected the <a>hit test</a> target for the pointer, see {{GlobalEventHandlers/pointerover}}, {{GlobalEventHandlers/pointerenter}}, {{GlobalEventHandlers/pointerout}} and {{GlobalEventHandlers/pointerleave}} for details.  The user agent MAY delay the firing of these boundary events because of performance reasons (e.g. to avoid too many hit-tests or layout changes caused by boundary event listeners).</p>
-                <div class="note">A stationary pointing device (that neither moved relative to the screen surface nor underwent any change in any properties) never fires a {{GlobalEventHandlers/pointermove} event.</div>
+                <div class="note">A stationary pointing device (that neither moved relative to the screen surface nor underwent any change in any properties) never fires a {{GlobalEventHandlers/pointermove}} event.</div>
             </section>
 
             <section>
--- a/index.html
+++ b/index.html
@@ -389,6 +389,7 @@ interface PointerEvent : MouseEvent {
                     <p>Instead, chorded button presses can be detected by inspecting changes to the <code>button</code> and <code>buttons</code> properties. The <code>button</code> and <code>buttons</code> properties are inherited from the {{MouseEvent}} interface, but with a change in semantics and values, as outlined in the following sections.</p>
                     <p>The modifications to the <code>button</code> and <code>buttons</code> properties apply only to pointer events. For any <a>compatibility mouse events</a> the value of <code>button</code> and <code>buttons</code> MUST follow [[UIEVENTS]].</p>
                 </section>
+
                 <section>
                     <h4>The <code>button</code> property</h4>
                     <p>To identify button state transitions in any pointer event (and not just {{GlobalEventHandlers/pointerdown}} and {{GlobalEventHandlers/pointerup}}), the <code>button</code> property indicates the device button whose state change fired the event.</p>
@@ -406,6 +407,7 @@ interface PointerEvent : MouseEvent {
                     </table>
                     <div class="note">During a mouse drag, the value of the <code>button</code> property in a {{GlobalEventHandlers/pointermove}} event will be different from that in a <code>mousemove</code> event. For example, while moving the mouse with the right button pressed, the {{GlobalEventHandlers/pointermove}} events will have the <code>button</code> value -1, but the <code>mousemove</code> events will have the <code>button</code> value 2.</div>
                 </section>
+
                 <section>
                     <h4>The <code>buttons</code> property</h4>
                     <p>The <code>buttons</code> property gives the current state of the device buttons as a bitmask (same as in <code>MouseEvent</code>, but with an expanded set of possible values).</p>
@@ -423,6 +425,7 @@ interface PointerEvent : MouseEvent {
                     </table>
                 </section>
             </section>
+
             <section>
                 <h3>The <dfn>primary pointer</dfn></h3>
                 <p>In a multi-pointer (e.g. multi-touch) scenario, the <code>isPrimary</code> property is used to identify a master pointer amongst the set of <a data-lt="active pointer">active pointers</a> for each pointer type.</p>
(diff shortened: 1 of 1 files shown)


smaug----:
@masayuki-nakano you may have some opinion on the pr

index.html
```diff
@@ -716,26 +725,41 @@ <h3>Converting between <code>tiltX</code> / <code>tiltY</code> and <code>altitud
...
+                <p>The user agent MUST <a>fire a pointer event</a> named {{GlobalEventHandlers/pointerover}} when any of the following occurs:</p>
+                <ul>
+                    <li>A pointing device is moved into the <a>hit test</a> boundaries of an element.  Note that <code>setPointerCapture()</code> or <code>releasePointerCapture()</code> might have changed the <a>hit test</a> target.  Also note that while a pointer is captured it is considered to be always inside the boundaries of the capturing element for the purpose of firing boundary events.</li>
+                    <li>A <a data-lt="events-from-layout-changes">layout change</a> causes the <a>hit test</a> boundaries of an element to move underneath an uncaptured primary pointer of a stationary pointing device.</li>
```
flackr: Why do you call out "primary pointer" here? I think that any uncaptured pointer is supposed to generate boundary events, right? https://w3c.github.io/pointerevents/#the-pointerover-event

index.html
```diff
@@ -783,39 +809,47 @@ <h3>The <dfn data-dfn-for="GlobalEventHandlers" data-event-type="event">pointeru
...
+                    <li>After the user agent fires the {{GlobalEventHandlers/pointerup}} event for a device that <a href=#mapping-for-devices-that-do-not-support-hover>does not support hover</a> (see {{GlobalEventHandlers/pointerup}}).</li>
                     <li>The user agent has detected a scenario to <a>suppress a pointer event stream</a>.</li>
                 </ul>
-                <p>This event type is similar to {{GlobalEventHandlers/pointerout}}, but differs in that it does not bubble and that it MUST not be fired until the pointing device has left the boundaries of the element and the boundaries of all of its descendants.</p>
-                <div class="note">There are similarities between this event type, the <code>mouseleave</code> event described in [[UIEVENTS]], and the CSS <code>:hover</code> pseudo-class described in [[CSS21]]. See also the <code>pointerenter</code> event.</div>
+                <div class="note">This event type is similar to {{GlobalEventHandlers/pointerout}} but with two differences: {{GlobalEventHandlers/pointerleave}} does not bubble, and its dispatch considers the <a>hit test</a> boundaries of even descendant elements.</div>
```
flackr, line 839: nit: I prefer the originaly terminology for referring to the descendants, i.e. "boundaries of all of its descendant elements".

index.html
```diff
@@ -716,26 +725,41 @@ <h3>Converting between <code>tiltX</code> / <code>tiltY</code> and <code>altitud
...
+                <p>The user agent MUST <a>fire a pointer event</a> named {{GlobalEventHandlers/pointerover}} when any of the following occurs:</p>
+                <ul>
+                    <li>A pointing device is moved into the <a>hit test</a> boundaries of an element.  Note that <code>setPointerCapture()</code> or <code>releasePointerCapture()</code> might have changed the <a>hit test</a> target.  Also note that while a pointer is captured it is considered to be always inside the boundaries of the capturing element for the purpose of firing boundary events.</li>
```
flackr: Instead of calling out the hit test boundary and interaction with capture, it seems like we should refer to the generic targeting of pointerevents, e.g. the target determination in https://w3c.github.io/pointerevents/#firing-events-using-the-pointerevent-interface which already has taken this and capture into account.

index.html
mustaqahmed: The UI Event spec currently mentions that the stationary device boundary events come only from the "primary pointing device", e.g. see the [mouseeneter](https://w3c.github.io/uievents/#event-type-mouseenter) section.

I can imagine this was done to minimize event-dispatch related overhead for a stationary device but not sure.  Any chance hover effect is limited to primary (first) touch pointer?

index.html
mustaqahmed: The normative points above still mentions "all descendants".  In this non-normative text, I wanted to highlight only the main difference, i.e., whether descendants are considered or not.

See the very similar note above for `pointerenter`: it was about "one descendant".

index.html
mustaqahmed: You are right, this is all about generic targeting.  I can easily refer to Sec4.1.3 for `pointerover` and `pointerout` but the same becomes tricky for `pointerenter` and `pointerleave` because of the involvement of descendant element boundaries here.  Any suggestions?

index.html
mustaqahmed: On Android, the hover effect is shown only for the primary touch pointer (and it vanishes at the moment a non-primary touch pointer becomes active).

index.html
flackr: It's still the same thing, just if the target is any of the inclusive descendants rather than just the element itself.

E.g. for pointerleave it's if the determined target of the pointer is any of the inclusive descendants of the element.

mustaqahmed: Great suggestion!  Done.

index.html
```diff
@@ -606,6 +609,12 @@ <h4>Suppressing a pointer event stream</h4>
                 </section>
             </section>
 
+            <section>
+                <h3><dfn data-lt="events-from-layout-changes">Boundary events caused by layout changes</dfn></h3>
+                <p>A pointing device that moved relative to the screen surface or underwent some change in any of its properties fires various events as defined in <a>Pointer Event types</a>.  For a stationary pointing device (that neither moved relative to the screen surface nor underwent any change in any properties), the user agent MUST fire certain boundary events after a layout change that affected the <a>hit test</a> target for the pointer, see {{GlobalEventHandlers/pointerover}}, {{GlobalEventHandlers/pointerenter}}, {{GlobalEventHandlers/pointerout}} and {{GlobalEventHandlers/pointerleave}} for details.  The user agent MAY delay the firing of these boundary events because of performance reasons (e.g. to avoid too many hit-tests or layout changes caused by boundary event listeners).</p>
```
masayuki-nakano, line 614: > The user agent MAY delay the firing of these boundary events because of performance reasons (e.g. to avoid too many hit-tests or layout changes caused by boundary event listeners).

I think it should be defined here that if the last `pointerover` target appears under the pointer after temporarily removed. I think `pointerover` should be fired again to avoid the timing issue of the delay vs. async handling in web apps.  Otherwise, we could see another issue which is the case when the last `pointerover` target is back under the pointer but some ancestor elements which got `pointerenter` events fired are not back again.

(It's fine to handle this in a new issue.)

masayuki-nakano (changes requested):

index.html
mustaqahmed: I totally agree this PR is unclear about such DOM modifications.  Let's fix it separately after this PR lands.

I think the proper way to address such cases would be introducing a state variable for "current element under pointer" for each pointer, and dispatch boundary events whenever this changes.  That will also help us port the [Mouse Event Order](https://w3c.github.io/uievents/#events-mouseevent-event-order) section from UIEvent into PoitnerEvent spec (which will nicely cover a part of our TPAC 2024 plan).

masayuki-nakano (approved):
Putting off to fix my concern in the separate issue. So, I'm now fine with the change.

mustaqahmed:
> Putting off to fix my concern in the separate issue. So, I'm now fine with the change.

Thanks.  I just linked this to #520, let's continue the discussion there.

index.html
mustaqahmed: Removed "primary" as per PEWG discussion today.

PR #637, opened 2026-02-05
Allow pointerup-activated elements to set click target differently.

dbaron:
This allows other specifications to define particular elements to be pointerup-activated elements, which have different behavior for choosing the element that gets the click event.  These elements can get a click event when the pointerup is inside them, even when the pointerdown is not.

This is the approach I'd like to take to implement what the OpenUI CG resolved that it wanted in:
https://github.com/openui/open-ui/issues/1312#issuecomment-3469585173

This is desirable so that author-defined `click` handlers will work reliably for activations of these elements, which would not otherwise be the case.  This allows avoiding a behavior that would be confusing to authors where a `click` handler would work for many user gestures that activate the element but also fail to work for a significant number of the gestures that do so.

I'm curious what you think of this approach.  I realize it might require some amount of discussion, but given that I think the spec change needed is quite small I figured I'd try starting that discussion with the PR.

***
<a href="https://pr-preview.s3.amazonaws.com/dbaron/pointerevents/pull/637.html" title="Last updated on Feb 5, 2026, 3:52 PM UTC (c2b0269)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/pointerevents/637/146766e...dbaron:c2b0269.html" title="Last updated on Feb 5, 2026, 3:52 PM UTC (c2b0269)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -894,9 +894,13 @@ function tilt2spherical(tiltX, tiltY) {
                         <li>
                             <p>Define |target| as follows:</p>
 
-                            <p>If |event| is a <code>contextmenu</code> event, or |userEvent| was dispatched while the corresponding pointer was captured, then let |target| be the target of |userEvent|.</p>
+                            <ol>
+                                <li>If |event| is a <code>contextmenu</code> event, or |userEvent| was dispatched while the corresponding pointer was captured, then let |target| be the target of |userEvent|.</li>
 
-                            <p>Otherwise (|event| is a <code>click</code> or <code>auxclick</code> event for which |userEvent| is a <code>pointerup</code> event that was dispatched uncaptured) let |target| be the nearest common inclusive ancestor of the corresponding <code>pointerdown</code> and <code>pointerup</code> targets in the DOM at the moment |event| is being dispatched.</p>
+                                <li>Otherwise, if the nearest ancestor |specialAncestor| of the <code>pointerup</code> target that is a <a>pointerup-activated element</a> is not also an ancestor of the <code>pointerdown</code> target, let |target| be |specialAncestor|.</li>
+
+                                <li>Otherwise, let |target| be the nearest common inclusive ancestor of the corresponding <code>pointerdown</code> and <code>pointerup</code> targets in the DOM at the moment |event| is being dispatched.</li>
+                            </ol>
                         </li>
 
                         <li>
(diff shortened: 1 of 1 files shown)


dbaron:
This PR [was discussed](https://www.w3.org/2026/02/11-pointerevents-minutes.html#6a6e) in the February 11 Pointer Events meeting.

The discussion there suggested using [Pointer Capture](https://w3c.github.io/pointerevents/#pointer-capture) as an alternative.  I think this seems possible, but it also seems like it has a set of issues of its own.

In particular, I think the model if we went with pointer capture would be that when the mouse is down and over a menuitem element, that element captures the pointer (and release it when it leaves).  This capturing probably isn't strictly necessary for the first menuitem (before the pointer moves into a different one), but it's probably good to do it anyway to be consistent.  Then, when the mouseup happens (and I *think* the mouseup still gets sent to its normal target -- I hope so at least), the click will go to the menuitem that's doing the capturing.

This seems a little "weird" in 2 ways:
1. The capturing is constantly changed across a single user gesture.  This seems like an "unusual" use of capturing, but maybe that's OK.
2. It's a little unclear to me how a web platform primitive is supposed to invoke setting and releasing of pointer capture.  Is it OK to just invoke those algorithms, with the resulting effects on any developer-set pointer capture and on the `hasPointerCapture` API, or does it need to guard against developer-set pointer capture, and only capture if there is none?

I'm curious if anyone (e.g., @flackr @mustaqahmed @smaug----) has opinions on whether that seems like a workable approach and what you think of the comments above.

dbaron:
Or, alternatively, I now see @smaug---- 's comment that "capturing on the up might work?", so maybe that is the desired approach?

dbaron:
Also, after some further investigation I think capturing during mouseup is going to be a better approach than what I suggested initially, because there actually need to be conditions for which mouseup events do capturing and which ones don't.  I think the interesting question then becomes whether those conditions should be specified in the HTML spec or just discussed in a note in the HTML spec that leaves it up to the implementation to implement good UI for menu elements within certain parameters.

dbaron:
My initial attempt at capturing during mouseup didn't work.  Notes about why are in the commit message of https://crrev.com/c/7899892 (current version is https://crrev.com/c/7899892/3).  I haven't yet worked out whether the things I ran into are specific to Chromium's implementation or are required by the spec.

sb3nder:
Another approach could be to fire another click event for the `pointerup`/`mouseup` interaction.
It would be somewhat in the same category as the `click` event fired by a `label` element.

---

> I haven't yet worked out whether the things I ran into are specific to Chromium's implementation or are required by the spec.

From a web development perspective, I always thought that it was impossible to call pointer capture on `pointerup`.

dbaron:
The concepts that made it not work are at least in the spec, see "pending pointer capture target override" in step 6 of https://w3c.github.io/pointerevents/#setting-pointer-capture and in step 3 of https://w3c.github.io/pointerevents/#dfn-process-pending-pointer-capture .

flackr:
My preferred approach would be if we can have menus use a nested structure, the click handler for the menu element could determine the correct menu item to dispatch the click to, somewhat similar to labels, but based on the pointerup target it had observed. This wouldn't require any special pointer retargeting.

dbaron:
I ended up doing https://github.com/w3c/pointerevents/pull/637#issuecomment-4684760540 in https://github.com/whatwg/html/commit/89a843fb8716098bf3c772062b5a8cf1489b0ca3 and https://crrev.com/c/8133028 and https://github.com/web-platform-tests/wpt/pull/61506 so closing this PR as it's no longer needed.

